PR #2, opened 2017-01-12
Add signal handling and pidfile support to prep for init scripts

basepi:


basepi:
Added the init script as well. It's not being included in the rpm packages currently.

pkg/hubble
```diff
@@ -0,0 +1,95 @@
...
+# Default-Stop:      0 1 6
+# Short-Description: Start daemon at boot time
+# Description:       Enable service provided by daemon.
+### END INIT INFO
+
+dir="/etc/hubble"
+cmd="hubble -d"
+user="root"
+
+name=`basename $0`
```
cedwards: backticks?! gross.

basepi: Meh, I stole the template. I could run through and replace them with `$()` if it makes you feel better. :)

PR #8, opened 2017-02-02
Pyinstaller cleanups and included the osquery installer

kravivar:


scripts/installessentials.sh
```diff
@@ -0,0 +1,6 @@
+# Installing required packages
+apt-get install python-dev libffi-dev libssl-dev libxml2-dev libxslt1-dev libjpeg8-dev zlib1g-dev make cmake python-setuptools gcc -y
```
anuragpaliwal80: How is this going to work on RedHat family?

PR #23, opened 2017-02-13
Finish rpm build script, reduce verbosity of default daemon logging

basepi:


basepi:
@kravivar I fixed init_pkg.sh to automatically install python27 and use scl in the sub calls. It seems to be working beautifully for me, but you should take a look. I'm going to merge this but we can fix it easily.

PR #37, opened 2017-03-09
Clean up packaging code

basepi:
Primarily this pulls everything into a single `pkg/` directory, including pyinstaller hooks and init_pkg.sh. It also fixes up build_rpms.sh and build_debs.sh to be a bit smarter and not rely on the home directory at all.

@kravivar I haven't uncovered any issues in my testing of these changes, but it's probably still worth a glance from you.


kravivar:
@basepi : Looks good to my eyes

PR #38, opened 2017-03-10
Fixing aws details issues for azure machine

anuragpaliwal80:
On azure machine, it's trying to populate aws data, this needs to be fixed.
screen shot 2017-03-10 at 5 31 40 pm
screen shot 2017-03-10 at 5 31 46 pm
screen shot 2017-03-10 at 5 32 01 pm

### DIFF
--- a/hubblestack/extmods/returners/splunk_nebula_return.py
+++ b/hubblestack/extmods/returners/splunk_nebula_return.py
@@ -40,7 +40,7 @@ be skipped:
 '''
 import socket
 # Import AWS details
-from aws_details import *
+from aws_details import get_aws_details
 
 # Imports for http event forwarder
 import requests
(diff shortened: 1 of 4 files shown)


hubblestack/extmods/returners/splunk_nebula_return.py
```diff
@@ -39,6 +39,8 @@
               - product_group
 '''
 import socket
+# Import AWS details
+from aws_details import *
```
basepi: Can we change this to `from aws_details import get_aws_details` please? I don't like `*` imports, it's hard to tell where functions are coming from, and can trample your global context.

Same with the other files, but I'll only leave the one comment.

anuragpaliwal80: Sure. Agree. Will do that.

PR #59, opened 2017-06-13
Apaliwal/add misc functions

anuragpaliwal80:
PLEASE DO NOT MERGE IT FOR NOW.

### DIFF
--- a/hubblestack/files/hubblestack_nova/misc.py
+++ b/hubblestack/files/hubblestack_nova/misc.py
@@ -44,7 +44,6 @@ misc:
 from __future__ import absolute_import
 import logging
 
-import subprocess
 import fnmatch
 import yaml
 import os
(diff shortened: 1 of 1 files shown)


anuragpaliwal80:
@basepi Can you please review and let me know if this is correct way to implement checks in misc file ?

basepi:
This looks pretty good -- however, instead of re-implementing shell commands, can you use `__salt__['cmd.run']()` instead? I think there are examples in the other nova modules. We have access to all the salt modules so there's no need to reinvent the wheel.

basepi:
Also, it looks like in https://github.com/hubblestack/hubblestack_data/pull/26 you're not actually passing in a reason. In that case, if the function doesn't need any arguments, you can just have it take zero arguments.

anuragpaliwal80:
Got it. Sure let me try with __salt__['cmd.run'](). Thanks for the direction.

anuragpaliwal80:
If any checks fails and we want to return the reason, we can use first argument to return failure with reason. I referred:
`def test_failure_reason(reason):
    '''
    Automatically returns failure, with a reason (first arg)
    '''
    return reason`
That's the reason why I have passed blank arg so that we can show the failures to the users. Like
` 'Failure': [{'CIS-5.2.4': {'args': [''],
                            'description': 'Ensure system accounts are non-login',
                            'failure_reason': 'zabbix:x:498:497::/home/zabbix:/bin/bash',
                            'function': 'system_account_non_login',
                            'module': 'misc',
                            'nova_profile': 'misc',
                            'tag': 'CIS-5.2.4'}}],`

basepi:
Right, I just noticed you weren't using the reason in your other pull request, so I wanted to be clear that it's optional.

In fact, we could give it a default, like `reason=''` or something, that way you could leave off the reason rather than passing in an empty string.

anuragpaliwal80:
Sure. Got it.

mew1033:
A handful of these could be solved with an awk nova module. @fossam and I have been talking about it for a while now.

anuragpaliwal80:
@mew1033 Can you please share the PR for the same so that I can have a look? I am not able to see awk nova module. -thanks

mew1033:
No PR for it yet. We've just talked about the idea.

anuragpaliwal80:
Oh ok. Got it.

anuragpaliwal80:
@basepi We are done for this sprint, please review and merge the changes (if all is ok). FYI... We will be adding more checks in coming sprint and hence will start new PR for the same.

PR #97, opened 2017-07-20
Added unit tests

rashmip29:


basepi:
Ideally, unit tests should need few (or no) external dependencies or setup. They should set themselves up, and tear themselves down. However, I see a few references to absolute paths, like `'/root/myhubblefiles/hubblestack_nebula_queries.yaml'`. Can we make these relative paths and pull these files into a subdirectory under the `unittests/` directory? I want to make it so you can clone this repo to almost any machine, in almost any directory, and run the unit tests successfully without having to put specific files in specific folders.

Additionally, can we rename the `unittests` folder to `tests` instead? Eventually we'll be adding integration tests and I think a more generic name would be good.

basepi:
I'm going to close this for now. We can re-open this or create a new PR when you're ready to get back to this.

PR #115, opened 2017-07-25
initial working Logstash HTTP input returners

cedwards:
These three returners are initial working concepts for ingesting Hubble data into Logstash for use in ELK stacks. The http input module for Logstash is required, as well as enabling the json filter source:

Example logstash config:
```
input {
  http {
    port => '8080'
    host => '0.0.0.0'
    type => 'hubble'
    user => 'hubble'
    password => 'password'
  }
}

filter {
  json {
    source => 'message'
  }
}

output {
  elasticsearch { hosts => ["localhost:9200"] }
  stdout { codec => rubydebug }
}
```

### DIFF
--- a/hubblestack/extmods/returners/logstash_nebula_return.py
+++ b/hubblestack/extmods/returners/logstash_nebula_return.py
@@ -1,5 +1,23 @@
 # -*- encoding: utf-8 -*-
 '''
+HubbleStack Nebula-to-Logstash (http input) returner
+
+:maintainer: HubbleStack
+:platform: All
+:requires: HubbleStack
+
+Deliver HubbleStack Nebula query data into Logstash using the HTTP input
+plugin. Required config/pillar settings:
+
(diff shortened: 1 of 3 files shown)


fossam:
My suggestion is to use the official python library ( https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/index.html ). It will probably make it easier dealing with json structure.

cedwards:
I looked at that, but I didn't want to require `elasticsearch` as a dependency.

fossam:
It might make it easier to keep the code synced with the api.
There seems to be difference between using put over post - https://www.elastic.co/guide/en/elasticsearch/reference/current/_modifying_your_data.html

basepi:
This is going to do a *lot* of PUTs per job, but that's probably fine since it's all localhost. Is batching possible?

I'm going to merge this for now, and we can iterate if we eventually want to use the official library or add batching.

PR #116, opened 2017-07-26
copypasta error in splunk returners

cedwards:
I noticed these copypasta errors while I was code-diving.


basepi:
Awesome, thanks!

PR #117, opened 2017-07-26
Win upgrade uninstall fix

madchills:
These few simple changes make it possible to get a working environment with the powershell script, and allow for correct upgrading.


basepi:
Will this handle deletions of files if we rename a file between versions? The `ifdiff` piece makes me wonder.

basepi:
We'll deal with renamed files in the future.

PR #120, opened 2017-07-28
Adding pulsar config name in splunk data

anuragpaliwal80:
Linux:
screen shot 2017-07-27 at 4 32 01 pm
Windows:
screen shot 2017-07-28 at 2 23 22 pm


anuragpaliwal80:
FYI... This will work fine for single config file case but I think we are good for now as we are using single config file only. And we can take multiple file case depending upon it's priority in future.

basepi:
Looks great. First config is fine for now, if we need it for multiple when we start having team-specific configs we can address it then.

Issue #99, opened 2017-07-20
Canary FIM events


basepi:
It's sometimes hard to detect whether FIM isn't working on a host, or if a host is just really quiet. We want to add a scheduled job once per day to create and delete a temporary file to trigger a FIM event, so all hosts will have daily FIM data.

basepi:
https://jira.corp.adobe.com/browse/CSTTEAM-41394

PR #125, opened 2017-07-31
Add canary FIM events

basepi:
Fixes https://github.com/hubblestack/hubble/issues/99


basepi:
Will throw a create/delete event at startup and every 24 hours afterwards. Tested and working on linux. Will work on Windows once we add watches for the hubble dir.

PR #129, opened 2017-08-01
Top pulsar

rashmip29:


hubblestack/extmods/modules/pulsar.py
```diff
@@ -76,7 +76,7 @@ def _get_notifier():
     return __context__['pulsar.notifier']
 
 
-def process(configfile='salt://hubblestack_pulsar/hubblestack_pulsar_config.yaml',
+def process(configfile,
```
basepi: Not sure why this default got removed. It needs to stay in.

hubblestack/extmods/modules/pulsar.py
```diff
@@ -395,3 +395,44 @@ def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
             for k in upd:
                 dest[k] = upd[k]
         return dest
+
+#TODO:change paths for top.pulsar and config yaml to 'salt://hubblestack_pulsar/top.file'
+
+def top(topfile='/root/myhubblefiles/topfile/top.pulsar',
```
basepi: Just reiterating your TODO. The default path should probably be `salt://hubblestack_pulsar/top.pulsar`. On the windows side, it will be `salt://hubblestack_pulsar/win_top.pulsar`.

hubblestack/extmods/modules/pulsar.py
```diff
@@ -395,3 +395,44 @@ def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
...
+        debug=None):
+
+    results_pulsar = []
+
+    configs = get_top_data(topfile)
+
+    if not isinstance(configs, list):
+        configs = configs.split(',')
```
basepi: I don't think you need this `if` statement, because you know you'll always be returning a list from `get_top_data()`. I expect it was copy pasta.

hubblestack/extmods/modules/pulsar.py
```diff
@@ -395,3 +395,44 @@ def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
...
+
+    configs = ['/root/myhubblefiles/topfile/' + config + '.yaml'
+               for config in configs]
```
basepi: Here, you'll want to prepend `salt://hubblestack_pulsar/` and append `.yaml` instead of the absolute path you're using.

hubblestack/extmods/modules/pulsar.py
```diff
@@ -395,3 +395,44 @@ def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
...
+
+    results_pulsar.extend(process(configs, verbose=False))
```
basepi: I think here, since `results_pulsar` is empty, you can just do a straight return:

```
return process(configs, verbose=verbose)
```

Note that I have `verbose=verbose` -- the `top()` function should take a verbose argument and pass it through to `process()`.

PR #133, opened 2017-08-04
Add option to extract fields at index time in Splunk returners

mew1033:
Also hard code a few fields that we know would be useful.

Ref: https://dev.splunk.com/view/event-collector/SP-CAAAFB6

### DIFF
--- a/hubblestack/extmods/returners/splunk_nebula_return.py
+++ b/hubblestack/extmods/returners/splunk_nebula_return.py
@@ -75,6 +75,14 @@ def returner(ret):
         timeout = opts['timeout']
         custom_fields = opts['custom_fields']
 
+        # Set up the fields to be extracted at index time. The field values must be strings.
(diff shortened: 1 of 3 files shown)


basepi:
Looks perfect.

hubblestack/extmods/returners/splunk_nebula_return.py
```diff
@@ -132,6 +140,14 @@ def returner(ret):
                             payload.update({'sourcetype': opts['sourcetype']})
                         payload.update({'event': event})
 
+                        # Potentially add metadata fields:
+                        fields = {}
+                        for item in index_extracted_fields:
+                            if item in payload['event'] and not isinstance(payload['event'], (list, dict, tuple)):
```
basepi: I didn't know you could pass tuples into `isinstance`. Cool!

PR #134, opened 2017-08-04
Fossam dockeromnibus

fossam:
images from these dockerfiles are over 4.5 Gb per image.


basepi:
All of these so far appear to be working great! Looking forward to the cent6 and debian7 ones! You're awesome!

tsharpe:
That is impressive. Are these tested? Because @cedwards mentioned this to me on the drive home. I think I know what I going to do this weekend.

basepi:
They are tested, but there are known issues that I think @fossam has resolved but not pushed yet. On the list:

- pulling the hubble config from the Dockerfile directory if present
- adding symlinks to /usr/bin for hubble binary

PR #135, opened 2017-08-04
Fix nebula to prefer our bundled osqueryi if present

basepi:


basepi:
@fossam good?

basepi:
Ah, yes, I forgot you added that. I'll make that change and test it while you finish testing your docker files. The one problem is that this grain will not prefer our bundled path, it defaults to the one on the system PATH. So I'll change it to prefer `/opt/osquery/osqueryi` first.

basepi:
Ooh, except I haven't tested on Windows, and may not have time to test on Windows.....I may leave it for the moment and wait until we can test better.

basepi:
Meh, nothing you did is any different than what I did. If it breaks on windows we'll do a 2.2.1. No big deal.

basepi:
Fixed in #137

PR #138, opened 2017-08-07
Rev to 2.2.0

basepi:


basepi:
@fossam Note the changes to the dockerfiles. I realized that our tag format wasn't completely compatible with your dockerfile, so the "version" is 2.2.0 and I add the `v` on the checkout so it checks out the release tag properly (`v2.2.0`)

We may want to split out "tag version" vs "source version" so that we can specify the version and the SHA/tag separately in the future. This will do for now.

PR #154, opened 2017-08-11
Fossam dockerwork

fossam:
Two items in these commits covered:
* Dockerfile for debian 7
* working pygit2 in debian 8 package


basepi:
Awesome!

PR #158, opened 2017-08-15
Documentation

whackyhack:
Instructions about release packages and local FS profile


basepi:
This is great! I am planning some other big docs updates, so stay tuned!

PR #164, opened 2017-08-21
Deprecate show_profile

whackyhack:
Remove deprecated show_profile from example schedule


basepi:
Thanks for catching that.

PR #169, opened 2017-08-25
changes added for nebula topfile

rashmip29:


hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -74,11 +74,19 @@ def queries(query_group,
         salt '*' nebula.queries hour verbose=True
         salt '*' nebula.queries hour pillar_key=sec_osqueries
     '''
+
+    if isinstance(query_file, list):
+        query_file = _dict_update(query_file,
+                                  yaml.safe_load(f),
+                                  recursive_update=True,
+                                  merge_lists=True)
+
```
basepi: Can you look over this code again? Looks like you have an uninstantiated variable (`f`), and you're trying to `_dict_update` on something that's not a dict (`query_file`, which is a list, and probably not a good final destination for the config).

You also need to do the `__salt__['cp.cache_file']()` for each file before you'll have them for reading anyway.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -74,11 +74,52 @@ def queries(query_group,
...
+                    with open(path, 'r') as f:
+                        query_file = _dict_update(new_config,
+                                                  yaml.safe_load(f),
+                                                  recursive_update=True,
+                                                  merge_lists=True)
+                else:
+                    log.error('Path {0} does not exist or is not a file'.format(path))
+        else:
+            log.error('Nebula beacon \'paths\' data improperly formatted. Should be list of paths')
+
```
basepi: Looks like we have a bunch of extra pulsar stuff that we don't need here. We don't need to store the config in a global or worry about config staleness, because the most often nebula runs is every 15 minutes, so we can just read in the config fresh every time. Basically, all you need to do here is iterate over the list of config files, and use dictupdate to merge each of them into a starting empty dictionary, in order.

I recommend you use pulsar as a guide for when you get stuck, but I recommend you write this piece of code from scratch. It's different enough from pulsar to be worth keeping away from copy paste where possible. If you don't understand what needs to happen well enough to envision the logic, ping me and we can walk through it again.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -74,11 +74,32 @@ def queries(query_group,
...
+            if os.path.isfile(file):
+                with open(file, 'r') as f:
+                    config = _dict_update(config,
+                                         yaml.safe_load(f),
+                                         recursive_update=True,
+                                         merge_lists=True)
+        if salt.utils.is_windows():
+            query_file = "salt://hubblestack_nebula/hubblestack_nebula_win_queries.yaml"
+        else:
+            query_file = 'salt://hubblestack_nebula/hubblestack_nebula_queries.yaml'
```
basepi: this if/else statement isn't needed here, because if we want to run those files, they'll be in the list above and we will have already processed them

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -74,11 +74,32 @@ def queries(query_group,
...
+
+        query_file = __salt__['cp.cache_file'](query_file)
+        with open(query_file,'w') as yaml_file:
+            yaml.dump(config, yaml_file, default_flow_style=False)
```
basepi: There are a couple of problems here. I can sort of see what you were going for -- let's write it out to a file the logic below doesn't have to be changed. However, it won't work, primarily because the logic below also calls `cp.cache_file` (on line 143/164), which would overwrite your `yaml.dump`ed changes.

It's also not necessary. If we move this list handling logic further below, we can do it all in memory and never have to write to disk.

Here's the basic logic, in pseudocode:

```
if query_file is None:
    <set it to hubblestack_nebula_queries or hubblestack_nebula_win_queries depending on windows/linux>
if query_file is not a list:
    query_file = [query_file]  # This way we can treat all runs the same. Normal runs will just be a list of one file
config = {}
for f in query_file:
    <merge each file into config>
```

At the end of this logic we'll have a `config` variable that is a combination of one or more query files, and we can do the normal osquery run on that data, without ever having to write to disk.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -138,14 +154,11 @@ def queries(query_group,
             else:
                return None
 
-
     orig_filename = query_file
-    query_file = __salt__['cp.cache_file'](query_file)
+
     if query_file is None:
         log.error('Could not find file {0}.'.format(orig_filename))
         return None
```
basepi: This error logic is no longer necessary here -- can we move it up into your `for` loop? We want to log an error in case any of the files passed into the list cannot be found.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -74,11 +74,27 @@ def queries(query_group,
...
     if query_file is None:
         if salt.utils.is_windows():
             query_file = 'salt://hubblestack_nebula/hubblestack_nebula_win_queries.yaml'
         else:
             query_file = 'salt://hubblestack_nebula/hubblestack_nebula_queries.yaml'
+
+    if not isinstance(query_file, list):
+        query_file = [query_file]
+
+    for file in query_file:
```
basepi: Can you use a different variable name than `file`? `file` is a reserved built-in in python and is being [shadowed](https://en.wikipedia.org/wiki/Variable_shadowing) here.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -138,14 +154,11 @@ def queries(query_group,
...
-    query_file = __salt__['cp.cache_file'](query_file)
+
     if query_file is None:
         log.error('Could not find file {0}.'.format(orig_filename))
         return None
-    with open(query_file, 'r') as fh:
-        query_data = yaml.safe_load(fh)
 
     if not isinstance(query_data, dict):
         raise CommandExecutionError('Query data is not formed as a dict {0}'
```
basepi: Another instance of error handling that we can move up into your for loop. We should probably check if a given file's data is formed as a dict before we attempt to merge it with our query_data.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -74,11 +74,34 @@ def queries(query_group,
...
             query_file = 'salt://hubblestack_nebula/hubblestack_nebula_queries.yaml'
+
+    if not isinstance(query_file, list):
+        query_file = [query_file]
+
+    for fh in query_file:
+        if 'salt://' in fh:
+            fh = __salt__['cp.cache_file'](fh)
+        if fh is None:
+            log.error('Could not find file {0}.'.format(fh))
```
basepi: `fh` will be `None` here. Can you store the old fh (before cp.cache_file) for logging in this error message?

rashmip29: Done :)

basepi:
Everything is good except that there were changes made to nebula_osquery.py on the develop branch that are conflicting with your changes.

I recommend you do this:

```
git fetch upstream
git merge upstream/develop
# resolve conflicts and commit
git push origin top-nebula
```

Assuming your fork is `origin` and the parent repo is `upstream`.

basepi:
Weirdly, your merge from develop did not resolve the conflicts. You might want to revert your merge (it looks like it also added some whitespace changes and so forth) and rebase instead.

You could also start with a fresh branch from upstream/develop and manually apply your changes there. Might actually be the easiest way to get it clean again.

basepi:
Closing in favor of https://github.com/hubblestack/hubble/pull/176

PR #170, opened 2017-08-28
Two changes

madchills:
Added logging when path specied in yaml doesnt exist.  Changed blacklist from 'or' to 'list'


hubblestack/extmods/modules/win_pulsar.py
```diff
@@ -28,6 +28,8 @@
 CONFIG = None
 CONFIG_STALENESS = 0
 
+__version__ = 'v2017.8.3'
+
```
basepi: hubble-salt copy pasta

PR #192, opened 2017-09-18
Adding mount and systemctl modules

anuragpaliwal80:
* Adding mount and systemctl modules for hubble


basepi:
@anuragpaliwal80 We were looking back at this code and noticed that it mixed tabs and spaces and has a couple other smaller issues. Make sure you're testing before you commit because I'm pretty sure Python will throw a syntax error if you mix tabs/spaces.

basepi:
(I'll be submitting a PR shortly with the fixes.)

PR #193, opened 2017-09-18
Hubble readme

rashmip29:


README.md
```diff
@@ -1,100 +1,240 @@
...
+Installs a hubble "binary" into `/usr/bin/`.
 
-Installs a `hubble` "binary" into `/usr/bin/`.
+A config template has been placed in `/etc/hubble/hubble`. Modify it to your specifications and needs. You can do `hubble -h` to see the available options.
 
-## Building standalone packages (CentOS)
+The first two commands you should run to make sure things are set up correctly are `hubble --version` and `hubble test.ping`.
 
-```bash
+### Building standalone packages (CentOS)
```
basepi: Can you please remove these two sections that use the build_rpms.sh and build_debs.sh scripts? I don't think they're working at this point so we want to point people to the dockerfiles.

README.md
```diff
@@ -1,100 +1,222 @@
...
-```bash
-sudo yum install git -y
-git clone https://github.com/hubblestack/hubble
-cd hubble/pkg
-./build_rpms.sh  # note the lack of sudo, that is important
+### Buidling Hubble packages through Dockerfile
+Dockerfile aims to make building Hubble v2 packages easier. Dockerfiles can be found at `hubblestack/hubble/pkg`. 
+To build an image
+```sh
+1. copy pkg/scripts/pyinstaller-requirements.txt to directory with this Dockerfile
```
basepi: This step is no longer necessary. Just the docker build. Can you also mention something about `cd`ing into the directory for the distribution you want to build?

README.md
```diff
@@ -1,100 +1,222 @@
...
-## Building standalone packages (Debian)
-
-```bash
-sudo yum install git -y
-git clone https://github.com/hubblestack/hubble
-cd hubble/pkg
-./build_debs.sh  # note the lack of sudo, that is important
+To run the container
+```sh
+docker run -it --rm <image_name>
```
basepi: This is not the full command from the dockerfile, and won't work. We need the `/data` mount that the full command from the dockerfile has.

README.md
```diff
@@ -1,100 +1,222 @@
...
-```
-
-Packages will be in the `hubble/pkg/dist/` directory. The only difference
-between the packages is the inclusion of `/etc/init.d/hubble` for el6 and
-the inclusion of a systemd unit file for el7. There's no guarantee of glibc
-compatibility.
-
-## Building standalone packages (Debian)
+### Buidling Hubble packages through Dockerfile
+Dockerfile aims to build the Hubble v2 packages easier. Dockerfiles for the distribution you want to build can be found at the path `hubblestack/hubble/pkg`. For e.g. for centos6 distribution the dockerfile is at the path `hubblestack/hubble/pkg/centos6/`
```
basepi: Can you remove the `hubblestack/hubble/` from these paths? Any relative paths assume the root of the repo, so just `pkg/` and `pkg/centos6/` should suffice. You also have an extra "For" at the beginning of the last sentence.

README.md
```diff
@@ -1,100 +1,222 @@
...
-```bash
-sudo yum install git -y
-git clone https://github.com/hubblestack/hubble
-cd hubble/pkg
-./build_debs.sh  # note the lack of sudo, that is important
+To build an image
+```sh
+docker build -t <image_name> 
+```
+To run the container
```
basepi: "To run the container (which will output the package file in your current directory):"

PR #194, opened 2017-09-18
Add splunk log handler

basepi:
Turned off by default, have to set `hubblestack:splunklogging` to True. Uses the existing splunk returner config.

### DIFF
--- a/hubblestack/splunklogging.py
+++ b/hubblestack/splunklogging.py
@@ -60,83 +60,79 @@ class SplunkHandler(logging.Handler):
     Log handler for splunk
     '''
     def __init__(self):
-        try:
-            super(SplunkHandler, self).__init__()
+        super(SplunkHandler, self).__init__()
 
-            self.opts_list = _get_options()
-            self.clouds = get_cloud_details()
-            self.endpoint_list = []
+        self.opts_list = _get_options()
+        self.clouds = get_cloud_details()
+        self.endpoint_list = []
 
-            for opts in self.opts_list:
-                http_event_collector_key = opts['token']
-                http_event_collector_host = opts['indexer']
-                http_event_collector_port = opts['port']
-                hec_ssl = opts['http_event_server_ssl']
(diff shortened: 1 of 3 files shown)


hubblestack/splunklogging.py
```diff
@@ -0,0 +1,315 @@
...
+        splunk_opts['token'] = __salt__['config.get']('hubblestack:nova:returner:splunk:token').strip()
+        splunk_opts['indexer'] = __salt__['config.get']('hubblestack:nova:returner:splunk:indexer')
+        splunk_opts['port'] = __salt__['config.get']('hubblestack:nova:returner:splunk:port', '8088')
+        splunk_opts['index'] = __salt__['config.get']('hubblestack:nova:returner:splunk:index')
+        splunk_opts['custom_fields'] = __salt__['config.get']('hubblestack:nova:returner:splunk:custom_fields', [])
+        splunk_opts['sourcetype'] = __salt__['config.get']('hubblestack:nova:returner:splunk:sourcetype')
+        splunk_opts['http_event_server_ssl'] = __salt__['config.get']('hubblestack:nova:returner:splunk:hec_ssl', True)
+        splunk_opts['proxy'] = __salt__['config.get']('hubblestack:nova:returner:splunk:proxy', {})
+        splunk_opts['timeout'] = __salt__['config.get']('hubblestack:nova:returner:splunk:timeout', 9.05)
+        splunk_opts['index_extracted_fields'] = __salt__['config.get']('hubblestack:nova:returner:splunk:index_extracted_fields', [])
```
mew1033: Looks like the old style config is pointing to the nova config here.

basepi: This is actually on purpose. I didn't want to add a new block of old style config (since I'd like to deprecate that config eventually), and nova was the most likely block to exist.

That said, I think I'm going to require new style config for this one. Since it has to be opt in anyway, they should be updating their config style.

PR #197, opened 2017-09-19
Add splunkindex grain

### LINKED ISSUE #107: Add splunk index name to grains
This is a pretty splunk-specific feature but would be very useful for the topfile features in #106. Team-specific targeting and the like.

mew1033:
Not sure if this is how you were thinking of adding the grain, but this should fix #107.


mew1033:
Also, this only supports the new style config.

basepi:
Have you tested this? I want to make sure there's no gotchas in terms of the config.get usage.

fossam:
Besides the code, we have a choice here - we are going to depend on this grain on top files to do  matching, so the top files may have to always be specific to returner or at least depend on it. The choice is to either to accept that dependency, or have a more generic grain that will leave room for other returners.

I personally would lean towards more generic approach. For example, for testing purposes would be great to have a returner that just writes to plain files and it would be neat not to change top files for every test or use case like that. However, these are my first on thoughts on the matter and it could be that I am overlooking relevant details that make this less of an issue.

basepi:
I mean, we're only going to use the index for targeting on our internal top.sls file. And all internal hosts must use the splunk returner. So I think it's OK to introduce that dependency.

hubblestack/extmods/grains/splunkindex.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def splunkindex():
+    '''
+    Return splunk index from config file in grain
+    '''
+    # Provides:
+    #   splunkindex
+    grains = {}
+    index = __salt__['config.get']('hubblestack:returner:splunk:index', default='unknown')
```
basepi: Let's either default this to `None`, or just not define it at all if we can't find that value.

fossam:
```
if __salt__['config.get']('hubblestack:returner:splunk:index')
        grains['splunkindex'] = index
return grains
```

mew1033:
I'm going to change this up a bit to support custom grains in the config.

mew1033:
@basepi @fossam Alright, I changed this to allowing any custom grains defined in config.

hubblestack/extmods/grains/configgrains.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+
```
basepi: Can you just add a docstring with sample config? This looks great otherwise.

mew1033: Done

PR #204, opened 2017-09-25
Fix command injection vulnerabilities in misc.py

basepi:
@anuragpaliwal80 Can you share these changes with your team, and also review them for me? I did not test them but I'm pretty sure I kept the logic correct.

`python_shell=True` should be used as a last resort, and never with a command which has user arguments in it, or else you risk command injection. In many places, you were using `python_shell=True` just so you could do `; echo $?`. This is unnecessary, as salt has `cmd.retcode` which just returns the retcode. Similarly, you have other places where you pipe output to /dev/null -- this is also unnecessary, as you have all of these options in salt:

- `cmd.run_all` (which gives you a dictionary with stdout, stderr, retcode, etc)
- `cmd.run` (combines stdout and stderr into a single string)
- `cmd.run_stdout` (only returns stdout)
- `cmd.run_stderr` (only returns stderr)

After this release I'll likely go through and remove `_execute_shell_command` and have each function call `__salt__[cmd.<func>]()` itself. `python_shell=True` should be an exception, not the default.

### DIFF
--- a/hubblestack/files/hubblestack_nova/misc.py
+++ b/hubblestack/files/hubblestack_nova/misc.py
@@ -906,7 +906,7 @@ def mail_conf_check(reason=''):
     mail_addresses = mail_addresses.split(',') if mail_addresses != "" else []
     mail_addresses = map(str.strip, mail_addresses)
     invalid_addresses = list(set(mail_addresses) - set(valid_addresses))
-    
+
     return str(invalid_addresses) if invalid_addresses != [] else True
 
 def check_if_any_pkg_installed(args):
--- a/hubblestack/files/hubblestack_nova/misc.py
+++ b/hubblestack/files/hubblestack_nova/misc.py
@@ -313,26 +313,29 @@ def root_is_only_uid_0_account(reason=''):
     return True if result.strip() == 'root' else result
 
 
-def test_mount_attrs(mount_name,attribute,check_type='hard'):
+def test_mount_attrs(mount_name, attribute, check_type='hard'):
     '''
     Ensure that a given directory is mounted with appropriate attributes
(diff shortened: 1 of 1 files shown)


hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -313,26 +313,29 @@ def root_is_only_uid_0_account(reason=''):
...
     Ensure that a given directory is mounted with appropriate attributes
     If check_type is soft, then in absence of volume, True will be returned
     If check_type is hard, then in absence of volume, False will be returned
     '''
     # check that the path exists on system
-    command = 'test -e ' + mount_name + ' ; echo $?'
-    output = _execute_shell_command( command)
-    if output.strip() == '1':
+    command = 'test -e ' + mount_name
+    results = __salt__['cmd.run_all'](command)
```
fossam: It looks that much of the safety of cmd.run* functions depend on __python_shell_default_ function (https://github.com/saltstack/salt/blob/develop/salt/modules/cmdmod.py). We need to make sure that python_shell is indeed False, otherwise it's not much different what we had before.

basepi: It is. If it weren’t, Salt would be full of vulnerabilities. The reason that function exists is so we can set it to True if someone is calling `cmd.run` as the top level job, from the command line. In that case it means they already have direct access to `cmd.run` and so we don’t need to worry about injection. (If they are malicious they have all the access they need anyway). 

`__pub_jid` is a special argument that gets passed to the top level call in salt, so the cmd module uses it to detect that top level call. We never pass that argument in hubble so we’re safe.

PR #218, opened 2017-10-03
Unittest

rashmip29:


tests/unittests/test_nebula_osquery.py
```diff
@@ -0,0 +1,119 @@
...
+        def cp_cache_file(queryFile):
+            return 'tests/unittests/resources/hubblestack_nebula_queries.yaml'
+        __salt__ = {}
+        __salt__['cp.cache_file'] = cp_cache_file
+        hubblestack.extmods.modules.nebula_osquery.__salt__ = __salt__
+        hubblestack.extmods.modules.nebula_osquery.__grains__ = {}
+        def cmd_run_all(cmd):
+            return {'retcode': 0, 'pid': 3478, 'stdout':\
+'[{"build":"","codename":"xenial","major":"16","minor":"4","name":"Ubuntu","patch":"","platform":"ubuntu","platform_like":"debian","query_time":"1500395829","version":"16.04.2 LTS (Xenial Xerus)"}]',\
+ 'stderr': ''}
```
basepi: More indentation

tests/unittests/test_nebula_osquery.py
```diff
@@ -0,0 +1,119 @@
...
+        hubblestack.extmods.modules.nebula_osquery.__grains__ = {}
+        def cmd_run_all(cmd):
+             return {'retcode': 0, 'pid': 3478, 'stdout': \
+'[{"build":"","codename":"xenial","major":"16","minor":"4","name":"Ubuntu","patch":"","platform":"ubuntu","platform_like":"debian","query_time":"1500395829","version":"16.04.2 LTS (Xenial Xerus)"}]', \
+'stderr': ''}
+        __salt__['cmd.run_all'] = cmd_run_all
+        hubblestack.extmods.modules.nebula_osquery.__salt__ = __salt__
+        var = hubblestack.extmods.modules.nebula_osquery.queries(query_group, query_file,verbose=False,report_version_with_day=True)
+        hubblestack.extmods.modules.nebula_osquery.__salt__ = {}
+        assert var != 0
```
basepi: This is another place where it will always succeed. nebula.queries never returns an integer. This continues throughout this file so I'm not going to leave a comment on every one.

tests/unittests/test_pkg_nova.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def test_audit_for_Success(self):
+       val = {}
+       data_list = [('ubuntu-1604-level-1-scored-v1-0-0', {'stat': {'passwd_owner_group': {'data': {'Ubuntu-16.04': \
+                   [{'/etc/passwd': {'gid': 0, 'tag': 'CIS-12.4', 'group': 'root', 'uid': 0, 'user': 'root'}}]}, \
+                     'description': 'Verify User/Group Ownership on /etc/passwd'}}})]
+       __tags__ = {'CIS-6.5': [{'tag': 'CIS-6.5', 'name': 'ntp', 'nova_profile': 'ubuntu-1604-level-1-scored-v1-0-0', 'type': 'whitelist', \
+                   'module': 'pkg', 'description': 'Configure Network Time Protocol (NTP)'}]}
+       val = hubblestack.files.hubblestack_nova.pkg.audit(data_list, __tags__, debug=False)
+       assert val['Success'] != 0
```
basepi: This file has similar problems to the grep tests in that _merge_yaml doesn't actually check with "real" data, and most of the other tests are comparisons with 0, which don't make sense.

tests/unittests/test_stat_nova.py
```diff
@@ -0,0 +1,71 @@
...
+        profile = None
+        val = hubblestack.files.hubblestack_nova.stat_nova._merge_yaml(ret, data, profile)
+        assert val['test1'] == 'val1'
+
+    def test_merge_yaml_different_val_for_data(self):
+        ret = {}
+        data = {'test2':'val2'}
+        profile = None
+        val = hubblestack.files.hubblestack_nova.stat_nova._merge_yaml(ret, data, profile)
+        assert val['stat'] != 0
```
basepi: More 0 comparisons that don't make sense considering the data in question.

tests/unittests/test_stat_nova.py
```diff
@@ -0,0 +1,71 @@
...
+    def test_audit(self):
+        val = {}
+        data_list = [('ubuntu-1604-level-1-scored-v1-0-0', {'stat': {'passwd_owner_group': {'data': {'Ubuntu-16.04': [{'/etc/passwd': \
+                    {'gid': 0, 'tag': 'CIS-12.4', 'group': 'root', 'uid': 0, 'user': 'root'}}]}, 'description': 'Verify User/Group Ownership on /etc/passwd'}}})]
+        __tags__ = '*'
+        hubblestack.files.hubblestack_nova.stat_nova.__salt__ = {}
+        try:
+            val = hubblestack.files.hubblestack_nova.stat_nova.audit(data_list, __tags__, debug=False)
+        except KeyError:
+            assert val != 0
```
basepi: Was this assert supposed to be in the except block? Are you expecting a keyerror?

Also, it's another 0 comparison and should probably be fixed anyway.

tests/unittests/test_pulsar.py
```diff
@@ -0,0 +1,161 @@
...
+        verbose = False
+        def config_get(value, default):
+            return default
+        __salt__ = {}
+        __salt__['config.get'] = config_get
+        hubblestack.extmods.modules.pulsar.__salt__ = __salt__
+        hubblestack.extmods.modules.pulsar.__opts__ = {}
+        hubblestack.extmods.modules.pulsar.process(configfile,verbose)
+        hubblestack.extmods.modules.pulsar.__salt__ = {}
+        assert hubblestack.extmods.modules.pulsar.CONFIG != 0
```
basepi: Another 0 comparison. This file has a bunch of them as well. In this particular case you probably just want to check if it's empty.

PR #227, opened 2017-10-17
Adding Azure SubscriptionId details

anuragpaliwal80:


basepi:
@mew1033 Are you OK with this new index extracted field on our side as well?

PR #229, opened 2017-10-20
Fixing few cases in win_secedit module

anuragpaliwal80:


basepi:
@madchills can you please review this?

PR #236, opened 2017-10-24
Update osquery to 2.9.2 + add rpm-build to rh reqs

basepi:


fossam:
Hah, I was just doing this and fixing rpm-build. :+1:

PR #237, opened 2017-10-26
Fixing a bug in win_reg module to distinguish False from 0

yassingh:


hubblestack/files/hubblestack_nova/win_reg.py
```diff
@@ -67,7 +67,7 @@ def audit(data_list, tags, debug=False, **kwargs):
                     current = _find_option_value_in_reg(reg_dict['hive'], reg_dict['key'], reg_dict['value'])
                     if isinstance(current, dict):
                         tag_data['value_found'] = current
-                        if False in current.values():
```
yassingh: line 70 evalues to True even when current.values() contains an integer 0. This is not expected behaviour.
line 70 should evaluate to True only if current.values() contains a boolean False

hubblestack/files/hubblestack_nova/win_reg.py
```diff
@@ -195,8 +195,8 @@ def _find_option_value_in_reg(reg_hive, reg_key, reg_value):
         for sid in key_list:
             if len(sid) <= 15 or '_Classes' in sid:
                 continue
-            reg_key = reg_key.replace('<SID>', sid)
```
yassingh: this loop only works for first sid. Once "<SID>" is replaced by the first sid, it is never replaced again in further loops.

basepi:
@madchills Added you as a reviewer. Please merge if this looks good.

PR #241, opened 2017-11-03
Remove empty fields from splunk return

mew1033:


fossam:
Let us think about this over the weekend. May have impacts where none were intended.

fossam:
Additional query with a version field will make it possible to distinguish between up to date and out of date profiles. For now the query will have to be manged manually until a more dynamic solution is found.

PR #246, opened 2017-11-07
Add dev packaging

basepi:
I put it in its own directory to make merging between develop/master
less nightmarish.

Only four lines actually changed. If you look at the second commit, you can see that diff much more easily than looking at the entire PR diff.

### DIFF
--- a/pkg/dev/amazonlinux2016.09/Dockerfile
+++ b/pkg/dev/amazonlinux2016.09/Dockerfile
@@ -91,8 +91,8 @@ RUN yum install -y ruby ruby-devel rpmbuild rpm-build rubygems gcc make \
 #pyinstaller start
 #commands specified for ENTRYPOINT and CMD are executed when the container is run, not when the image is built
 #use the following variables to choose the version of hubble
-ENV HUBBLE_CHECKOUT=v2.2.10
-ENV HUBBLE_VERSION=2.2.10
+ENV HUBBLE_CHECKOUT=develop
+ENV HUBBLE_VERSION=2.2.10_develop
 ENV HUBBLE_GIT_URL=https://github.com/hubblestack/hubble.git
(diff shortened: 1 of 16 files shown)


pkg/dev/amazonlinux2016.09/Dockerfile
```diff
@@ -0,0 +1,148 @@
...
+ENV HUBBLE_SRC_PATH=/hubble_src
+ENV _HOOK_DIR="./pkg/"
+ENV _BINARY_LOG_LEVEL="INFO"
+ENV _INCLUDE_PATH=""
+ENV LD_LIBRARY_PATH=/lib:/lib64:/usr/lib:/usr/lib64:/usr/local/lib:/usr/local/lib64
+RUN git clone "$HUBBLE_GIT_URL" "$HUBBLE_SRC_PATH" \
+ && cd "$HUBBLE_SRC_PATH" \
+ && git checkout "$HUBBLE_CHECKOUT" \
+ && cp -rf "$HUBBLE_SRC_PATH" /hubble_build \
+ && echo __version__ = \"`git describe --tags | sed 's/v//'`\" > /hubble_build/hubblestack/__init__.py
```
fossam: Would a string after version need to be in (single) quotes? Like
`version = '<string>'`

basepi: Python accepts either. I used double quotes because it was easier to deal with on the shell.

basepi:
The problem is that to get `git describe` to work correctly, I have to merge back to develop from master after tagging, or `git describe` can't see the tag. So, any commits we stick on the master branch would then come back to the develop branch, and have to be reverted there. Vicious cycle. I decided a second directory was easier. If it gets to be a pain we can revisit it.

PR #249, opened 2017-11-29
Use paramter value rather than hard coded value in win_pulsar

anuragpaliwal80:


basepi:
This looks fine to me but I want @madchills to review as well.

Issue #248, opened 2017-11-29
Audit shows '[ERROR ] Problem occurred trying to invalidate hash cach for azurefs'


anuragpaliwal80:
I am not sure if this has been logged already so on a safer side logging a new ticket.

```
[root@ip-10-249-71-213 ~]# hubble hubble.audit
[ERROR ] Problem occurred trying to invalidate hash cach for azurefs
Traceback (most recent call last):
File "/opt/hubble/hubble-libs/hubblestack/extmods/fileserver/azurefs.py", line 268, in update
shutil.rmtree(hash_cachedir)
File "shutil.py", line 239, in rmtree
File "shutil.py", line 237, in rmtree
OSError: [Errno 2] No such file or directory: '/var/cache/hubble/azurefs/hashes'
[ERROR ] Problem occurred trying to invalidate hash cach for azurefs
Traceback (most recent call last):
File "/opt/hubble/hubble-libs/hubblestack/extmods/fileserver/azurefs.py", line 268, in update
shutil.rmtree(hash_cachedir)
File "shutil.py", line 239, in rmtree
File "shutil.py", line 237, in rmtree
OSError: [Errno 2] No such file or directory: '/var/cache/hubble/azurefs/hashes'
[ERROR ] Problem occurred trying to invalidate hash cach for azurefs
Traceback (most recent call last):
File "/opt/hubble/hubble-libs/hubblestack/extmods/fileserver/azurefs.py", line 268, in update
shutil.rmtree(hash_cachedir)
File "shutil.py", line 239, in rmtree
File "shutil.py", line 237, in rmtree
OSError: [Errno 2] No such file or directory: '/var/cache/hubble/azurefs/hashes'
[ERROR ] Problem occurred trying to invalidate hash cach for azurefs
Traceback (most recent call last):
File "/opt/hubble/hubble-libs/hubblestack/extmods/fileserver/azurefs.py", line 268, in update
shutil.rmtree(hash_cachedir)
File "shutil.py", line 239, in rmtree
File "shutil.py", line 237, in rmtree
OSError: [Errno 2] No such file or directory: '/var/cache/hubble/azurefs/hashes'
[ERROR ] Problem occurred trying to invalidate hash cach for azurefs
Traceback (most recent call last):
File "/opt/hubble/hubble-libs/hubblestack/extmods/fileserver/azurefs.py", line 268, in update
shutil.rmtree(hash_cachedir)
File "shutil.py", line 239, in rmtree
[...]

PR #251, opened 2017-11-30
Fix errors for azurefs when hash_cachedir doesn't exist

basepi:
Closes https://github.com/hubblestack/hubble/issues/248


basepi:
@fossam It would suppress those spurious errors, but it would also suppress real errors if the file was not able to be removed for some other reason. I just wanted to suppress errors from the file not being present (since that's not actually an error condition).

PR #258, opened 2017-12-11
Adding the ability to parse match_type

madchills:
this allows for an extra value in the yaml that will determine if the value must be equal or can be greater than or less than.  match_type


basepi:
Should we also add `<=` and `>=` or are the three that are already there sufficient?

madchills:
What is there is sufficient.  I just change the number to be 1 less than what the document says it needs to be or more (84 or more would translate to more than 83 in my yaml)

PR #259, opened 2017-12-11
Win_pulsar NTFS Change Journal

madchills:
A complete redesign of win_pulsar.  Instead of using Windows Folder Auditing and the event log, it instead uses the NTFS Change Journal.  Currently it uses FSUTIL to query the journal, and parses the output into a list of dictionaries.  The it filters out what we don't need / want to see


basepi:
Looks like this is conflicting, probably due to the pep8 changes that were made recently. Resolving the merge conflict is probably going to be a nightmare, so instead I recommend you copy the new win_pulsar.py out of your git repo to somewhere else, then do the `git pull`, and to resolve the conflicts you just copy your external copy back in and commit. That should resolve everything so we can merge.

hubblestack/extmods/modules/win_pulsar.py
```diff
@@ -31,70 +30,75 @@
...
+    if '2012' not in win_version and '2016' not in win_version:
+        return False, 'This module only works with Server 2012 (Win8) or higher'
     return __virtualname__
 
-
 def process(configfile='salt://hubblestack_pulsar/hubblestack_pulsar_win_config.yaml',
             verbose=False):
     '''
-    Watch the configured files
+    Watch the confugred files
```
basepi: s/confugred/configured

hubblestack/extmods/modules/win_pulsar_winaudit.py
```diff
@@ -0,0 +1,554 @@
...
+
+import salt.ext.six
+import salt.loader
+
+log = logging.getLogger(__name__)
+DEFAULT_MASK = ['ExecuteFile', 'Write', 'Delete', 'DeleteSubdirectoriesAndFiles', 'ChangePermissions',
+                'TakeOwnership'] #ExecuteFile Is really chatty
+DEFAULT_TYPE = 'all'
+
+__virtualname__ = 'pulsar'
```
basepi: Because we're keeping this old module around, it's going to try to load both the old and the new modules under the same `__virtualname__`. We should change one of them to avoid this conflict.

PR #260, opened 2017-12-14
Add periodic grains refresh and session/host UUIDs

basepi:
This deserves a fair amount of review. @fossam to make sure this hits all his bullet points WRT unique IDs for sessions and hosts. @madchills for any problems you see regarding the python.

Defaulting to hourly grains refreshes, and sending those to splunk as well if configured with splunk_logging

@mew1033 and @fossam is there a good way to get the data structure to nest in the splunk event? Right now the grains are just hitting splunk as a giant string:

https://www.dropbox.com/s/1fcpct1la3xaye6/Screenshot%202017-12-14%2011.43.17.png?dl=0

I would love to get them as a nice nested thing so you can query the fields. I tried just sticking the dictionary in the message field as an actual dictionary, but the event never made it to splunk. Do I need to shove them in as top-level fields?

Also note that in future events the loggername will be hubblestack.grains_report, not hubblestack.daemon like it's showing in that screenshot.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -302,7 +302,7 @@ def load_config():
     salt.config.DEFAULT_MINION_OPTS['log_level'] = None
     salt.config.DEFAULT_MINION_OPTS['file_client'] = 'local'
     salt.config.DEFAULT_MINION_OPTS['fileserver_update_frequency'] = 43200  # 12 hours
-    salt.config.DEFAULT_MINION_OPTS['grains_refresh_frequency'] = 3600  # 12 hours
+    salt.config.DEFAULT_MINION_OPTS['grains_refresh_frequency'] = 3600  # 1 hour
     salt.config.DEFAULT_MINION_OPTS['scheduler_sleep_frequency'] = 0.5
     salt.config.DEFAULT_MINION_OPTS['default_include'] = 'hubble.d/*.conf'
 
@@ -428,7 +428,7 @@ def refresh_grains(initial=False):
                 self.asctime = asctime
                 self.name = name
         handler = hubblestack.splunklogging.SplunkHandler()
-        handler.emit(MockRecord(str(__grains__), 'INFO', time.asctime(), 'hubblestack.grains_report'))
+        handler.emit(MockRecord(__grains__, 'INFO', time.asctime(), 'hubblestack.grains_report'))
 
 
 def parse_args():
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -13,6 +13,7 @@ import os
 import random
 import signal
 import sys
+import uuid
 
 import salt.fileclient
 import salt.utils
@@ -24,6 +25,8 @@ from hubblestack import __version__
 log = logging.getLogger(__name__)
 
 __opts__ = {}
+# This should work fine until we go to multiprocessing
+SESSION_UUID = uuid.uuid4()
 
 
 def run():
@@ -80,6 +83,8 @@ def main():
         run_function()
         sys.exit(0)
 
+    last_grains_refresh = time.time()
+
     log.info('Starting main loop')
     while True:
         # Check if fileserver needs update
@@ -89,6 +94,9 @@ def main():
                 last_fc_update = time.time()
             except Exception as exc:
                 log.exception('Exception thrown trying to update fileclient.')
+
+        if time.time() - last_grains_refresh >= __opts__['grains_refresh_frequency']:
+            refresh_grains()
 
         try:
             log.debug('Executing schedule')
@@ -292,15 +300,11 @@ def load_config():
     salt.config.DEFAULT_MINION_OPTS['log_level'] = None
     salt.config.DEFAULT_MINION_OPTS['file_client'] = 'local'
     salt.config.DEFAULT_MINION_OPTS['fileserver_update_frequency'] = 43200  # 12 hours
+    salt.config.DEFAULT_MINION_OPTS['grains_refresh_frequency'] = 3600  # 12 hours
     salt.config.DEFAULT_MINION_OPTS['scheduler_sleep_frequency'] = 0.5
     salt.config.DEFAULT_MINION_OPTS['default_include'] = 'hubble.d/*.conf'
 
     global __opts__
-    global __grains__
-    global __utils__
-    global __salt__
-    global __pillar__
-    global __returners__
 
     __opts__ = salt.config.minion_config(parsed_args.get('configfile'))
     __opts__.update(parsed_args)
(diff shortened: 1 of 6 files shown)


hubblestack/daemon.py
```diff
@@ -381,21 +387,48 @@ def load_config():
...
+        class MockRecord(object):
+            def __init__(self, message, levelname, asctime, name):
+                self.message = message
+                self.levelname = levelname
+                self.asctime = asctime
+                self.name = name
         handler = hubblestack.splunklogging.SplunkHandler()
-        handler.setLevel(logging.ERROR)
-        root_logger.addHandler(handler)
+        handler.emit(MockRecord(str(__grains__), 'INFO', time.asctime(), 'hubblestack.grains_report'))
```
mew1033: Just get rid of the `str()` around `__grains__` and the message will get sent in all extracted and beautiful.

mew1033: Just get rid of the `str()` around `__grains__` and the message will get sent in all extracted and beautiful.

hubblestack/daemon.py
```diff
@@ -292,15 +302,11 @@ def load_config():
     salt.config.DEFAULT_MINION_OPTS['log_level'] = None
     salt.config.DEFAULT_MINION_OPTS['file_client'] = 'local'
     salt.config.DEFAULT_MINION_OPTS['fileserver_update_frequency'] = 43200  # 12 hours
+    salt.config.DEFAULT_MINION_OPTS['grains_refresh_frequency'] = 3600  # 12 hours
```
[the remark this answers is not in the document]
fossam: Unless it's too heavy operation, once an hour could be fine also.

hubblestack/extmods/grains/hostuuid.py
```diff
@@ -0,0 +1,28 @@
...
+def host_uuid():
+    '''
+    Generate a unique uuid for this host, storing it on disk so it persists
+    across restarts
+    '''
+    cached_uuid = os.path.join(os.path.dirname(__opts__['configfile']), 'hubble_cached_uuid')
+    try:
+        if os.path.isfile(cached_uuid):
+            with open(cached_uuid, 'r') as f:
+                return {'host_uuid': f.read()}
```
fossam: Can the existing host_uuid be compared to the one read from the file, compared against each other, and emit a log message with relevant details if they don't match?

Also, would it be possible to log differently a failure to read a cached host_uuid if there is supposed to be one? This would show that there _was_ a host_uuid but it got removed. Given that host_uuid exists in program memory, but disappeared on disk, we should probably attempt to save it by writing the same one back.

Finally, a failed attempt to write the host_uuid should also be logged.

I think changing host_uuid's should not be easy or expected. All possible avenues of changes to this field should be watched for and logged to establish continuity between the two values.

hubblestack/daemon.py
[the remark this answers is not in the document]
basepi: I wanted once per hour, just didn't change the comment.

hubblestack/daemon.py
basepi: Hmmm. I thought I tried that and it didn't work, but maybe something else was going on. Glad it's working for you! I'll do some more testing when we get closer to release as well.

hubblestack/extmods/grains/hostuuid.py
basepi: Done. I think I hit on all your points.

basepi:
@mew1033 I addressed both of your changes, and also added more checking around host_uuid at the request of @fossam.

hubblestack/extmods/grains/hostuuid.py
```diff
@@ -15,14 +15,36 @@ def host_uuid():
...
+        elif existing_uuid:
+            log.error('host_uuid was previously generated, but the cached '
+                      'file is no longer present: {0}'.format(cached_uuid_path))
+    except Exception:
+        log.exception('Problem retrieving cached host uuid from file: {0}'
+                      .format(cached_uuid_path))
+
+    # Generate a fresh one if needed
+    if not existing_uuid:
+        existing_uuid = str(uuid.uuid4())
```
fossam: Can we also add a log message on generating a fresh one? 
Chances are that this should fire rarely. When it does fire, either there was a clean wipe (stop & cache delete - something we'd like to be aware of) or it's a completely new host - also a cause for celebration.

basepi: Done. I added it further up, with the rest of the messages around the cache file: https://github.com/hubblestack/hubble/pull/260/commits/d9fe303f5ef4d754e08dd34b5718f82f5866ee8b

mew1033:
Do you think it'd be a good idea to have the returners call the grain refresh function every time before they run? That way every return always has the most up to date grain data./

basepi:
I do not. Grains refresh is a fairly heavy operation and I don’t want to run it more often than hourly, which is already probably overkill. 

Grains do not and should not change very often. But if they do we want to know about it.

hubblestack/extmods/grains/hostuuid.py
```diff
@@ -0,0 +1,56 @@
...
+    try:
+        if os.path.isfile(cached_uuid_path):
+            with open(cached_uuid_path, 'r') as f:
+                cached_uuid = f.read()
+                # Check if it's changed out from under us -- problem!
+                if existing_uuid and cached_uuid != existing_uuid:
+                    log.error('host_uuid changed on disk unexpectedly!'
+                              '\nPrevious: {0}\nNew: {1}\nKeeping previous.'
+                              .format(existing_uuid, cached_uuid))
+                    return {'host_uuid': existing_uuid}
```
fossam: Trying to think about all possible outcomes, so set me straight: here we exit with returning the existing uuid. But when do we overwrite the cache file? If we don't overwrite, this fill keep logging every grains refresh.

basepi: Good catch. I fixed it.

hubblestack/extmods/grains/hostuuid.py
```diff
@@ -0,0 +1,63 @@
...
+                return {'host_uuid': cached_uuid}
+        elif existing_uuid:
+            log.error('host_uuid was previously generated, but the cached '
+                      'file is no longer present: {0}'.format(cached_uuid_path))
+        else:
+            # TODO: once we figure out how to get a custom log level for
+            # non-error splunk logs, we should move this to that log level so
+            # it doesn't look like an error.
+            log.error('generating fresh uuid, no cache file found. '
```
fossam: Is it possible to structure this so that when a new uuid is generated, it is included with the log message? 
The reason being that depending on how this is logged, on the siem side stitching together messages and figuring out a new uuid becomes harder. This is also one case when this field is not (yet) present when sent to a returner, so details would help.

PR #263, opened 2018-01-10
Fix run on start issue

anuragpaliwal80:
Audit module is not getting started even after setting run_on_start as True. It might be because of this splay check condition.
If we set splay as 0 in config, it works as expected. 

Can someone please explain splay usages/impact ? I know it's range of time plus minus to the seconds variable.


basepi:
A missing splay should default to zero on line 188 above where your change was made. If it's not we need to investigate.

Splays are designed to allow you to randomize the start time for jobs, in case you restart your hubble daemon everywhere all at once, and don't want them to overwhelm your splunk server or whatever. It is only applied once, on first run of the job.

anuragpaliwal80:
Thanks for the update.
Right now if splay is set to any value other than 0 suppose 30 and run_on_start is True (for example for audit module). I was expecting that after restarting the hubble service, audit module will be executed with in 30 sec irrespective of value mentioned in seconds. But that's not the case, the audit is not getting executed in 30 sec.

I will make some changes as per my correct understanding of one time usages of splay parameter.

basepi:
So the correct behavior is for the audit to run anywhere from 0 to 30 seconds after startup. Since you have a workaround for the time being, I'm going to let this sit until my brain is in a better place to review it. I'll make sure I review it when I'm back to work in a week.

basepi:
I spotted the bug! I don't think your solution was quite right. It had some duplication around detecting whether the splay had already been calculated, and a couple of other problems. As I was writing the review I spotted the original bug and realized it would be faster if I just fixed it. Luckily, with Github's awesome new PR review process I could just add a commit directly to your branch! I hope that's OK.

@anuragpaliwal80 can you please review this and test it to make sure it solves your original problem? I think it will.

anuragpaliwal80:
Working as expected, thanks for making the changes.

PR #268, opened 2018-01-22
sqlite returner for testing

jettero:
I've got a whole test setup planned out that will use this to verify (eg) FIM is working as intended. Whether the test setup gets used or not, this might still be useful to someone not using splunk.


hubblestack/extmods/returners/sqlite.py
```diff
@@ -0,0 +1,131 @@
+# -*- coding: utf-8 -*-
+'''
+Return hubble data to ssqlite (intended for testing)
```
mew1033: Got an extra s there

hubblestack/extmods/returners/sqlite.py
```diff
@@ -0,0 +1,131 @@
...
+            os.makedirs(_d, 0755)
+        log.debug('connecting to database in {0}'.format(_p))
+        _conn = sqlite3.connect( _options.get('dumpster', 'hubble-returns.db') )
+        log.debug('creating ret table')
+        _conn.execute('''create table if not exists ret(
+            jid text, id text, fun text, fun_args json,
+            ret json)''')
+    return _conn
+
+## {u'fun': u'hubble.audit',
```
mew1033: Was this just a test event you were referencing while you were building out the returner?

cedwards:
I like this idea! Thanks, @jettero.
Let's clean it up a bit (see @mew1033 's comments). Otherwise, I imagine a lot of potential uses for this.

hubblestack/extmods/returners/sqlite.py
jettero: Yeah, it's for reference. Should I remove that?

hubblestack/extmods/returners/sqlite.py
jettero: fixed.

jettero:
Where are we on this? If it needs additional changes, I'm not sure what they would be.

basepi:
Looks like we got some travis and other testing files in here. Are those ready to be merged or should we move them to another branch?

Otherwise I was just waiting on @mew1033 to approve your changes, since he had the "requested changes" review.

jettero:
oh, I didn't realize. I'll clean that up for sure.

hubblestack/extmods/returners/sqlite.py
```diff
@@ -0,0 +1,132 @@
...
+    except:
+        pass
+    if GOT_JSON and GOT_SQLI:
+        return __virtualname__
+    msgs = []
+    if not GOT_JSON:
+        msgs.append("json module is missing")
+    if not GOT_SQLI:
+        msgs.append("sqlite3 module is missing")
+    return False, ', '.join(msgs)
```
basepi: The "salty" pattern for gated imports is to do these try:except blocks outside the `__virtual__`. Can you move these imports accordingly?

jettero: I could change that and have no problem doing it. But I don't understand why. Moving them to global space means polluting the global() dict with a bunch of ALL_CAPS variables so you can generate a meaningful error. Even if you import more than once by mistake, the import just fails to actually import, so you don't lose anything.

Is the motivation pure distaste for the 'global' keyword or something else?

jettero: (Though, the more I think about it... because of the way I'm doing it, it might just fool python into re-importing every __virtual__() call ... not really sure.)

basepi: If we're going to bring code lint into it I still think it belongs outside, so it's grouped with the rest of the imports.

basepi: That also has the nice side effect of removing the need for the `global` keyword, which is nice.

jettero:
OK... I took the "gate" off json (it's got to be there, c'mon) and moved the sqlite3 gate to the top. Then rebased.

PR #273, opened 2018-01-23
pidfile issues (check_pidfile() to see if Hubble is already running)

jettero:
If there's a pidfile, check to see if it's hubble. If it is, kill the other one or die.

My thinking is that if we `hubble -d` we wish to start or restart. Writing and checking pidfiles does us no good at all if we're in systemd's simple process tracking mode though — in which mode, hubble would be started without the '-d' and would never attempt to daemonize.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -50,9 +50,11 @@ def run():
         sys.exit(0)
 
     if __opts__['daemonize']:
-        check_pidfile()
+        check_pidfile(kill_other=True)
         salt.utils.daemonize()
         create_pidfile()
+    else:
+        check_pidfile(kill_other=False)
 
     signal.signal(signal.SIGTERM, clean_up_process)
     signal.signal(signal.SIGINT, clean_up_process)
@@ -467,10 +469,15 @@ def parse_args():
                         help='Any arguments necessary for a single function run')
     return vars(parser.parse_args())
 
-def check_pidfile():
+def check_pidfile(kill_other=False):
     '''
     Check to see if there's already a pidfile. If so, check to see if the
     indicated process is alive and is Hubble.
+
+    kill_other
+        Default false, if set to true, attempt to kill detected running Hubble
+        processes; otherwise exit with an error.
+
     '''
     pidfile = __opts__['pidfile']
     if os.path.isfile(pidfile):
@@ -487,14 +494,18 @@ def check_pidfile():
                     with open("/proc/{pid}/cmdline".format(pid=xpid),'r') as f2:
                         cmdline = f2.readline().strip().strip('\x00').replace('\x00',' ')
                         if 'hubble' in cmdline:
-                            log.warn("process seems to still be alive and is hubble, attempting to shutdown")
-                            os.kill(int(xpid), signal.SIGTERM)
-                            time.sleep(1)
-                            if os.path.isdir("/proc/{pid}".format(pid=xpid)):
-                                log.error("failed to shutdown process successfully; abnormal program exit")
-                                exit(1)
+                            if kill_other:
+                                log.warn("process seems to still be alive and is hubble, attempting to shutdown")
+                                os.kill(int(xpid), signal.SIGTERM)
+                                time.sleep(1)
+                                if os.path.isdir("/proc/{pid}".format(pid=xpid)):
+                                    log.error("failed to shutdown process successfully; abnormal program exit")
+                                    exit(1)
+                                else:
+                                    log.info("shutdown seems to have succeeded, proceeding with startup")
                             else:
-                                log.info("shutdown seems to have succeeded, proceeding with startup")
+                                log.error("refusing to run while another hubble instance is running")
+                                exit(1)
                         else:
                             log.info("process does not appear to be hubble, ignoring")
 
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -50,10 +50,14 @@ def run():
         sys.exit(0)
 
     if __opts__['daemonize']:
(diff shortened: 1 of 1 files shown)


hubblestack/daemon.py
```diff
@@ -466,6 +469,45 @@ def parse_args():
...
+    Check to see if there's already a pidfile. If so, check to see if the
+    indicated process is alive and is Hubble.
+
+    kill_other
+        Default false, if set to true, attempt to kill detected running Hubble
+        processes; otherwise exit with an error.
+
+    '''
+    pidfile = __opts__['pidfile']
+    if os.path.isfile(pidfile):
```
fossam: I think a concern was surfaced elsewhere along the lines that, what if a pid file is missing, would we still get a duplicate process?

hubblestack/daemon.py
```diff
@@ -466,6 +469,45 @@ def parse_args():
...
+                        if 'hubble' in cmdline:
+                            if kill_other:
+                                log.warn("process seems to still be alive and is hubble, attempting to shutdown")
+                                os.kill(int(xpid), signal.SIGTERM)
+                                time.sleep(1)
+                                if os.path.isdir("/proc/{pid}".format(pid=xpid)):
+                                    log.error("failed to shutdown process successfully; abnormal program exit")
+                                    exit(1)
+                                else:
+                                    log.info("shutdown seems to have succeeded, proceeding with startup")
```
fossam: FYI, error will be sent to a returner with a sourcetype of _hubble_log_. I think it would be useful to detect and see in that data indications that this action (killing a concurrent hubble process) is happening.

It also would probably be useful to include some details with the action, like the _pid_.

hubblestack/daemon.py
jettero: Probably. I'm not aware of many daemons that check to see if another instances is running without using a pidfile like this. Databases tend to do it, because they notice eachother accessing the same database files. That's just about it though.

I could certainly add further checks, but crawling through the process list for things that happen to be named hubble might result as many false positives as not. Another way to approach it might be to keep a lock on some file in /var/lib/hubble or something. But older versions of hubble wouldn't have that lock, so it wouldn't help there either.

hubblestack/daemon.py
jettero: I'm not convinced the killed process would invoke any sort of return. A TERM shutdown seems to be caught as a single to do a normal shutdown (as if it had come from sysv or systemd). And in the case where it can't send a TERM or it doesn't work and the new process shuts down in stead, it will not have gotten yet to the code that invokes returns. The exit() is pre-fork, so it would print logs to stdout/stderr and exit.

fossam: Sorry, I meant the process that does the killing, not the other way around. I was commenting on the _log.error_ and _log.info_ usage. 
I know it's good to differentiate error level from info level, but sometimes we also want to have things be observable from the data collection side.

jettero: OK, I'm confused then. I should set the log.info() to log.error()? or the other way around?

fossam: That's something for you and/or @basepi to decide.

I personally think it's tricky to detect problems from the data if all the relevant details are not there and if all the logs are just the _error_ kind. Think it from the point of view where you have no access to a log file on disk - the only access you have is to the data already sent to a returner.

basepi: One thing I've been considering is trying to add another custom log level, like `log.splunk` or something, so that we can log things to splunk log handler without having to call them "errors". I'm not sure if it's possible, I'm still a bit of a noob when it comes to the inner workings of the `logging` module.

hubblestack/daemon.py
fossam: /proc/<pid\>/exe seems to point to _/opt/hubble/hubble-libs/hubble_ even in older versions.
This needs a case when a process is there but the file has been deleted during the upgrade. It should still have a pointer, even when it says deleted.

Given the possible slow adoption schedule for new versions I think it's a good idea to fix this problem so that is unlikely to occur again.

hubblestack/daemon.py
```diff
@@ -50,8 +50,11 @@ def run():
         sys.exit(0)
 
     if __opts__['daemonize']:
+        check_pidfile(kill_other=True)
         salt.utils.daemonize()
         create_pidfile()
+    else:
```
basepi: Can you please add a check to see if it's a single function run or not?

https://github.com/hubblestack/hubble/blob/develop/hubblestack/daemon.py#L81-L84

If you're just running a single function, rather than running the scheduler forever, I want to execute the single function even if hubble is already running on the system.

jettero: I put the pidfile check next to the pidfile drop... it's only applicable when invoked in daemon mode. I can double check to see if it works that way, but I'm pretty sure.

hubblestack/daemon.py
jettero: I'm not sure I understand the problem here. I'm going to review the comments, but this all happens before daemonizing, so all the info and error results are printed to console (or systemd logs or whatever) and wouldn't end up in a return.

Except where I send a kill single 15, which would get caught by the usual shutdown signal handler.

hubblestack/daemon.py
jettero: oh, nm, I see. I set it to refuse to run if there's a daemon running without checking function.

jettero: I'm not convinced it really needed the single function check, but I added it just the same. Confirmed that if hubble is already running and I run the thing with `-d`, it replaces the running process with the new one. Also, if I run it `hubble grains.get fqdn` it just returns the grain and leaves the running process alone.

PR #275, opened 2018-01-29
win_secedit fix for correct match failures

madchills:
Added an if statement that passes through the match variable if its already in SID format on the yaml

### DIFF
--- a/hubblestack/files/hubblestack_nova/win_secedit.py
+++ b/hubblestack/files/hubblestack_nova/win_secedit.py
@@ -319,7 +319,8 @@ def _translate_value_type(current, value, evaluator, __sidaccounts__=False):
             else:
                 return True
     elif 'account' in value:
-        evaluator = _account_audit(evaluator, __sidaccounts__)
+        if "*S-" not in evaluator:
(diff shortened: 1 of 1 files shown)


hubblestack/files/hubblestack_nova/win_secedit.py
```diff
@@ -384,6 +384,12 @@ def _account_audit(current, __sidaccounts__):
     ret_string = ''
     if __sidaccounts__:
         for usr in user_list:
+            if usr.startswith("*S-"):
+                if not ret_string:
+                    ret_string = current
+                else:
+                    ret_string = current
```
basepi: This if-else block does the same thing in both branches. If that's intentional, then it can be removed. If it's unintentional it's a bug.

madchills: I was trying to be smart with my yaml.  In the end you are right, it didn't makes sense and I just fixed my yaml.

madchills:
I just removed it completely from the _account_audit function.  This function tries to convert non SID account names to SID. If the yaml already has SIDs in place of names, we don't need to call this function at all.  I did the check before going into the function instead.

PR #276, opened 2018-01-31
Systemd service issue

ryandamour:
Not sure if this is the best approach, but as it stands the binary the current hubble.service file is looking for (via /opt/hubble/hubble) doesn't exist.  It was suggested to use a symlink in a prior PR, but the binary itself is non-existent in this case. ,

### DIFF
--- a/pkg/centos7/hubble.service
+++ b/pkg/centos7/hubble.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=Hubblestack
+
+[Service]
+Type=forking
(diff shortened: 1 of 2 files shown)


basepi:
Ah, right, the existing service file wouldn't work for source installs.

This is probably a fine solution for the moment. I can't remember if all the packages stick the binary in /usr/bin/hubble or if maybe one or two hits /usr/local/bin/hubble or something. We should evaluate those packages and get everything in the same file (or two).

PR #277, opened 2018-02-01
splunk DRY things

jettero:
I was talking with George Starcher about a minor issue in hubblestack/splunklogging's hec object. I fixed that minor bug, but later learned sendEvent() isn't actually used (just batch events).

While trying to fix the non-issue I noticed there were four versions of his http_event_collector object and four ways to fetch the options for it. I've attempted to DRY those objects and their _get_options for easier ongoing maintenance. (Suppose the next bug is real, but now we have 4 variants of the HEC to modify.)

Thoughts welcome.


jettero:
probably some other things too... I'll rebase it and see how it looks.

jettero:
(this does not rebase cleanly...)

jettero:
rebasing this caused like a billion merge conflicts... please be sure to click the #7489667 commit. I'm not really sure removing that stuff was right. Am I missing something?

basepi:
We need all the splunk files to retrieve cloud_details from `__grains__` -- looks like we lost some of https://github.com/hubblestack/hubble/pull/300/files in your 7489667a270adeb3f5b555688f8da62c2c587e89

jettero:
grr, I thought I found all those.

I had a halfwit thought of trying to cache the grain (cuz it's used all over); but then relented...

During the merge(conflicts) in the rebase, I had to go unthink that thought all over the place.

basepi:
Sometimes it can be worth starting from scratch. :) But I'm OK with whatever you decide!

basepi:
@jettero before you do too much extensive testing, this is conflicting, so it needs a merge or a rebase.

jettero:
Yeah, I'm not nearly done. This is mostly the old PR I think. I'll push first thing tomorrow.

There's still easy to reproduce bugs I have to work out.

jettero:
Er, I should also say, I resloved the conflicts first thing.

jettero:
Oh, you know what? I never did fix the merge conflicts. I really thought I did. This is going to be rough.

hubblestack/hec/dq.py
```diff
@@ -0,0 +1,127 @@
...
+            # corrupted or the wrong sqlite version or whatever; just give up
+            # and consider those msgs lost.
+            log.error("problem opening the diskqueue at %s: %s", self.directory, e)
+            log.error("deleting the diskqueue")
+            cache_file = os.path.join(self.directory, 'cache.db')
+            if os.path.isfile(cache_file):
+                for f in glob.glob(cache_file + '*'):
+                    os.unlink(f)
+            # If it still doesn't work, crash in the previous way
+            return Deque(directory=self.directory)
```
basepi: I think you can just `raise` and it will raise the exception from this context. Should be more performant (and more precise)

hubblestack/hec/dq.py
```diff
@@ -0,0 +1,127 @@
+
```
basepi: Instead of leading whitespace, please add:

```
# -*- encoding: utf-8 -*-
```

hubblestack/hec/dq.py
```diff
@@ -0,0 +1,127 @@
...
+import time
+from diskcache import Deque
+from sqlite3 import DatabaseError as SQLiteDBError
+import shutil, os
+import glob
+
+# fakelogging goes to a file iff specified in environment
+# and otherwise does nothing (for debugging)
+import fakelogging as logging
+log = logging.getLogger('dq')
```
basepi: I think using `__name__` is the standard way to do this -- why hardcode it if you don't have to?

hubblestack/hec/fakelogging.py
```diff
@@ -0,0 +1,28 @@
...
+    '''
+    def __init__(self, name):
+        self.file = os.environ.get('HUBBLE_HEC_DEBUG')
+        self.name = 'hubblestack.hec.' + name
+    def _show(self, level, fmt, *a):
+        if not self.file:
+            return
+        p = os.getpid()
+        t = time.ctime()
+        with open(self.file, 'a') as fh:
```
basepi: I don't love avoiding the logging system altogether -- this means that you won't see these emitted on STDERR if running in the foreground, that kind of stuff. We may have to revisit this.

hubblestack/hec/obj.py
```diff
@@ -0,0 +1,327 @@
...
+
+__version__ = '1.0'
+
+_max_content_bytes = 100000
+http_event_collector_debug = False
+
+# the list of collector URLs given to the HEC object
+# are hashed into an md5 string that identifies the URL set
+# these maximums are per URL set, not for the entire disk cache
+max_diskqueue_items = 200
```
basepi: This is likely too small. We probably just want to control with disk size, and put this to "unlimited" or an arbitrarily high number.

hubblestack/hec/obj.py
```diff
@@ -0,0 +1,327 @@
...
+import urllib3
+
+from . dq import DiskQueue
+
+# fakelogging goes to a file iff specified in environment
+# and otherwise does nothing (for debugging)
+import fakelogging as logging
+log = logging.getLogger('obj')
+
+__version__ = '1.0'
```
basepi: Does this need versioning independent of hubble?

hubblestack/hec/opt.py
```diff
@@ -0,0 +1,94 @@
...
+def _get_splunk_options(space, modality, **kw):
+    ret = list()
+
+    base_opts = {
+        'token': REQUIRED,
+        'indexer': REQUIRED,
+        'index': REQUIRED,
+        'port': '8088',
+        'custom_fields': [],
+        'sourcetype': 'hubble_log',
```
basepi: Unless I've missed something, this module doesn't handle sourcetype correctly. In each of the `_get_options()` that were collapsed into this one, they looked at specific config values for their sourcetype. For example, in the nova returner:

```
            processed['sourcetype'] = opt.get('sourcetype_nova', 'hubble_audit')
```

I don't think this will be handled correctly in the new code.

hubblestack/splunklogging.py
```diff
@@ -41,25 +41,18 @@
...
-
 class SplunkHandler(logging.Handler):
     '''
     Log handler for splunk
     '''
     def __init__(self):
         super(SplunkHandler, self).__init__()
 
-        self.opts_list = _get_options()
+        self.opts_list = get_splunk_options('hubblestack:returner:splunk')
```
basepi: The argument here is unnecessary, right? Because it defaults to this value?

hubblestack/hec/__init__.py
```diff
@@ -0,0 +1,3 @@
+
```
basepi: Instead of leading whitespace, please add:

```
# -*- encoding: utf-8 -*-
```

hubblestack/hec/fakelogging.py
```diff
@@ -0,0 +1,28 @@
+
```
basepi: Instead of leading whitespace, please add:

```
# -*- encoding: utf-8 -*-
```

hubblestack/hec/obj.py
```diff
@@ -0,0 +1,327 @@
+
```
basepi: Instead of leading whitespace, please add:

```
# -*- encoding: utf-8 -*-
```

hubblestack/hec/opt.py
```diff
@@ -0,0 +1,94 @@
+
```
basepi: Instead of leading whitespace, please add:

```
# -*- encoding: utf-8 -*-
```

tests/unittests/test_hec_dq.py
```diff
@@ -0,0 +1,63 @@
+
```
basepi: Instead of leading whitespace, please add:

```
# -*- encoding: utf-8 -*-
```

hubblestack/hec/obj.py
jettero: I feel like the items limiter is usually going to be better. The disk size is somewhat random and I feel like (without being able to justify it fully) it's going to be more reliable. 200 may be a bit low. I really have no idea what reasonable settings for either of these are though, so we'll 0 this one and set the size to 50megs or something? 100megs? It's per hec endpoint, so probably a bit lower is better.

basepi: Yeah I'm not sure either. I expect @fossam will want it "as high as possible" but have no concept of how many events we'll need to buffer to actually cover small outages.

basepi:
Cool. I resolved the threads that were resolved. We'll talk more after splunk conf

hubblestack/extmods/returners/splunk_pulsar_return.py
```diff
@@ -315,203 +312,3 @@ def _dedupList(l):
...
-        if not isinstance(returner_opts, list):
-            returner_opts = [returner_opts]
-        for opt in returner_opts:
-            processed = {}
-            processed['token'] = opt.get('token')
-            processed['indexer'] = opt.get('indexer')
-            processed['port'] = str(opt.get('port', '8088'))
-            processed['index'] = opt.get('index')
-            processed['custom_fields'] = opt.get('custom_fields', [])
-            processed['sourcetype'] = opt.get('sourcetype_pulsar', 'hubble_fim')
```
basepi: It's easy to miss, but this line is different in each of the returners. The result is that all data is being returned under `hubble_log`, rather than `hubble_osquery` or `hubble_audit`, or what have you.

Can you find a way to parameterize this piece of config? I figure you add an optional parameter to `get_splunk_options()` so that a returner can define how you would look up its sourcetype.

hubblestack/hec/opt.py
basepi: Turns out I did notice this earlier. Might be worth going through and reviewing any "unresolved" comments of mine.

hubblestack/extmods/returners/splunk_pulsar_return.py
jettero: ugh. this is harder than I thought.

jettero: wait, maybe not... I put all my new updates in the stash ... testing. I think I handled this but don't recall how it works...

jettero: yeah. this is easy again. nevermind the several hours I wasted making this difficult.

hubblestack/hec/dq.py
jettero: It's a comment bug, not a performance issue or a raise style problem...

We unlinked the old sqlite3 file assuming some sort of journal or version mismatch or something, then we re-try the write but without catching it.

If there's any bug here (aside from the comment, which I'll re-phrase), it's probably that this could take the whole hubble-daemon down with it.

hubblestack/hec/obj.py
jettero: no, probably not. I think originally I was thinking of disk queue versioning that might need something like that.

jettero: right, it has to do with the hubble-hec/1.0 user agent string more than anything. I could just move that into the user_agent population area and leave it out of globals, but it made sense to me at the time I wrote it; and seems to still knowing about user-agent strings.

hubblestack/splunklogging.py
jettero: yeah, that's the default. I think originally I planned to be really specific, but later learned the sourcetype options are name-prefixed in that same options namespace.

hubblestack/hec/opt.py
jettero: I think I mentioned elsewhere... I did think to handle this, but forgot to actually put the code in the modules. I came up with a poorly named (I couldn't think of a good name) strategy for remaping name-prefix options to their proper name...

... I've added a commit to populate those correctly.

basepi: Still doesn't work:

```
[DEBUG   ] Options: {"index": "default_hubble", "sourcetype": "hubble_log", "http_event_server_ssl": true, "index_extracted_fields": [], "indexer": "<redacted>", "token": "<redacted>", "proxy": {}, "timeout": 9.05, "port": "8088", "custom_fields": [], "http_event_collector_ssl_verify": true}
```

Same result for nebula except there's randomly a new field:

```
[DEBUG   ] Options: {"index": "default_hubble", "timeout": 9.05, "sourcetype": "hubble_log", "http_event_server_ssl": true, "index_extracted_fields": [], "indexer": "<redacted>", "token": "<redacted>", "add_query_to_sourcetype": true, "proxy": {}, "sourcetype_nebula": "hubble_osquery", "port": "8088", "custom_fields": [], "http_event_collector_ssl_verify": true}
```

basepi: The new field is probably from the fact that the three modules have slightly different kwargs [in your commit](https://github.com/hubblestack/hubble/pull/277/commits/23a67e820cc5f6d8eeee0c74b7e921b71f12db26)

jettero: well, that's what the _nick mapping is for... it remaps those slightly different names to the right name. It's totally working for me. still investigating though.

hubblestack/extmods/returners/splunk_pulsar_return.py
basepi: Tested and working.

hubblestack/hec/dq.py
basepi: ohhhh I totally misread this. Sorry.

hubblestack/hec/opt.py
basepi: Worked with @jettero over slack and this is now working.

PR #284, opened 2018-02-07
Fix retry logic for fileserver issues

basepi:
On startup, by default will retry every 30 seconds forever, unless it's
a single-function run.

After startup, by default will retry every 900 seconds (fifteen minutes)
while continuing to execute the schedule as normal.


jettero:
I feel a little silly "approving" your changes... but it looks solid.

PR #285, opened 2018-02-08
Add grains to detect default gateway

basepi:


ryandamour:
For what it's worth, you can obtain the default gateway if the "route" command is installed, even with "ip" not being available:

`/usr/sbin/route -n | head -n 3 | tail -n 1 | awk '{print $2}'`

basepi:
Let's see, I swear there was a reason I used `ip` instead of `route`. I'll have to think about that more.

I'd have to do more work than just your command, because honestly the thing we care about most is whether there *is* a default gateway, not what it is, at least for our use case. We want to differentiate between data center hosts with no outbound and hosts that can hit the internet.

Additionally, `route` output is much less standardized because of all the potential for multiple routing tables and the like. The first line will not always be the default gateway.

We could still probably get most of the way there, and it's not a bad idea to have a fallback even if it's a naïve one.

fossam:
As far as Linux goes the `ip` command is pretty standard now. And if all else fails, there is /proc/net/route.

Another way get to the goal is to _query_ the host routing table for the address you're interested in. For example, if you want to make sure you have a route for `github.com` resolve the domain (192.30.255.112 in this case) and issue command `ip route get 192.30.255.112`. A host with a route for that domain will respond with output similar to:
```
192.30.255.112 via 192.168.122.1 dev ens3  src 192.168.122.165 
    cache
``` 
and exit code `0`
while the host with no route will respond with output 
```
RTNETLINK answers: Network is unreachable
```
and exit code `2`.
This can cover even cases where the host does not have a default route, but still has a specific route to a needed domain (hosts running bgp routing with full internet routing tables).

basepi:
That's not a bad idea. Doesn't fit very cleanly into the "grain" paradigm, but would be useful if we found that some hosts could reach github or azure but didn't have a default gateway.

fossam:
Successful run of the command will tell you the gateway that the host will use for that domain/ip (3rd value). But it's not necessarily the "default".

jettero:
Yeah, route will be everywhere even if iproute2 is missing for some reason. I'd probably grep for actual default routes (`route -n | grep ^0.0.0.0 | head -n 1`) rather than just the 1st or 3rd line.

But note that even without a default route, you might have a route to the splunk (or whatever) endpoints that doesn't go through the default gateway. You could have a static route to some private network. It's totally conceivable.

Along those lines, I'd try to ask `ip route get $splunk_IP_here` and see if I get a route out of it.

basepi:
Good points, but again, those types of changes belong in https://github.com/hubblestack/hubble/pull/286 rather than in general-use grains.

I think I'm going to stick with `ip` here for now, to match the unreleased grains in upstream salt.

@fossam @jettero could one of you approve this if you agree so we can merge it?

jettero:
right, I didn't understand the context apparently.

PR #286, opened 2018-02-08
Add fallback to fileserver in case of no default gateway

basepi:


fossam:
If there's one thing, can the `python_shell=True` be avoided in #285 ?

basepi:
@fossam Maybe. But it's not relevant to this PR, and not necessary in any case, as it doesn't use user input, so is not a command injection vulnerability.

fossam:
You're correct, I just wanted to avoid shelling out since it was easily avoidable in this case. Sorry to derail conversation in this PR.

To help us get back on the subject - what happens in case there is no default gateway but a configured fileserver is reachable regardless?

basepi:
It will fall back. Which I think is fine for this stop-gap solution -- see the conversation over [here](https://github.com/hubblestack/hubble/pull/285) where other solutions were discussed. We should probably implement a more targeted solution in the future, but with the looming release I think we should stick with this for now.

sumanmehta:
Does the token remain same for fallback_indexer? Do you not need a fallback_token as well?

basepi:
For DMa, the auth information is the same. It's just whether we use internal DNS or external DNS. One resolves, the other doesn't, for these hosts.

We could add alternate auth as an option if your team would like to use it.

sumanmehta:
ok. PR looks fine to me.

PR #288, opened 2018-02-12
Readme changes

dave-pollock:
* Fixes to contents
* Add getting started section
* Clarify file paths
* Remove references to minons since this is a salt term
* Add example for nova controls
* Add comment in hubble config file clarifying that both gitfs and file_roots can be used in combination

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ Table of Contents
    * [HUBBLE](#hubble)
       * [Packaging / Installing](#packaging--installing)
          * [Installing using setup.py](#installing-using-setuppy)
-         * [Building standalone packages (CentOS)](#building-standalone-packages-centos)
-         * [Building standalone packages (Debian)](#building-standalone-packages-debian)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -49,20 +49,43 @@ To run the container (which will output the package file in your current directo
...
+Now, edit the Hubble configuration file `/etc/hubble/hubble`. Uncomment the following lines:
+```
+# file_roots:
+#   base:
+#     - /srv/salt/hubblestack_data
+```
+With the file_root configuration added:
+1. Run `hubble hubble.sync` to pull the files into Hubble's cache
+1. Run `hubble hubble.load`
+1. Run `hubble hubble.audit`. The agent should now only run the network.ssh profile.
```
basepi: So these three steps are not actually required. By default, `hubble.audit` will both sync and load before it runs. I think we can remove the first two steps from the docs here.

dave-pollock: Ah ok great thanks. I wondered if that were the case. I've updated it. Cheers.

PR #290, opened 2018-02-17
Disable all `find` commands that traverse the whole disk

basepi:
@anuragpaliwal80 @yassingh If you need these checks, we need to figure out how to do them more responsibly, or we need to figure out how to only enable them on DME hosts. Executing a `find` across the whole disk once is irresponsible -- it will peg the CPU for a long time, especially on big disks. But doing it 6 times is unacceptable.

My preference would be to remove these functions altogether, but for now I just make them fail fast.

### DIFF
--- a/hubblestack/files/hubblestack_nova/misc.py
+++ b/hubblestack/files/hubblestack_nova/misc.py
@@ -94,7 +94,14 @@ def audit(data_list, tags, debug=False, **kwargs):
                 kwargs = tag_data.get('kwargs', {})
 
                 # Call the function
-                result = function(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


basepi:
I am going to merge this for now, though I think it was a bit of a knee-jerk reaction to my midnight crisis last week. Obviously if we want to be fully CIS compliant we need to figure out a way to run this check, I'd just like to work to minimize impact. Ideally I'd like to limit it to maybe 20% CPU even while the server is idling, so we don't even show up on monitoring graphs. This is probably not possible until we add multiprocessing to hubble so we can better manage the child processes.

PR #293, opened 2018-02-22
Don't tell splunk to index more things

jettero:
There probably isn't ever a good time to tell Splunk to index extra fields.

The changes to the indexed fields intended by the removed code have to be
accompanied by changes to the fields configuration in Splunk or oddball
problems will occur (e.g., duplicate field indexing). The best plan (if you
really really need extra index fields) is to ask the Splunk admins to add them
to the index through their usual methods.

This affords them the opportunity to say, "that's not a good idea because ____"
and prevents the data from getting wonky in their indexes.

There are certainly fields one might wish to add to the index. It's best to
coordinate them carefully with the Splunk admins. (And these fields are more
rare than you'd think if you're accustomed to traditional databases.)

### DIFF
--- a/hubblestack/extmods/returners/splunk_nebula_return.py
+++ b/hubblestack/extmods/returners/splunk_nebula_return.py
@@ -74,14 +74,6 @@ def returner(ret):
             timeout = opts['timeout']
             custom_fields = opts['custom_fields']
 
-            # Set up the fields to be extracted at index time. The field values must be strings.
-            # Note that these fields will also still be available in the event data
-            index_extracted_fields = ['aws_instance_id', 'aws_account_id', 'azure_vmId', 'azure_subscriptionId']
-            try:
-                index_extracted_fields.extend(opts['index_extracted_fields'])
-            except TypeError:
-                pass
-
             # Set up the collector
             hec = http_event_collector(http_event_collector_key, http_event_collector_host, http_event_port=http_event_collector_port, http_event_server_ssl=hec_ssl, proxy=proxy, timeout=timeout)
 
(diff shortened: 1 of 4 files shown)


jettero:
It seems reasonable to me to remove the options and interfaces all at once and rely on tickets with the splunk admins to add index fields. Not everyone would agree though. (lemme know).

I also failed to test this in any way (long story). I can check this a little later today.

basepi:
This was added by the security team to improve their searches. So this will need sign-off (and probably discussion) from both @mew1033 and @fossam

jettero:
man, it took me a while to figure out that I pushed this branch to upstream instead of my own github repo

jettero:
What did we decide to do with this? I recall saying we should take out the aws_ auto-populate items. And I recall saying we should leave the config option in, but add config file warning or something. I also recall we were going to talk to the splunk admins before we did anything with this.

basepi:
@jettero I think the decision was to leave the option for index-extracted fields so that users of hubble could use it if they want (and we may continue to use it), but default it to an empty list instead of forcing our list.

jettero:
OK, so I'll alter this so it's just removing the forced-on-default fields, but it'll still be pending discussion.

mew1033:
I'm working with one of our Splunk contractors to figure out exactly what we're going to do. I'll update here once we come up with something.

jettero:
Seems to me moving the `aws_*` to `cloud_*` and putting in a grain solved the problem neatly. Should we just close this one?

mew1033:
We need to make some additional changes. I'm not 100% sure what, but for starters, we need to pull the `index_extracted_fields = ['aws_instance_id', 'aws_account_id', 'azure_vmId', 'azure_subscriptionId']` line out of the returners. Instead of hard coding those fields (which have different names now anyway), we should just take what's in the config only. Then we need to adjust our default config to include the values we want.

We also might need to change the code so that index extracted fields are _removed_ from the event payload. That's the part I'm not sure on yet.

jettero:
continued in new PR #323

PR #295, opened 2018-02-26
Win pkgcreator upgrade

hyzer801:
Updated the Powershell scripts to make them more robust.

### DIFF
--- a/pkg/windows/build_pkg.ps1
+++ b/pkg/windows/build_pkg.ps1
@@ -70,8 +70,22 @@ if (!(Test-Path '.\dist\hubble\etc\hubble')) {
 if($default){
     $confFile = C:\temp\hubble\pkg\windows\hubble.conf
 }
+if($confFile){
+    while(!(test-path $confFile)){
+
+        write-host "The path you suppplied doesn't exists. Please enter a correct path."
+        $confFile = read-host
+    
+    }
+}
 else{
     $confFile = read-host "Please specify the full file path to the .conf file you would like to use."
+    while(!(test-path $confile)){
+
+        write-host "The path you suppplied doesn't exists. Please enter a correct path."
+        $confFile = read-host
+    }
+    
 }
 Copy-Item $confFile -Destination '.\dist\hubble\etc\hubble\'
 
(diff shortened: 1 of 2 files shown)


pkg/windows/build_pkg.ps1
```diff
@@ -41,12 +55,29 @@ if (!($modified)) {
 # Run pyinstaller
 pyinstaller .\hubble.spec
 
-# Copy hubble.conf to correct location
+# Checks to see if a conf file has been supplied. If not, it prompts the user for a file path then Copies the hubble.conf to correct location
```
madchills: Can you make this into 2 lines?  It runs off the page here

pkg/windows/build_pkg.ps1
```diff
@@ -73,15 +108,17 @@ If (Test-Path "C:\Program Files (x86)") {
     Invoke-WebRequest -Uri 'http://repo.saltstack.com/windows/dependencies/32/vcredist_x86_2008_mfc.exe' -OutFile "$instDIR\vcredist.exe"
 }
 
-
 # Build Installer
 if ($version -eq $null) {
 	$gitDesc = git describe
 	if ($gitDesc -eq $null) {
-		$version = 'Beta'
+		$version = read-host "What would you like to name this build?"
```
madchills: This is good, but add a condition that if default is true, have it get the most recent tag. Otherwise if it isn't specified, ask the user

pkg/windows/build_pkg.ps1
```diff
@@ -73,15 +108,17 @@ If (Test-Path "C:\Program Files (x86)") {
...
 	} else {
 		$version = $gitDesc
-	}
+    }
+    else {
```
madchills: the else statement isn't needed if it is just contains a continue.  removing it will do the same thing

pkg/windows/create_build_env.ps1
```diff
@@ -51,49 +58,74 @@ if (!($7zip)) {
...
-7z x PortableGit.7z.exe -o"$path\hubble\PortableGit" -y
-& "$path\hubble\PortableGit\post-install.bat"
+Move-Item -Path $ChocoTools\git\ -Destination C:\Temp\hubble\PortableGit\ 
 
 # Install osquery for executible
 if (!(Test-path C:\ProgramData\osquery)) {
 	choco install osquery -y
 } else {
-	choco update osquery -y
+	choco upgrade osquery -y
```
madchills: if update is being deprecated.  please change update to upgrade everywhere we are updating the version of a package with chocolatey

PR #298, opened 2018-02-28
make a global out of __context__ and pass it to the module loader

jettero:
This change is needed to preserve `pulsar.queue`, `pulsar.notifier` ... and the `cp.fileclient_########`.

I'm still testing it under `grains_refresh_frequency: 10` in my hubble config; but it seems to do the job.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -442,6 +442,10 @@ def refresh_grains(initial=False):
     global __salt__
     global __pillar__
     global __returners__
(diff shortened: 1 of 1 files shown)


basepi:
Ooh, good call! This should work nicely, I totally missed that you could pass in `__context__`

jettero:
Yeah, went source diving in $NA/loader.py : minion_mods() and thought .oO( is it really this easy? )

PR #301, opened 2018-03-06
Automatically remove old gitfs locks

basepi:


mew1033:
👍

fossam:
Have we had chance to test this? ( to verify that it fixes the problem we're after )

basepi:
Yes, I used `kill -9` to get Hubble to leave a lockfile and then tested to make sure this code removes it and that Hubble goes back to running fine.

PR #302, opened 2018-03-08
Fixing exception on Debain 7

anuragpaliwal80:
On Debain 7:

root@eqx-temp7:/var/log# cat hubble
2018-03-06 22:02:14 [hubblestack.daemon][INFO ] Setting up the fileclient/fileserver
2018-03-06 22:02:24 [hubblestack.daemon][INFO ] Starting main loop
2018-03-06 22:02:30 [hubblestack.daemon][ERROR ] Error executing schedule
File "hubblestack/daemon.py", line 95, in main
File "hubblestack/daemon.py", line 210, in schedule
File "/opt/hubble/hubble-libs/hubblestack/extmods/modules/pulsar.py", line 274, in process
sub['stats'] = salt'file.stats'
File "/opt/hubble/hubble-libs/salt/modules/file.py", line 3375, in stats
raise CommandExecutionError('Path not found: {0}'.format(path))
CommandExecutionError: Path not found: /etc/hubble/fim_canary.tmp

### DIFF
--- a/hubblestack/extmods/modules/pulsar.py
+++ b/hubblestack/extmods/modules/pulsar.py
@@ -271,7 +271,10 @@ def process(configfile='salt://hubblestack_pulsar/hubblestack_pulsar_config.yaml
                     sub['checksum'] = __salt__['file.get_hash'](pathname, sum_type)
                     sub['checksum_type'] = sum_type
                 if config.get('stats', False):
-                    sub['stats'] = __salt__['file.stats'](pathname)
+                    if os.path.exists(pathname):
+                        sub['stats'] = __salt__['file.stats'](pathname)
+                    else:
+                        sub['stats'] = {}
 
                 ret.append(sub)
             else:


hubblestack/extmods/modules/pulsar.py
```diff
@@ -271,7 +271,10 @@ def process(configfile='salt://hubblestack_pulsar/hubblestack_pulsar_config.yaml
                     sub['checksum'] = __salt__['file.get_hash'](pathname, sum_type)
                     sub['checksum_type'] = sum_type
                 if config.get('stats', False):
-                    sub['stats'] = __salt__['file.stats'](pathname)
+                    if os.path.exists(pathname):
+                        sub['stats'] = __salt__['file.stats'](pathname)
+                    else:
```
sumanmehta: This else block seems unnecessary and os.path.exists(pathname) can be clubbed with upper if statement.

anuragpaliwal80: Actually we need to set the value of sub['stats'] in all cases either with actual value or as {}.

sumanmehta: ok

fossam:
How is this impacting only Debian7?

basepi:
I actually don't think it is. But maybe Debian fails harder? In any case, it's good error handling for any platform.

PR #303, opened 2018-03-09
output file for single run function results

abhikhattri:
When running single run functions, user can give an optional output file path where the results will be written (in json form).

New option -o <file-path> has been added for this purpose.
Example command: hubble hubble.audit -o C:\Users\HubbleTest\auditout.txt

In the above command, the results of hubble.audit will be writted to the mentioned file.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -312,6 +312,14 @@ def run_function():
     else:
         print(ret)
 
+    if(__opts__['outfile']):
+        log.info('Writing output to file {0}'.format(__opts__['outfile']))
+        try:
+            outputfile = open(__opts__['outfile'], "w")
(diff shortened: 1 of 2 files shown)


hubblestack/daemon.py
```diff
@@ -312,6 +313,15 @@ def run_function():
     else:
         print(ret)
 
+    if(__opts__['outfile']):
+        log.info('Writing output to file {0}'.format(__opts__['outfile']))
+        try:
+            outputfile = open(__opts__['outfile'], "w")
```
anuragpaliwal80: Lets use with option rather than open.
Lets not dump info into log file as it's only for console.

anuragpaliwal80:
Closing this will open a new one.

PR #305, opened 2018-03-09
Introducing splay randomizer in hubble

MoodyMudit:


hubblestack/daemon.py
```diff
@@ -214,6 +215,22 @@ def schedule():
             log.error('Scheduled job {0} has an invalid value for seconds or '
                       'splay.'.format(jobname))
         args = jobdata.get('args', [])
+        if 'randomizeSplay' in jobdata and jobdata['randomizeSplay']:
```
yassingh: an example would be useful here. Please add a sample config around line 147 of this file ( in the schedule part of the example config) .

MoodyMudit: Done

fossam:
If this isn't ready, let us know.

MoodyMudit:
It is ready for review. Thanks.

hubblestack/daemon.py
```diff
@@ -214,6 +231,22 @@ def schedule():
...
+           if 'buckets' not in jobdata or int(jobdata['buckets'])<=0:
+              buckets = int(256)
+           else:
+              buckets = int(jobdata['buckets'])
+           try:
+              hostname = socket.gethostbyname(socket.gethostname())
+              ipValues = hostname.split('.')
+              ipValueSum = int(ipValues[0])+int(ipValues[1])+int(ipValues[2])+int(ipValues[3])
```
basepi: I don't know if summing the ip addresses will actually result in a normal distribution of servers. Why not just use the random library to let it decide the normal distribution?

```
bucketnumber = random.randint(1, buckets + 1)
```

MoodyMudit:
The primary intent of these changes is to **not allow hubble to execute its processes on more than a certain number of machines**. By forming buckets and placing only a handful number of machines in each bucket, where each bucket executes at a different time, we make sure that hubble does not eat up the common hardware's resources. The IP of virtual machines in an industry are usually consecutive(not guaranteed though), meaning that there is an even distribution of machines in each bucket.

Moreover a larger value of splay does not guarantee that hubble processes necessarily run at different time on different machines. It could easily create a resource crunch on the underlying hardware if hubble runs on say 500 machines concurrently. Also this resource crunch would come as a surprise when hubble is restarted/updated and the value of splay gets reset. To fix this hubble would again have to be restarted, resetting the value of splay across all machines, which could again lead to the same setting and creating a resource crunch all over again. 

I do realize that the chances of such a setting are very low, but even if this happens only once in say 2-3 years, it would be very disruptive for the business. Hence the proposal of segregating the machines beforehand in buckets so that hubble processes are executed in a controlled manner.

hubblestack/daemon.py
MoodyMudit: The intent here is to place the machines evenly in all buckets. A random bucket number could lead to a situation where a lot of machines are put in a single bucket and a lot of buckets are empty. 

The IP was chosen here because usually in an industry VMs have consecutive IPs(not guaranteed though). This would help in the normal distribution of machines in the buckets.

fossam:
(neutral statement on the issue at hand)
@MoodyMudit thanks for writing down your reasoning for this feature. Explanations like this make working on the reviews much better.

basepi:
I should also note that you should calibrate the number of buckets based on the possible ip address distribution. For example, if you used 200 buckets, you'd probably end up with more hosts in the first 55 buckets, due to ip addresses going from 0 to 255. Modulus doesn't fix those distribution issues. A power of 2 is probably the best bet.

MoodyMudit:
What we are looking to guarantee is that **hubble processes should not execute on more than a few machines.** For this we need to place the machines in different buckets and take up these buckets for execution one at a time. We are not looking for a perfect distribution, but only for a distribution that would keep the machine count per bucket to a minimum. This would help us to distribute the load evenly throughout the scheduled period. 

For example, we configure our Nova to run once a day. let us say I want to run hubble on 10,000 machines. For this I keep the bucket count to 200(supplied externally through config). Now this setup would mean that I have 10,000 machines divided in 200 buckets(roughly 50 machines/bucket) and each bucket is guaranteed to run at a different time, so that roughly 50 machines(and not 10,000) are executing hubble processes simultaneously. 

Further, we have no specific requirement of keeping a timestamp based execution. On the contrary, keeping it diversified even for a bucket helps us to further reduce the load on the hardware. Rather than starting all the machines in a bucket simultaneously it is better to now start them randomly since the machine count is now way less than the total count.

This feature can be thought of as an extension of splay that enables us to use different ranges of splay rather than using the <0 to splay> range. Each bucket now uses the range <min_splay to min_splay+splay> where min_splay if different for all the buckets, essentially guaranteeing that hubble processes do not run simultaneously. 

Lastly, as I said, we aren't looking for a perfect distribution, so now even if some buckets contain a higher number of machines, this number would not be much higher. As an example, assuming that our machines have IP ranges x.y.z.[0-255] and I choose 200 buckets, the first 55 buckets will now have one machine more than the rest. This suits us and is not a problem since we are more concerned about the load on the hardware and not on the distribution pattern. The setting where a lot of IP addresses clump to a single bucket seems to be very rare and would probably be possible only via manual intervention.

basepi:
> Further, we have no specific requirement of keeping a timestamp based execution. On the contrary, keeping it diversified even for a bucket helps us to further reduce the load on the hardware. Rather than starting all the machines in a bucket simultaneously it is better to now start them randomly since the machine count is now way less than the total count.

The problem with this logic is that the randomization could cause buckets to overlap. As soon as you're not starting at the same time or pegging to a timestamp, the advantage of having buckets goes away, because your bucket 1 hosts could be executing at the same time as your bucket 2 hosts if they start a little later. In fact, over time you could end up with more buckets overlapping, potentially resulting in the very grouping that you were worried about when using randomized splays.

I understand what the code is doing and the motivation behind it, but I think the assumptions you're making cancels out the advantages. You didn't want to use `random` because you would be assuming a slight amount of risk due to the possible (though extremely improbable) situation where it didn't distribute the hosts well. By allowing your buckets to overlap (which will happen without a timestamp anchor), **you're assuming the same risk**. This is why I am pushing back on this feature as it is currently written. It adds additional complexity without actually reducing risk.

mew1033:
My vote is to favor using the `random` module over the sum of IP address parts. Python utilizes `os.urandom` where available, which is about as random as you can get. It also ensures a uniform distribution of random numbers.

fossam:
The spread is not as even with this method while I've seen surprisingly even spreads with random numbers when there are tens of thousands of hosts. (just don't seed your rng with time). The problem with this method is dependency on ip address distribution (sum of the octets is not random at all and too many hosts clump together).

But on to the problem at hand - when we're trying to shift execution time around, I think this speaks to the issue of perhaps being not as efficient as we could be.
@MoodyMudit do you know how many simultaneous Hubble instances executing presents a potential problem with underlying hardware resources? 
Perhaps there are things we can do to have less of an impact on resources.

MoodyMudit:
I understand that we have two problems currently.
1.	The sum of IP approach will not distribute the machines evenly.
2.	There is an overlapping of buckets with the current implementation.
So, I am planning to change the implementation to the following:
	**sum = IP[0] * 256^3 + IP[1] * 256^2 + IP[2] * 256 + IP[3]
	bucket_number = sum % buckets
	min_splay = int((seconds/buckets) * bucket_number)
	splay = min(splay, seconds / buckets) (This prevents overlapping)
	where buckets is provided externally through hubble config**
This solves both the problems.
I created a test script to simulate the distribution of 65,536 machines and compare the random distribution with the splay randomizer distribution using the above logic. 
Scenario 1: Random Distribution
The IPs are in the range x.y.[0-255].[0-255]. 
Process = hubble_audit
Splay = 86,400
Seconds = 86,400
Estimated time hubble_audit takes to run = 10 seconds
The following is the data for a 10 second window using a random distribution.
*********************************
At splay = 383 and count is 3
At splay = 384 and count is 1
At splay = 385 and count is 5
At splay = 386 and count is 1
At splay = 387 and count is 2
At splay = 388 and count is 0
At splay = 389 and count is 2
At splay = 390 and count is 1
At splay = 391 and count is 1
At splay = 392 and count is 0

Total Machines in a 10 second window = 16
*********************************
At splay = 452 and count is 3
At splay = 453 and count is 2
At splay = 454 and count is 2
At splay = 455 and count is 3
At splay = 456 and count is 1
At splay = 457 and count is 1
At splay = 458 and count is 1
At splay = 459 and count is 0
At splay = 460 and count is 1
At splay = 461 and count is 2

Total Machines in a 10 second window = 16
*********************************
image
This is the distribution of sum of machines in a 10 sec window

Scenario 2: Splay Randomizer
The IPs are in the range x.y.[0-255].[0-255]. 
Process = hubble_audit
Splay = 0
Seconds = 86,400
Buckets = 8,000
Estimated time hubble_audit takes to run = 10 seconds

Using this method, 
1.	Every bucket will now run at an interval of 10 seconds. 
2.	Every bucket contains only 8/9 machines. 
Here is the distribution pattern using this method.

*********************************
bucket no = 0 and count is 8
bucket no = 1 and count is 8
bucket no = 2 and count is 8
bucket no = 3 and count is 8
bucket no = 4 and count is 8
bucket no = 5 and count is 8
bucket no = 6 and count is 8
bucket no = 7 and count is 8
bucket no = 8 and count is 8
bucket no = 9 and count is 8

*********************************

bucket no = 959 and count is 8
bucket no = 960 and count is 9
bucket no = 961 and count is 9
bucket no = 962 and count is 9
bucket no = 963 and count is 9
bucket no = 964 and count is 9
bucket no = 965 and count is 9
bucket no = 966 and count is 9
bucket no = 967 and count is 9
bucket no = 968 and count is 9

*********************************
image

Here if we had a cap of 15 machines and wanted that no more than 15 machines should execute hubble concurrently, then the second scenario would have fared better.

basepi:
> splay = min(splay, seconds / buckets) (This prevents overlapping)

I think this is where our primary misunderstanding is.

Assuming equal start times, your old configuration would prevent bucket overlapping. What I'm saying is that even if hubble is started at the same time on every host initially, it will slowly drift over time. This is because new hosts will be spun up, hosts will be restarted, hubble will be restarted -- normal SRE work will cause many hubble instances to be started at different times.

If a group of hosts were in bucket 3, then when they're restarted they will still be bucket 3, but bucket 3's runtime will be different, because it's based on hubble's start time. So what if a non-trivial percentage of bucket 3 ends up executing during bucket 7's time slot? Now you have an overlap problem that is not solved with your code above.

--------------

I want to be very clear that I don't have a problem with buckets on principle. The thing I have a problem with is trying to fix `random`'s potential for grouping by writing a feature that can group in other ways, as shown above. At that point you're adding complexity without fixing the problem, which is why I recommended anchoring to a timestamp.

I should also note that your data shows that the distribution of hosts is pretty consistent using `random`. In both samples you ended up with 16 hosts per 10 second period, which is a tiny amount considering your fleet of 65k servers. Granted, if you divide it out, `(65536/86400)*10`, your average hosts per 10 seconds will be 8, so you have found some small groupings. If that number is unacceptable, you could double the splay and your number of hosts per 10 second period would average at ~4 and your potential for grouping would go down even further. That's why I think `random` is sufficient -- if you take the splay high enough, you can reduce your risk to effectively zero.

One final note: hubble is designed to be fairly lightweight. I think you may be overestimating the impact that it will have if many hosts run it at once. I think this is less of an issue than you think it is. Even if we assume hubble is a heavyweight process that will cause impact on the running servers: If 16 hosts out of 65k run at once and you notice impact, then you may have trouble anyway. That's only 0.02% -- what happens if you have a 0.02% uptick in traffic? :)

-----------------

**tl;dr**: I still don't think your changes solve the potential for grouping. If we want to do buckets we need a way to guarantee that a bucket's time slot doesn't shift, which requires using a timestamp anchor. Otherwise we should just use `random` and use a high enough splay that the risk of substantial grouping drops to effectively zero.

basepi:
Hey @MoodyMudit, we haven't heard from you in a couple of weeks. Did your team decide that the existing splay is sufficient?

MoodyMudit:
Hi basepi,
I am working on a different approach to my problem here. Will raise another PR when ready. Thanks for your suggestions and ideas.

PR #306, opened 2018-03-12
print output of single run functions in json format

abhikhattri:
This change is for printing the result of single run function in json format when the function is run with -p arguement, eg:

hubble hubble.audit -vv -p

(By default 'no_pprint' flag is True, hence pprint (pretty print) is used. -p parameter sets no_pprint flag to false and hence print method of python is used.)


abhikhattri:
Closing this PR as another solution will be implemented for this.

PR #307, opened 2018-03-12
argument to print output of single function run in json format

abhikhattri:
Added an optional parameter -j for printing the output of single run function in json format. If this parameter is not given when running any function, existing functionality works.

Example command:
hubble hubble.audit -vv -j

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -14,6 +14,7 @@ import random
 import signal
 import sys
 import uuid
+import json
 
 import salt.fileclient
 import salt.utils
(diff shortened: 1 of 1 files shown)


mew1033:
With nebula/osquery, should we use the osquery json output instead?

abhikhattri:
This gives the same output as we would expect from osquery json format. 

Also, as per nebula implementation, the result of osqueryi output is parsed to list of dictionaries, additional information is added to output and then is printed to console inside run_function() in daemon.py. Therefore, we need to explicitly convert it to json format prior to printing.

`cmd = [__grains__['osquerybinpath'], '--read_max', MAX_FILE_SIZE, '--json', query_sql]`
`res = __salt__['cmd.run_all'](cmd)`
`if res['retcode'] == 0:
            query_ret['data'] = json.loads(res['stdout'])`

PR #308, opened 2018-03-15
module to push hubble config to splunk

abhikhattri:
This module publishes the hubble configurations to splunk. 
The frequency of publishing can be set via schedule config as we do for other modules.

Below is a sample schedule for pushing the config once per day:

schedule:
  conf_publisher:
    function: conf_publisher.publish
    seconds: 86400
    splay: 5
    run_on_start: True

To avoid confidential/sensitive info being sent to splunk, we can omit certain key-value pairs of config by giving the patterns in config. All keys containing any of the patterns would be omitted from the pushed event.

hubblestack:
  conf_publisher:
    patterns_to_skip: password,token

### DIFF
--- a/hubblestack/extmods/modules/conf_publisher.py
+++ b/hubblestack/extmods/modules/conf_publisher.py
@@ -1,0 +1,53 @@
+
+import logging
+import hubblestack.splunklogging
+import copy
+import time
+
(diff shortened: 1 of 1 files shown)


abhikhattri:
Closing thjs. Will create new one.

PR #309, opened 2018-03-15
module to push hubble config to splunk

abhikhattri:
This module publishes the hubble configurations to splunk.
The frequency of publishing can be set via schedule config as we do for other modules.

Below is a sample schedule for pushing the config once per day:

```
schedule:
  conf_publisher:
    function: conf_publisher.publish
    seconds: 86400
    splay: 3600
    run_on_start: True
```

### DIFF
--- a/hubblestack/extmods/modules/conf_publisher.py
+++ b/hubblestack/extmods/modules/conf_publisher.py
@@ -1,0 +1,46 @@
+
+import logging
+import hubblestack.splunklogging
+import copy
+import time
+
+log = logging.getLogger(__name__)
+
+def publish():
+    log.info('Started publishing config to splunk')
+    
+    initialize_splunklogging()
+    filtered_conf=filter_config()
+    class MockRecord(object):
+            def __init__(self, message, levelname, asctime, name):
+                self.message = message
+                self.levelname = levelname
+                self.asctime = asctime
+                self.name = name
+
+    handler = hubblestack.splunklogging.SplunkHandler()
+    handler.emit(MockRecord(filtered_conf, 'INFO', time.asctime(), 'hubblestack.hubble_config'))
+    log.info('Published config to splunk')	
+
+def initialize_splunklogging():
+    hubblestack.splunklogging.__grains__ = __grains__
+    hubblestack.splunklogging.__salt__ = __salt__
+
+#Filter out keys containing certain patterns to avoid sensitive information being sent to splunk
+def filter_config():
+    opts_copy = copy.deepcopy(__opts__)
+    patterns_to_filter = ["password", "token"]
+    filtered_conf = remove_sensitive_info(opts_copy, patterns_to_filter)
+    return filtered_conf
+
+#Recursively removes key value pairs where key contains any of patterns_to_filter
+def remove_sensitive_info(obj, patterns_to_filter):
+    if isinstance(obj, dict):
+         obj = {
+             key: remove_sensitive_info(value, patterns_to_filter)
+             for key, value in obj.iteritems()
+             if not any(patt in key for patt in patterns_to_filter)}
+    elif isinstance(obj, list):
+         obj = [remove_sensitive_info(item, patterns_to_filter)
+                    for item in obj]
+    return obj


hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -0,0 +1,46 @@
+
+import logging
+import hubblestack.splunklogging
+import copy
+import time
+
+log = logging.getLogger(__name__)
+
+def publish():
```
basepi: Could you add some docstrings to these functions with some basic documentation about what they do? It's good to be consistent and document wherever possible.

hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -0,0 +1,46 @@
...
+
+    handler = hubblestack.splunklogging.SplunkHandler()
+    handler.emit(MockRecord(filtered_conf, 'INFO', time.asctime(), 'hubblestack.hubble_config'))
+    log.info('Published config to splunk')	
+
+def initialize_splunklogging():
+    hubblestack.splunklogging.__grains__ = __grains__
+    hubblestack.splunklogging.__salt__ = __salt__
+
+#Filter out keys containing certain patterns to avoid sensitive information being sent to splunk
```
basepi: These comments could be moved into docstrings for the function -- that's the accepted standard for python.

hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -0,0 +1,46 @@
...
+def filter_config():
+    opts_copy = copy.deepcopy(__opts__)
+    patterns_to_filter = ["password", "token"]
+    filtered_conf = remove_sensitive_info(opts_copy, patterns_to_filter)
+    return filtered_conf
+
+#Recursively removes key value pairs where key contains any of patterns_to_filter
```
basepi: Another candidate for a docstring

hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -0,0 +1,46 @@
+
+import logging
+import hubblestack.splunklogging
+import copy
+import time
+
+log = logging.getLogger(__name__)
+
+def publish():
+    log.info('Started publishing config to splunk')
```
basepi: Can we use debug log here instead of info? I don't think this is information that we need to log by default.

hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -0,0 +1,46 @@
...
+    class MockRecord(object):
+            def __init__(self, message, levelname, asctime, name):
+                self.message = message
+                self.levelname = levelname
+                self.asctime = asctime
+                self.name = name
+
+    handler = hubblestack.splunklogging.SplunkHandler()
+    handler.emit(MockRecord(filtered_conf, 'INFO', time.asctime(), 'hubblestack.hubble_config'))
+    log.info('Published config to splunk')
```
basepi: Another log that should probably be debug instead of info

hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -0,0 +1,46 @@
...
+                self.asctime = asctime
+                self.name = name
+
+    handler = hubblestack.splunklogging.SplunkHandler()
+    handler.emit(MockRecord(filtered_conf, 'INFO', time.asctime(), 'hubblestack.hubble_config'))
+    log.info('Published config to splunk')	
+
+def initialize_splunklogging():
+    hubblestack.splunklogging.__grains__ = __grains__
+    hubblestack.splunklogging.__salt__ = __salt__
```
basepi: Is this function necessary? Since it's only two lines I wonder if we should just move these two lines up into `publish()`

hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -0,0 +1,46 @@
...
+import copy
+import time
+
+log = logging.getLogger(__name__)
+
+def publish():
+    log.info('Started publishing config to splunk')
+    
+    initialize_splunklogging()
+    filtered_conf=filter_config()
```
basepi: I think you're going to find that `__opts__` is huge, even after filtering out tokens/passwords. What would you think about having `publish()` take a list of opts to log? That way you can just collect the data you want and keep the noise level low?

basepi:
Should be fine. The code specifically filters out our tokens.

That said, I would still recommend against running this with no args.

basepi:
I pushed a couple of commits to your branch (Github allows for this on branches with open PRs by default) fixing a couple of code lint issues, so it should pass the pylint checks that we're planning to add soon. If it looks good and @sumanmehta approves, we can get this merged.

PR #314, opened 2018-03-20
Clear cache of azurefs containers which are not accessible anymore

yassingh:
If an azurefs container is not accessible for hubble host, it's cache folder is deleted so that the stale files are not served from that container anymore.

NOTE: In case all the containers are inaccessible, cache folder will deleted for all containers. But in this case, the code will not reach the part where 'hashes' directory is deleted. As a result, hubble will keep serving the file from the last azurefs state if all the containers are inaccessible.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -375,6 +375,7 @@ def load_config():
     salt.config.DEFAULT_MINION_OPTS['default_include'] = 'hubble.d/*.conf'
     salt.config.DEFAULT_MINION_OPTS['logfile_maxbytes'] = 100000000 # 100MB
     salt.config.DEFAULT_MINION_OPTS['logfile_backups'] = 1 # maximum rotated logs
+    salt.config.DEFAULT_MINION_OPTS['delete_on_unavailable'] = False
 
     global __opts__
 
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -375,7 +375,7 @@ def load_config():
     salt.config.DEFAULT_MINION_OPTS['default_include'] = 'hubble.d/*.conf'
     salt.config.DEFAULT_MINION_OPTS['logfile_maxbytes'] = 100000000 # 100MB
     salt.config.DEFAULT_MINION_OPTS['logfile_backups'] = 1 # maximum rotated logs
-    salt.config.DEFAULT_MINION_OPTS['delete_on_unavailable'] = False
+    salt.config.DEFAULT_MINION_OPTS['delete_inaccessible_azure_containers'] = False
 
     global __opts__
 
--- a/hubblestack/extmods/fileserver/azurefs.py
+++ b/hubblestack/extmods/fileserver/azurefs.py
@@ -63,6 +63,7 @@ import salt.utils
 try:
     import azure.storage.common
     import azure.storage.blob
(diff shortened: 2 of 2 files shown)


basepi:
Thank you for making this default to False. And I really like the fact that it continues working in case all of them are inaccessible, because otherwise this could be problematic during a network outage.

basepi:
Thank you for making this default to False. I'm not sure that your evaluation of the "if all are inaccessible" situation is correct, however. The hashes directory won't be lost which means that files that have been previously served will continue to be served correctly. But new files will not. What happens in the case of a network outage? Is the error different from `<AuthenticationErrorDetail>Signature did not match.`, saving us from accidental deletion in this case?

yassingh:
>> What happens in the case of a network outage? 

I tried to hinder the internet connectivity ( by creating a fake dns server in /etc/resolv.conf ).
The exception message was different in this case, something like 
```Failed to establish a new connection...```
And containers were not affected anyhow in this case.
If you can think of some other way(s) to test it, do let me know. I will check with them as well.



>>  I'm not sure that your evaluation of the "if all are inaccessible" situation is correct, however....But new files will not.

If even a single container is accessible, 'hashes' folder will be deleted. So the updated files will be served. 
'hashes' folder will not be deleted only in one case, i.e. when all the containers are inaccessible. And in this case, it would be best to serve stale files. Fortunately, this is already happening with current code.

yassingh:
Let me explain a little more about this change and why we need it.

1. We want to use two ( or more ) azurefs containers for fallback/backup/safety. Idea is that if we deactivate first container ( c1 ) , hubble should start picking files from second  container ( c2 ). 
This does not happen as of now. Right now, if you rotate the storage key of c1, or delete c1 altogether from azure, hubble still keeps serving files from c1 ( because the local cached folder of c1 is not deleted).
So, hubble never switches to c2. Hubble keeps serving stale files from (deleted ) c1 container's cache.

In order to achieve the first task, we have decided to remove the cached folder of c1 ( or any other container that is not accessible ).

2. Next thing we want is that in case all the containers are inaccessible ( because of network glitch perhaps ), hubble should not stop working. Hubble should serve stale/cached profiles in that case.

For this second task, we are relying on the below mentioned logic:

Azurefs' 'hashes' directory is deleted whenever a container's contents are updated from internet. Updated files are served only if 'hashes' directory is deleted. If 'hashes' folder is  not deleted, hubble keeps picking up the files from '.../files/base' directory. 
Now, in this second case, all the containers are unreachable. So, an exception is thrown for each container, and the for loop ( after catching exception ) would continue without executing the rest of code. Since the code for deletion of 'hashes' directory is never reached, 'hashes' folder is never deleted. So, hubble will keep picking old files from '.../files/base' directory although the cached folder for all the containers have been deleted.

I have verified this.

yassingh:
```<AuthenticationErrorDetail>Signature did not match.``` messages appears in both the cases where we want to delete the cache:
1. When we rotate the key of storage container's account
2. When we delete the storage container for azure cloud.

This error message does not appear if there is network connectivity issue.


So, my code change is based on this 'string' logic. 
I could not think of any other better solution.
If you have some other idea for this task, let me know. I can give it a try.

basepi:
Thanks for the clarifications. I think this is fine.

PR #315, opened 2018-03-21
Adding match_on_file_missing flag for stat module

anuragpaliwal80:
Added match_on_file_missing flag for stat module


hubblestack/files/hubblestack_nova/stat_nova.py
```diff
@@ -101,14 +109,16 @@ def audit(data_list, tags, debug=False, **kwargs):
                 if not salt_ret:
                     if None in expected.values():
```
basepi: I know this is not your fault, but I happened to notice this and I think that it's wrong.

I think this is what the author meant to do:

```
if not expected:
```

Otherwise a single argument with `None` will result in success. Can you evaluate this and make that change if you agree?

hubblestack/files/hubblestack_nova/stat_nova.py
```diff
@@ -276,4 +286,4 @@ def _is_permission_in_limit(max_permission, given_permission):
     if given_x and (not allowed_x):
         return False
 
-    return True
+    return True
```
basepi: Looks like we lost the trailing newline for some reason. Can you add it back in?

hubblestack/files/hubblestack_nova/stat_nova.py
anuragpaliwal80: I agree, I will make these changes as well.

hubblestack/files/hubblestack_nova/stat_nova.py
anuragpaliwal80: sure

hubblestack/files/hubblestack_nova/stat_nova.py
anuragpaliwal80: I agree, will make these changes as well.

anuragpaliwal80:
I still needs to fix RHEL 6 issue, but will do that.

anuragpaliwal80:
Will fix one more edge case, i.e. if some one uses `allow_more_strict as True` and check for `mode: null ` Right now it throws exception as we are converting the mode value to int inside _is_permission_in_limit function.

anuragpaliwal80:
Done. Please review and let me know if any issue/concern.

PR #316, opened 2018-03-21
upstream changed salt.utils.jid.gen_jid

jettero:
Older Salts exepct 0 arguments: eg, `salt.utils.jid.gen_jid()`
Newer Salts want `__opts__`: eg, `salt.utils.jid.gen_jid(__opts__)`

I instructed daemon.py to adapt to this, but it's a little weird. Is it better to interrogate the Salt import version?

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -32,6 +32,11 @@ __opts__ = {}
 # This should work fine until we go to multiprocessing
 SESSION_UUID = str(uuid.uuid4())
 
+_gen_jid = salt.utils.jid.gen_jid
+
+import inspect
+if len(inspect.getargspec( salt.utils.jid.gen_jid ).args) == 1:
+    _gen_jid = lambda: salt.utils.jid.gen_jid(__opts__)
 
 def run():
     '''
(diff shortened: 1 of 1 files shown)


basepi:
We have a couple of these hitting us with the new version of salt. We should probably just peg to the newer version of salt (to make sure that we're getting bugfixes and CVE fixes) and just deal with the growing pains.

basepi:
Since we control the salt version through our package build process, we don't need to provide backwards compat

jettero:
Oh, we package a version of salt right along with hubble?

How does it not conflict with other versions of salt on the system? I need to go slog around in there I think.

So I should correct this to just use the gen_jid(opts) call directly and expect a 2018.x peg soon instead?

basepi:
This is only for the pyinstaller packages we create with Docker. Because we use pyinstaller, it's the rough equivalent of static linking -- we include python and all the deps in a portable directory that doesn't interfere with existing installs.

basepi:
@jettero Will you modify this to just pass in `__opts__` please? We're just standardizing on 2018.3 I think.

jettero:
I changed it to just always pass in opts... then rebased to develop ... and got a merge conflict on the fast forward.

Seems someone already did the ask. I suggest we just delete or reject this PR

PR #317, opened 2018-03-21
osquery timing + custom splunk loglevel + ssl verify for splunk + small fixes

basepi:
This PR kind of spiraled out of control a little bit. It does a bunch of things.

# Top changes

* Time the osquery queries and report to splunk if splunklogging=True
* Add new loglevel for splunk reporting (`splunk`)
* **Turn on SSL verify for the splunk returners and splunklogging.py**

# Other changes

* Migrates to salt.utils.path.which (salt 2018.3)
* Migrates to salt.utils.platform.is_windows (salt 2018.3)
* Fixes gen_jid calls (salt 2018.3)
* Improves error reporting if net-tools is not installed on centos
* Fixes a transient issue related to salt's custom log records
* Adds network requires to systemd unit file
* Adds optional sleep on startup as a backup for network startup issues

@sumanmehta The SSL verify change is potentially a big one and should be tested against your splunk endpoints. If they don't have good certificates, we should either fix that or make the SSL verify configurable and silence the warnings (or else they get logged to splunk, which triggers a new warning, which gets logged to splunk...bad loop).

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -293,7 +293,7 @@ def schedule():
                     continue
                 log.debug('Returning job data to {0}'.format(returner))
                 returner_ret = {'id': __grains__['id'],
-                                'jid': salt.utils.jid.gen_jid(),
+                                'jid': salt.utils.jid.gen_jid(__opts__),
                                 'fun': func,
                                 'fun_args': args + ([kwargs] if kwargs else []),
                                 'return': ret}
@@ -331,7 +331,7 @@ def run_function():
         else:
             log.info('Returning job data to {0}'.format(returner))
             returner_ret = {'id': __grains__['id'],
-                            'jid': salt.utils.jid.gen_jid(),
+                            'jid': salt.utils.jid.gen_jid(__opts__),
                             'fun': __opts__['function'],
                             'fun_args': args + ([kwargs] if kwargs else []),
                             'return': ret}
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -461,7 +461,8 @@ def load_config():
                                         __opts__['log_level'],
                                         max_bytes=__opts__.get('logfile_maxbytes', 100000000),
                                         backup_count=__opts__.get('logfile_backups', 1))
-    logging.SPLUNK = 25 # logs below warning, above info
+    # splunk logs below warning, above info by default
+    logging.SPLUNK = int(__opts__.get('splunk_log_level', 25))
     logging.addLevelName(logging.SPLUNK, 'SPLUNK')
     def splunk(self, message, *args, **kwargs):
         if self.isEnabledFor(logging.SPLUNK):
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -467,6 +467,12 @@ def load_config():
         if self.isEnabledFor(logging.SPLUNK):
             self._log(logging.SPLUNK, message, args, **kwargs)
     logging.Logger.splunk = splunk
+    if __salt__['config.get']('hubblestack:splunklogging', False):
+        root_logger = logging.getLogger()
+        handler = hubblestack.splunklogging.SplunkHandler()
+        handler.setLevel(logging.SPLUNK)
+        root_logger.addHandler(handler)
+
     # 384 is 0o600 permissions, written without octal for python 2/3 compat
     os.chmod(__opts__['log_file'], 384)
     os.chmod(parsed_args.get('configfile'), 384)
@@ -477,12 +483,6 @@ def load_config():
     if __grains__.get('ip_gw', None) is False and 'fallback_fileserver_backend' in __opts__:
         log.info('No default gateway detected; using fallback_fileserver_backend.')
         __opts__['fileserver_backend'] = __opts__['fallback_fileserver_backend']
-
-    if __salt__['config.get']('hubblestack:splunklogging', False):
-        root_logger = logging.getLogger()
-        handler = hubblestack.splunklogging.SplunkHandler()
-        handler.setLevel(logging.SPLUNK)
-        root_logger.addHandler(handler)
 
 
 def refresh_grains(initial=False):
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -457,6 +457,12 @@ def load_config():
                                         __opts__['log_level'],
                                         max_bytes=__opts__.get('logfile_maxbytes', 100000000),
                                         backup_count=__opts__.get('logfile_backups', 1))
+    logging.SPLUNK = 25 # logs below warning, above info
+    logging.addLevelName(logging.SPLUNK, 'SPLUNK')
+    def splunk(self, message, *args, **kwargs):
+        if self.isEnabledFor(logging.SPLUNK):
+            self._log(logging.SPLUNK, message, args, **kwargs)
+    logging.Logger.splunk = splunk
     # 384 is 0o600 permissions, written without octal for python 2/3 compat
     os.chmod(__opts__['log_file'], 384)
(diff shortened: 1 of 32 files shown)


hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -179,7 +180,10 @@ def queries(query_group,
         }
 
         cmd = [__grains__['osquerybinpath'], '--read_max', MAX_FILE_SIZE, '--json', query_sql]
+        t0 = time.time()
         res = __salt__['cmd.run_all'](cmd)
+        t1 = time.time()
+        log.splunk('osquery query \'{0}\' took {1}'.format(query_sql, t1-t0))
```
fossam: These stats would be good. Would it be possible get them combined in an event, once per hour?
If I understand it right, this would fire 6 times separately every 15 minutes and in addition 10 times separately per hour.

sumanmehta: I was wondering can we not make this information as part of the Splunk event itself. Like we add aws_account_id to every event, we can add query_execution_time to nebula events. Thoughts?

anuragpaliwal80: I think we should have the timing details for every query because that will give us the clear picture about time taken for all individual queries.

basepi: I think @fossam was suggesting that we still track the time for every query, but that we cache it and send it every hour. This is a little bit problematic because I can only send these stats when nebula runs, but I could potentially use a global variable to cache this data and track send times to batch theses. We could end up losing timing data if a grains refresh happens at the wrong time, but that's probably fine since we're just looking for large patterns.

I also do like the idea of adding the timing to the actual nebula events that we send to splunk. The reason I didn't do this initially is because I wasn't sure *where* to add it. A nebula query for running processes, for example, gets split up with a single process per event. Should I add the timing to every one of these events? Or do some sort of summary event? That's why I did this via splunklogging.py.

If splunklogging is enabled, it does result in slower runs because these network calls are not immediate, but I don't think that matters from an end user standpoint -- it's not high CPU. But it is annoying when testing. :) (But for testing you can just disable splunklogging, so it's fine there, too)

fossam: The issue with separate events is that there are many of them and that could put an additional weight on storage requirements. Each event has additional metadata, etc.
`Should I add the timing to every one of these events?`
No. The timing information would be the same for all rows per query output since you can't measure how long it would take to execute each row. These rows become later separate events and therefore the execution time is the same for all of them.
`Or do some sort of summary event?`
Yes, and preferably not with each query, but an aggregation of them over a time period. Though, I think that it'd be ok to insert the length of time to _only_ the last/first event of the query. We'd have to make sure that the field name would not collide with anything, but this would be probably be the most efficient way.

This brings us to the topic of what additional information to include for identifying queries from each other if we were to do a separate event for logging purposes. I think including the complete sql of the query is way too much and unnecessary. Instead, a query_name and a timestamp (t0) would be more preferable and would still do the job of uniquely pairing the length of time the query ran with the query.

sumanmehta: I would prefer adding execution time to the last event of the query. That ways we do not have to fire another event , it would no longer be tied to splunklogging and no repetitive metadata.
@fossam If understand it right, you prefer aggregation in order to reduce number of Splunk events.

mew1033: My vote is to send these in in separate events. Yes, it's an additional event, but it better follows the Splunk model of a single sourcetype being a single type of data. I also like @fossam's idea of aggregating them into a single event. Loosing one off stats due to grain refreshes would be acceptable since, as @basepi said, we're only going to be looking for large patterns.

Also, logging the entire SQL of each query every time seems like a Bad Idea™. We have the full SQL in hubblestack_data, no need to log it to Splunk every time. Just the name as found in hubblestack_data should be enough.

basepi: Good calls. I think it will be a separate summary event. However, if we don't do it on the hubble_log sourcetype I'm not sure where to put it. We have a separate sourcetype per query by default so a summary event has no obvious destination.

I see this as either being a single event on the same sourcetype as the query, or a summary event on hubble_log. Thoughts?

mew1033: I like the summary event in `hubble_log`. This feels more like hubble reporting on itself as opposed to hubble reporting on the system.

basepi: Agreed. I can easily do a single event for each query group, so that's the direction I'll head.

basepi: (I also wanted to quickly address Suman's recommendation to add the timing to the last event for each query. I don't think this follows the design pattern for splunk events very well. If data is present for a query I would expect it to be present for all events of that query. Otherwise an arbitrary event among hundreds of other identical events would have additional data. I think having it as a separate event is the better fit, and I think @mew1033 has a good reason for it to be `hubble_log`.)

basepi:
Ready for final review

hubblestack/extmods/modules/nebula_osquery.py
fossam: In case of summary event, `hubble_log` option would be preferable.
I don't have strong feelings on either option, as long as it can be picked up from somewhere and has enough details to tie it back to a particular query at a particular time.

Just as a side note (don't take it as a preference), there are currently fields in Hubble events which presence only happens under certain conditions - _aws_instance_id_, for example. One might wonder why not another field that is a metric about the query that produced the event. So when I search the sourcetype and filter on the existence of a certain field, I get the query and it's performance. At the same time, it's a good practice to produce events with more consistent fields.

sumanmehta: I am good as long as I can search for query name and it's timing detail. It seems with the current implementation, we will get splunk event like

`timing.running_procs: 100`

while alternative can be -
`query_name: running_procs`
`timing:100`
I am not sure if the current implementation is good for Splunk searches.

hubblestack/daemon.py
```diff
@@ -457,6 +461,18 @@ def load_config():
...
+    def splunk(self, message, *args, **kwargs):
+        if self.isEnabledFor(logging.SPLUNK):
+            self._log(logging.SPLUNK, message, args, **kwargs)
+    logging.Logger.splunk = splunk
+    if __salt__['config.get']('hubblestack:splunklogging', False):
+        root_logger = logging.getLogger()
+        handler = hubblestack.splunklogging.SplunkHandler()
+        handler.setLevel(logging.SPLUNK)
+        root_logger.addHandler(handler)
+
```
sumanmehta: Is it possible to only enable timing information in splunklogging and not enable errors or warnings? Also, can I configure splunklogging for errors only and not for warnings? I am asking because I am interested in timing data, but not sure If I need all the warnings in Splunk.

hubblestack/extmods/modules/nebula_osquery.py
sumanmehta: Never mind. I just looked again at the implementation again, we now have summary event, so please ignore my previous comment.

hubblestack/daemon.py
basepi: I think you want the errors and warnings in splunk -- even if you ignore them most of the time I think they are going go be useful to have when debugging issues for your customers. Any that explicitly aren't useful, like the insecure warnings that we had before we enabled SSL verify should be cleaned up as we see them.

If you decide that you don't want that data, we would need to make the `logging.SPLUNK` configurable and then we could set it to a value higher than ERROR so that logging at that level doesn't include anything else. Actually, that's a useful feature anyway, so I think I'll add it today anyway.

PR #318, opened 2018-03-26
labelling of tests in profiles

abhikhattri:
This is related to labelling/categorising of test cases in profiles. Field called 'label' can be added to profile tests which can be used to filter out or run only specific test cases. While running audit, user can specify the labels and only those tests would be run which have those labels defined.

Since the structure of tests varies across nova modules, a generic solution to filter labelled tests is hard to implement and can be error prone. I am currently thinking of making the changes to each of the modules.

For this pull request, changes have been made to 'pkg' and 'stat' modules. If these look good, I will go ahead and make changes to other modules as well. 

**This is just for review, kindly do not merge.**


abhikhattri:
Closing this as #329 was created for complete changes

PR #319, opened 2018-03-27
Dockerfile updates

basepi:
This does a few things:

* Automatic checksums for packages
* --iteration support
* Remove /etc/hubble/hubble from list of config files in FPM (so we can push new scheduled jobr with new versions)

We still need to update the osquery-building code.

### DIFF
--- a/pkg/amazonlinux2016.09/Dockerfile
+++ b/pkg/amazonlinux2016.09/Dockerfile
@@ -139,7 +139,7 @@ CMD [ "pyinstaller --onedir --noconfirm --log-level ${_BINARY_LOG_LEVEL} --addit
        -v ${HUBBLE_VERSION} \
     #todo: get rid of the git dependency with static bin in the future
        -d 'git' \
-       --config-files /etc/hubble --config-files /etc/osquery/osquery.conf \
+       --config-files --config-files /etc/osquery/osquery.conf \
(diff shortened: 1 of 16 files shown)


pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -137,11 +138,14 @@ CMD [ "pyinstaller --onedir --noconfirm --log-level ${_BINARY_LOG_LEVEL} --addit
     && fpm -s dir -t rpm \
        -n hubblestack \
        -v ${HUBBLE_VERSION} \
+       --iteration ${HUBBLE_ITERATION} \
     #todo: get rid of the git dependency with static bin in the future
        -d 'git' \
-       --config-files /etc/hubble --config-files /etc/osquery/osquery.conf \
+       --config-files --config-files /etc/osquery/osquery.conf \
```
yassingh: Why are we not marking /etc/hubble as config-file anymore?

basepi: In DX, we want the ability to deliver new scheduled jobs without having to work with the users manually, to keep maintenance costs low. This is why we added the `/etc/hubble/hubble.d/*.conf support` -- we want to move all user config into those files so that  when the hubble package upgrades, it can upgrade `/etc/hubble/hubble` as well and won't clobber user config or be clobbered by user config.

PR #320, opened 2018-03-28
Adding failure reasons to all nova modules

yassingh:


anuragpaliwal80:
@yassingh Please sync the latest code from develop branch and make your changes on top of that.

hubblestack/files/hubblestack_nova/win_reg.py
```diff
@@ -86,11 +88,13 @@ def audit(data_list, tags, debug=False, **kwargs):
                                 tag_data['value_found'] = current
                                 ret['Success'].append(tag_data)
                             else:
+                                tag_data['failure_reason'] = "Value of registry key '{0}:{1}:{2}' is set to value '{3}'. It should be set to '{4}({5})'.".format(reg_dict['hive'], reg_dict['key'], reg_dict['value'], current, match_output, tag_data['value_type'])
```
basepi: Can you shorten these very long lines? I don't necessarily want to enforce the pep8 80 column rule, but I do like to aim for 90-ish columns with a hard cap at 120 columns.

For example, I would probably format this line like this:

```
                                tag_data['failure_reason'] = "Value of registry key '{0}:{1}:{2}' should be set to " \
                                                             "'{4}({5})'. It is set to some other value for one or " \
                                                             "more SID(s)".format(reg_dict['hive'],
                                                                                  reg_dict['key'],
                                                                                  reg_dict['value'],
                                                                                  current,
                                                                                  match_output,
                                                                                  tag_data['value_type'])
```

yassingh:
I have shortened the lines. 
( And thanks for introducing to such pythonic standards and practices 👍 )

hubblestack/files/hubblestack_nova/grep.py
```diff
@@ -121,41 +121,41 @@ def audit(data_list, tags, debug=False, **kwargs):
                 if grep_ret:
                     found = True
                     failure_reason = "Found the blacklisted string '{0}' in file '{1}'." \
-                                    " The file should not contain any string like '{2}'" \
-                                    .format(grep_ret,
```
yassingh: Do you use some tool or plugin to auto-indent?
If yes, let me know about it.

basepi: Vim does it for me on the third line. But not the second. I just am OCD about that kind of stuff.

mew1033:
Awesome! Thanks!
Does this make it all the way into Splunk? Looking at the code, it looks like it does, I just want to verify.
@yassingh

PR #326, opened 2018-04-03
Add ability to fetch contents of changed files

basepi:
Includes a maximum file size (defaults to 20KB). Also adds a max file
size for checksums (defaults to 100MB). Adds the size from
os.path.getsize if `stat` is set.


basepi:
Tested and working great.

basepi:
@madchills We'll want to add similar support for file sizes and file contents to the windows pulsar as well.

mew1033:
👍

basepi:
Done.

Issue #327, opened 2018-04-03
Pulsar not recursively adding directory watches at startup


basepi:
It appears that pulsar is not recursively adding directory watches at startup.

For example, with this config:

```
/root:
  exclude:
    - \/root\/sudoers\.d\/.*\.swp$:
        regex: True
  recurse: True
  auto_add: True
checksum: sha256
stats: True
batch: True
```

If I add the directory `/root/sudoers.d/` *after* starting hubble, it will see its addition, add a new watch, and then any files inside of that new directory correctly trigger events.

However, if that directory is present when hubble is started, no files created or modified in that directory will trigger.

basepi:
@jettero This doesn't appear to be fixed. Can you hit me up on slack if you're unclear on my reproduction steps?

jettero:
Just an FYI, I've been working on this like crazy. It's a total mystery still. I have that overload where the new `PulsarWatchManager.add_watch()` hands the `add_watch()` to the `super()` with `rec=True` like it's supposed to, but it's not actually watching it, doesn't claim to be watching it -- the return results from super add_watch don't contain the subdirs like it should -- and indeed doesn't trigger events for the unwatched subdirs.

I'll figure it out, but I'm not really any closer.

jettero:
ahh, I think I found it. Just have to figure it out quick.

PR #328, opened 2018-04-04
Fim recurse round #2

jettero:
* this fixes some minor issues with config updating and `recurse: True`
* and adds testing of the above (which found the issues)


basepi:
Awesome! Thanks for the quick fix! Fixes #327

PR #330, opened 2018-04-05
Change index extracted fields logic

mew1033:
This removed the field from the event dict if it's going to go in the fields dict.

### DIFF
--- a/hubblestack/extmods/grains/cloud_details.py
+++ b/hubblestack/extmods/grains/cloud_details.py
@@ -31,7 +31,6 @@ def _get_aws_details():
     # Gather amazon information if present
     ret = {}
     aws = {}
-    aws['aws_ami_id'] = None
     aws['cloud_instance_id'] = None
     aws['cloud_account_id'] = None
     aws['cloud_type'] = 'aws'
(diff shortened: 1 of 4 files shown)


hubblestack/extmods/grains/cloud_details.py
```diff
@@ -44,8 +43,6 @@ def _get_aws_details():
         # Else it will throw an Exception
         aws['cloud_account_id'] = int(aws['cloud_account_id'])
 
-        aws['aws_ami_id'] = requests.get('http://169.254.169.254/latest/meta-data/ami-id',
-                                         timeout=1).text
```
basepi: Do we want to remove this? I'm pretty sure DME finds this useful since they have a standard parent AMI.

mew1033: I was talking it over with Scott, and with the instance id there, it's trivial to tie it back to the ami id with mavlink. There's not really a reason to have it in every single hubble event.

basepi: @sumanmehta is this change acceptable, given that you can tie the instance ID to the ami ID pretty easily later? Reducing the number of fields we add to splunk events has a huge effect over millions of events.

fossam: osquery allows easily to get additional aws instance metadata with a query if needed (https://osquery.io/schema/2.11.2#ec2_instance_metadata). This does not work for Azure, though. Throwing ami_id data in with _every_ event just adds to unnecessary storage pressure.

sumanmehta: @fossam Do you also plan to remove account-id as it also seems redundant given that instance ids are unique for AWS? As of now, we do not use this info, rather we use config data provided by Mavlink as it gives information about all the instances even where Hubble is not running.  So I am fine if we remove this field.

fossam: @sumanmehta For now, I don't plan to remove account id because we've had times when we had account id from Hubble and nothing from anywhere else. If you have such situation attributing the instance to somebody becomes challenging without account id. 
It would be fine to have it reporting once an hour in a different sourcetype, but as it is I don't think it is.

PR #331, opened 2018-04-06
Introducing bucketing in hubble

MoodyMudit:
With this implementation, a large fleet of virtual machines can be grouped in buckets and executed at a different time so that there is no resource crunch on the underlying hardware.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -265,7 +265,7 @@ def schedule():
                 buckets = int(jobdata['buckets']) if int(jobdata['buckets'])!=0 else 256
                 s = socket.gethostbyname(socket.gethostname())
                 ips = s.split('.')
-                sum = (int(ips[0])*256*256*256)+(int(ips[1])*256*256)+(int(ips[2]*256))+int(ips[3])
+                sum = (int(ips[0])*256*256*256)+(int(ips[1])*256*256)+(int(ips[2])*256)+int(ips[3])
                 bucket = sum%buckets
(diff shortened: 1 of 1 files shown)


hubblestack/daemon.py
```diff
@@ -162,6 +164,27 @@ def main():
             log.exception('Error executing schedule')
         time.sleep(__opts__.get('scheduler_sleep_frequency', 0.5))
 
+def getLastRunByBuckets(buckets, seconds):
```
basepi: Could you make this function lowercase instead of camelCase? Python only uses camelCase on class names. You can add underscores if you think it improves readability but they are not necessary.

hubblestack/daemon.py
```diff
@@ -162,6 +164,27 @@ def main():
...
+    buckets = int(buckets) if int(buckets)!=0 else 256
+    host_ip = socket.gethostbyname(socket.gethostname())
+    ips = host_ip.split('.')
+    sum = (int(ips[0])*256*256*256)+(int(ips[1])*256*256)+(int(ips[2])*256)+int(ips[3])
+    bucket = sum%buckets
+    current_time = time.time()
+    base_time = seconds*(math.floor(current_time/seconds))
+    splay = seconds/buckets
+    seconds_between_buckets = splay
+    randomInt = random.randint(0,splay-1) if splay !=0 else 0
```
basepi: Another one that should be lowercase.

MoodyMudit:
corrected the names as requested

PR #332, opened 2018-04-08
fix pruning to address #327

### LINKED ISSUE #327: Pulsar not recursively adding directory watches at startup

jettero:
Attempt to fix #327 (and fix a couple other incidentals would probably show to be broken later)

### DIFF
--- a/hubblestack/extmods/modules/pulsar.py
+++ b/hubblestack/extmods/modules/pulsar.py
@@ -461,7 +461,7 @@ class PulsarWatchManager(pyinotify.WatchManager):
     def prune(self):
         def _wd(l):
             for item in l:
-                yield self.watch_db[x]
+                yield self.watch_db[item]
         to_stop = self._prune_paths_to_stop_watching()
         to_rm = self._listify_anything( _wd(to_stop) )
         self.rm_watch(to_rm)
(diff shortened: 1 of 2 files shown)


hubblestack/extmods/modules/pulsar.py
```diff
@@ -42,6 +43,9 @@
     DEFAULT_MASK = None
 
 __virtualname__ = 'pulsar'
+CONFIG = None
+CONFIG_STALENESS = 0
```
basepi: Any reason you added these back in? I removed them because they appeared unused. Possible you just need a rebase.

jettero: hrm, I think that's a merge artifact... checking.

PR #333, opened 2018-04-09
implemented PCI-8.2.4, max days for passwd expiry

abhikhattri:
Added function which does the following:
1. Checks if value of PASS_MAX_DAYS is less than or equal to 90 (passed via profile args)
2. For each user (other than except_for_users), checks if maximum days for password expiry is less than 90

Sample profile data:

```
ensure_max_password_expiration:
    data:
      CentOS Linux-7:
       tag: CIS-5.4.1.1
       function: ensure_max_password_expiration
       args: 
         - 90
         - usr1, usr2
       description: Ensure max password expiration days is set correctly for each user
```

### DIFF
--- a/hubblestack/files/hubblestack_nova/misc.py
+++ b/hubblestack/files/hubblestack_nova/misc.py
@@ -973,6 +973,48 @@ def check_if_any_pkg_installed(args):
             break
     return result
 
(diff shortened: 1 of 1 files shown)


abhikhattri:
Closed this. Will open other.

PR #334, opened 2018-04-09
misc function for passwd expiry days check

abhikhattri:
Added function which does the following:

Checks if value of PASS_MAX_DAYS is less than or equal to 90 (passed via profile args)
For each user (other than except_for_users), checks if maximum days for password expiry is less than 90
Sample profile data:

```
ensure_max_password_expiration:
    data:
      CentOS Linux-7:
       tag: CIS-5.4.1.1
       function: ensure_max_password_expiration
       args: 
         - 90
         - usr1, usr2
       description: Ensure max password expiration days is set correctly for each user
```


hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -973,6 +973,45 @@ def check_if_any_pkg_installed(args):
...
+            result.append('User ' + user + ' has max password expiry days ' + user_passwd_expiry + ', which is more than ' + str(allow_max_days))
+            
+    return True if result == [] else str(result)
+            
+def _is_int(input):
+  try:
+    num = int(input)
+  except ValueError:
+    return False
+  return True
```
basepi: Can you please use 4-space indents to match the rest of the code?

PR #337, opened 2018-04-10
Another fix dump

basepi:
* Fix systemuuid grain for missing osquery
* Don't assume we have system_uuid in grains for splunk returner
* Log splunk options at debug level to better protect token
* Move splunk_index_extracted_fields to the top level of the config
* Fix osquery timing for better formatting for splunk
* Fix negative times in osquery timing

### DIFF
--- a/hubblestack/extmods/returners/splunk_nebula_return.py
+++ b/hubblestack/extmods/returners/splunk_nebula_return.py
@@ -64,7 +64,7 @@ def returner(ret):
         opts_list = _get_options()
 
         for opts in opts_list:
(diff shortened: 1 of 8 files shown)


hubblestack/extmods/returners/splunk_nebula_return.py
```diff
@@ -75,9 +75,9 @@ def returner(ret):
 
             # Set up the fields to be extracted at index time. The field values must be strings.
             # Note that these fields will also still be available in the event data
```
mew1033: These notes are no longer correct. Can you remove this line?

PR #340, opened 2018-04-11
Implementing Cron in hubble

MoodyMudit:
Implementing cron in hubble for users who want to execute hubble process at some definite time.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -171,6 +171,7 @@ def getlastrunbycron(cron_exp):
     this function will use the cron_exp provided in the hubble config to
     execute the hubble processes as per the scheduled cron time
     '''
+    global seconds
     base = datetime(2018, 1, 1, 0, 0)
     iter = croniter(cron_exp, base)
     next_datetime  = iter.get_next(datetime)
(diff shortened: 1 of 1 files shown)


basepi:
My only concern here is the addition of a new dependency. @MoodyMudit your team isn't using this, correct? You're just using the bucketing?

I'm currently leaning towards holding off on this. We can always come back to it later if someone requests it. But in general, I don't think hubble schedule needs this kind of precision.

Thoughts?

sumanmehta:
Actually, this feature was proposed as one of the solutions for ITC workflows but later bucketing was decided. I think we can still add it even though we are not using it at this moment.
Regarding additional dependency, does it increase memory footprint of Hubble?

basepi:
No, memory is not the problem. More deps just complicates the build/install process, and if nobody is planning to use the feature then I’m leaning towards waiting.

MoodyMudit:
If there isn't any problem with the addition of a dep, we'd like to go ahead with merging the cron implementation. Cron gives a flexibility of scheduling a process at a specific time which can be leveraged at multiple places.

I was able to make it work by adding the croniter to the pyinstaller-requirements.txt file. Do I need to add it anywhere else?
Thanks.

MoodyMudit:
Hi Basepi,
I have added croniter in all pkg files including dev folder and also in setup.py and hook-hubblestack file.

PR #341, opened 2018-04-11
Adding option to fetch custom grains and pillar

anuragpaliwal80:
Needs to add the following sort of section in config file:

```

##############################################
## For fetching custom garins and pillar info
##############################################

custom_grains_pillar:
      grains:
        - selinux: selinux:enabled
        - release: osrelease
      pillar:
        - ntpserver: network_services:ntpserver
```

To add any of these fields with each log, user can use custom_fields option, for ex to add release and ntpserver with each log:

```
hubblestack:
  returner:
    splunk:
      - token: <token>
        indexer: <indexer>
        index: hubble
        sourcetype_nova: hubble_audit
        sourcetype_nebula: hubble_osquery
        sourcetype_pulsar: hubble_fim
        sourcetype_log: hubble_log
        custom_fields:
          - release
          - ntpserver
  splunklogging: True
```

### DIFF
--- a/hubblestack/extmods/grains/custom_grains_pillar.py
+++ b/hubblestack/extmods/grains/custom_grains_pillar.py
@@ -1,6 +1,9 @@
 '''
 HubbleStack Custom Grains and Pillar
 
+Allows for fetching custom grain and pillar data from a local salt-minion via
+salt-call
+
 :maintainer: HubbleStack
 :platform: All
 :requires: SaltStack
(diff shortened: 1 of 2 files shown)


basepi:
I fixed one small bug and did a little bit of linting work. I'll get @jettero to review it as well and we'll get this in.

hubblestack/extmods/grains/custom_grains_pillar.py
```diff
@@ -0,0 +1,69 @@
...
+    for grain in custom_grains:
+        for key in grain:
+            if _valid_command(grain[key]):
+                value = __salt__['cmd.run']('salt-call grains.get {0}'.format(grain[key])).split('\n')[1].strip()
+                grains[key] = value
+    custom_pillar = __salt__['config.get']('custom_grains_pillar:pillar', [])
+    for pillar in custom_pillar:
+        for key in pillar:
+            if _valid_command(pillar[key]):
+                value = __salt__['cmd.run']('salt-call pillar.get {0}'.format(pillar[key])).split('\n')[1].strip()
```
jettero: this feels like a mistake could cause major problems... What about something like:

```
__salt__['cmd.run'](['salt-call', 'pillar.get', pillar[key]).split( ... )
```
?

hubblestack/extmods/grains/custom_grains_pillar.py
```diff
@@ -0,0 +1,69 @@
...
+    is only necessary for custom grains and pillar data.
+    '''
+    log.debug('Fetching custom grains and pillar details')
+    grains = {}
+    salt.modules.config.__opts__ = __opts__
+    custom_grains = __salt__['config.get']('custom_grains_pillar:grains', [])
+    for grain in custom_grains:
+        for key in grain:
+            if _valid_command(grain[key]):
+                value = __salt__['cmd.run']('salt-call grains.get {0}'.format(grain[key])).split('\n')[1].strip()
```
jettero: this feels like a mistake could cause major problems... What about something like:

```
__salt__['cmd.run'](['salt-call', 'grains.get', grain[key]).split( ... )
```
?

Also, why cmd.run? can't we call grains.get directly? This is because it's coming from salt, not hubble, right?

basepi: Salt, by default, uses shlex.split to split the command into an array to avoid command injection vulnerabilities. So we're safe unless you pass in `python_shell=True`.

And yes, this is for getting info from the local salt minion. Mainly pillar data, but also potentially custom grains.

hubblestack/extmods/grains/custom_grains_pillar.py
basepi: Salt, by default, uses shlex.split to split the command into an array to avoid command injection vulnerabilities. So we're safe unless you pass in `python_shell=True`.

PR #342, opened 2018-04-11
pylintrc and travis-ci yaml

jettero:
This requires further discussion.
I turned this on for my personal CI. You can see the last build output here: https://travis-ci.org/jettero/hubble

I've had some luck applying autopep8 to the source files, but that really doesn't resolve everything (or even most things).

The pylintrc is just the Salt pylintrc with very little (if any) modification. I think there's quite a few things we can remove and we'll probably want to tune the policies that it's checking.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+
+language: python
+
+cache: pip
+
+env:
+  - 'PYLINT_ENABLE=1'
+
+python:
+  - "2.7"
+
+install:
+  - pip install -r test-requirements.txt
+
+script:
+  - bash tests/travis-runner.sh
--- a/pylintrc
+++ b/pylintrc
@@ -1,0 +1,440 @@
+[MASTER]
+# Specify a configuration file.
+#rcfile=
+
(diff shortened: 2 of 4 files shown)


basepi:
Biggest two offenders are string formatting in logging and duplicate code.

We should fix the first one -- logging interpolation is lazy, so better in almost every case.

I think we should disable the duplicate code one for now. We know where we need to DRY.

I'll look at this more later.

jettero:
Should I make any attempt to slowly introduce the linting? 

IE, comment most of it out, make those the only actual checks for the moment, then we just turn on more parts as desired? or is it better to do it all at once?

jettero:
I'm reasonably happy with this setup. I have to doublecheck a few things, but I intend for it to be just the pylint "test" and a light application of the pylint. We should tune it to be more hardcore pep8 later, probably, but this is a good start imo.

jettero:
Yeah, let's just review this as is. I think we'll make a neverending series of passes at it as time goes on regardless. I can move on to another task and if I feel like I need to change something, I'll submit a new PR.

PR #352, opened 2018-04-17
[Don't Merge] Make http_event_collector_SSL_verify configurable

anuragpaliwal80:

PR #353, opened 2018-04-17
Making http_event_collector_ssl_verify configurable and unicode check for custom fields

anuragpaliwal80:
1) Making http_event_collector_ssl_verify configurable
2) Adding unicode check for custom fields.

### DIFF
--- a/hubblestack/extmods/returners/splunk_nebula_return.py
+++ b/hubblestack/extmods/returners/splunk_nebula_return.py
@@ -147,6 +147,8 @@ def returner(ret):
                                 custom_field_value = __salt__['config.get'](custom_field, '')
                                 if isinstance(custom_field_value, str):
                                     event.update({custom_field_name: custom_field_value})
+                                elif isinstance(custom_field_value, unicode):
+                                    event.update({custom_field_name: custom_field_value})
                                 elif isinstance(custom_field_value, list):
                                     custom_field_value = ','.join(custom_field_value)
                                     event.update({custom_field_name: custom_field_value})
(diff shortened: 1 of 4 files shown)


anuragpaliwal80:
Similar PR: https://github.com/hubblestack/hubble/pull/352
@basepi Let me know which approach you like (PR: 352 or this PR) and I will make similar changes in other files and close one of the PR.

basepi:
Let's do it this way, I like avoiding `global`.

Note that we need one more change. We want to disable the InsecureRequestWarning, because it happens at `warning` level which is logged to splunk by default....which will cause another InsecureRequestWarning which is logged to splunk...it's a vicious cycle.

These are the lines we need, and we should just always add these lines, they should not be part of the configuration.

```
import urllib3
urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```

One more note: please don't forget about splunklogging.py, it needs these changes as well.

hubblestack/extmods/returners/splunk_nebula_return.py
```diff
@@ -142,6 +147,8 @@ def returner(ret):
                                 custom_field_value = __salt__['config.get'](custom_field, '')
                                 if isinstance(custom_field_value, str):
                                     event.update({custom_field_name: custom_field_value})
+                                elif isinstance(custom_field_value, unicode):
+                                    event.update({custom_field_name: custom_field_value})
```
basepi: Instead of adding a new if statement, can we just change the existing one like this?

```
if isinstance(custom_field_value, (str, unicode)):
```

hubblestack/extmods/returners/splunk_nebula_return.py
```diff
@@ -49,11 +49,12 @@
 import json
 import time
 from datetime import datetime
+import urllib3
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
mew1033: Let's not disable these warnings globally. If the connection is insecure, we want to know about it. I like the ability to disable verification, but the warning should stay. Or there should be a configurable parameter to disable warnings, and the parameter should default to showing the warnings.

PR #355, opened 2018-04-17
Add s3fs with proper overriding

basepi:


hubblestack/extmods/fileserver/s3fs.py
```diff
@@ -0,0 +1,742 @@
...
+    key, keyid, service_url, verify_ssl, kms_keyid, location, path_style, https_enable = _get_s3_key()
+
+    # check the local cache...
+    if os.path.isfile(cached_file_path):
+        file_meta = _find_file_meta(metadata, bucket_name, saltenv, path)
+        if file_meta:
+            file_etag = file_meta['ETag']
+
+            if file_etag.find('-') == -1:
+                file_md5 = file_etag
```
fossam: Wouldn't this fail if the Etag is not md5?
It may not be as described here: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html and it is determined by how the file has been uploaded. 
The largest file we have is already almost 5MB so it's potentially getting close. Worse, if this fails every time, there would never be a cache hit, and we'd be downloading the largest file _every_ time.

basepi: Let me look. This isn't code that I touched, you can see my actual diff here: https://github.com/saltstack/salt/pull/47140

Even if we downloaded it every time, "every time" is only hourly, so not a huge deal, but I'll still see if I can fix it.

fossam: Depends on how many clients per hour and the bandwidth. Adds up.

basepi: I added a documentation note describing how to avoid multipart uploads.

PR #356, opened 2018-04-18
Another fix dump

### LINKED ISSUE #280: False Flags on User Home Directories
In the misc.py module, the functions "check_all_users_home_directory", "check_users_home_directory_permissions", and "check_users_own_their_home" have inconsistent script checks with what is available in "CIS_Ubuntu_Linux_16.04_LTS_Benchmark_v1.1.0.pdf", in that they do not skip the users that have a shell of "/bin/false" or "/user/sbin/nologin". This is causing the user "nobody" to be flagged as failing because it has a UID of 65534, but should not apply because it is non-login. The max_system_uid should not be increased to account for this, because it would then skip users that should have home directories and be validated as such.

Here is a snippet of the custom Nova profile I am using to conform with the spec update for Ubuntu 16.04 that calls these functions:

```
misc:
  check_home_directory:
    data:
      Ubuntu-16.04:
        tag: CIS-6.2.7
        function: check_all_users_home_directory
        kwargs:
            max_system_uid: 999
    description: Ensure all users' home directories exist
  home_directory_permission:
    data:
      Ubuntu-16.04:
        tag: CIS-6.2.8
        function: check_users_home_directory_permissions
    description: 'Ensure users home directories permissions are 750 or more restrictive'
  check_owner_home:
    data:
      Ubuntu-16.04:
        tag: CIS-6.2.9
        function: check_users_own_their_home
        kwargs:
            max_system_uid: 999
    description: Ensure users own their home directories

```

### LINKED ISSUE #287: Bug with check_directory_files_permission
In hubblestack_nova/misc.py , the function check_directory_files_permission has a bug in it on the develop branch:

The "path" parameter is not being used. Relatively, the _execute_shell_command call is always calling "/var/log". I think you may have meant to put the path variable in there?

This function should help to fulfill CIS-4.2.4 in the Ubuntu 16.04 v1.1.0 spec: Ensure permissions on all logfiles are configured.

basepi:
* Fixes #287 (Actually use the argument for checking directory permissions)
* Fixes #280 (Better support for ignoring nologin and false shells in user checks)
* Moves the source-install version of the systemd unit file to a more appropriate location
* Default misc.py executions to python_shell=False for safety


fossam:
If nothing else than a reminder for myself: test on all distros these are called from.

hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -295,7 +295,7 @@ def unowned_files_or_dir(reason=''):
     Ensure no unowned files or directories exist
     '''
     raise CommandExecutionError('Module disabled due to performance concerns')
-    result = _execute_shell_command('df --local -P | awk {\'if (NR!=1) print $6\'} | xargs -I \'{}\' find \'{}\' -xdev -nouser')
+    result = _execute_shell_command('df --local -P | awk {\'if (NR!=1) print $6\'} | xargs -I \'{}\' find \'{}\' -xdev -nouser', python_shell=True)
```
jettero: I find this form horrific. I'm not requesting any specific change, I just find it horrific. IMO, it should be several popen() calls chained together using p.communicate(); or native python calls or something.

hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -464,7 +464,7 @@ def check_duplicate_gids(reason=''):
     '''
     Return False if any duplicate group id exist in /etc/group file, else return True
     '''
-    gids = _execute_shell_command("cat /etc/group | cut -f3 -d\":\"").strip()
+    gids = _execute_shell_command("cat /etc/group | cut -f3 -d\":\"", python_shell=True).strip()
```
jettero: useless use of cat... again, not a critique of the actual patches here, but cut takes a file argument, so cat doesn't need to be lifted from disk. Also, python could perform the cut and the file open without invoking the shell at all. Not requesting a  change, just annoyed by the form.

anuragpaliwal80:
Changes looks fine to me but I need to verify on few OS to get comfortable. Will plan to do the same.

hubblestack/files/hubblestack_nova/misc.py
basepi: I think this call is probably pulled directly from the CIS doc, though I haven't checked. I would be fine with cleaning it up but that wasn't the purpose of my changes here. :)

hubblestack/files/hubblestack_nova/misc.py
basepi: Same as above -- could definitely be cleaned up.

PR #357, opened 2018-04-19
Updating create_build_env.ps1 for creating windows build

yassingh:
Faced some problems in creating/automating hubble build for windows. Making changes for the same.

@madchills 
Let me know if there there is some better way to handle these.

### DIFF
--- a/pkg/windows/create_build_env.ps1
+++ b/pkg/windows/create_build_env.ps1
@@ -86,6 +86,8 @@ if ($repo -notlike "https*") {
         $branch = Read-Host "Enter a Branch"
     }
 }
+Write-Error "---------------will be cloning1 : $repo"
+Write-Error "---------------will be cloning2 : $branch"
 git clone $repo
 Push-Location hubble\pkg\windows
 if ($branch -like "\W.+") {
--- a/pkg/windows/create_build_env.ps1
+++ b/pkg/windows/create_build_env.ps1
@@ -58,6 +58,7 @@ if (!($7zip)) {
 if (Test-Path .\salt) {
     Remove-Item -Recurse -Force salt
 }
+Push-Location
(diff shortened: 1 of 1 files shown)


pkg/windows/create_build_env.ps1
```diff
@@ -58,6 +58,7 @@ if (!($7zip)) {
 if (Test-Path .\salt) {
     Remove-Item -Recurse -Force salt
 }
+Push-Location
```
yassingh: this statement was lost in https://github.com/hubblestack/hubble/commit/b3c3ef0d03d4fe6bcf0946a9e938d843ef643150#diff-dfb17bd604fab45bce0fb57b3e965f93
code is not able to find salt directory in line number 66/67 ( Push-Location salt )

pkg/windows/create_build_env.ps1
```diff
@@ -113,6 +114,11 @@ if (!($port_git)) {
 set-location C:\Temp
 Import-Module "$env:ChocolateyInstall\helpers\chocolateyInstaller.psm1" -Force;
 $ChocoTools = Get-ToolsLocation
+
+if ([string]::IsNullOrEmpty($ChocoTools)) {
+    $ChocoTools = [System.Environment]::GetEnvironmentVariable("ChocolateyToolsLocation","User")
```
yassingh: In my case, $env:ChocolateyInstall seems to be un-defined. So I am getting error in importing module in line 114/115. Consequently, $ChocoTools is resolving to an empty string. 
 However, "ChocolateyToolsLocation" is defined under user variables. So we can use it as a fallback option.

yassingh:
Made some other changes by mistake. So closed this PR. WIll open another PR .

PR #361, opened 2018-04-19
Add option to ignore running hubble proc

basepi:
This option also disables the pidfile for the new proc.

This is obviously false by default, but for testing I want the ability to run more than one process.


jettero:
I thought I already did something like that... Must be my imagination.

PR #362, opened 2018-04-20
Pulsar fixes

basepi:
* Actually send `contents` to splunk
* Add periodic refreshes to pulsar.top (both linux and windows)
* Fix an import error in pulsar.py when pyinotify is missing (due to a class inheritance declaration)

### DIFF
--- a/hubblestack/extmods/modules/pulsar.py
+++ b/hubblestack/extmods/modules/pulsar.py
@@ -981,6 +981,8 @@ def get_top_data(topfile):
     Cache the topfile and process the list of configs this host should use.
     '''
     # Get topdata from filesystem if we don't have them already
+    global TOP
+    global TOP_STALENESS
     if TOP and TOP_STALENESS < 60:
         TOP_STALENESS += 1
         topdata = TOP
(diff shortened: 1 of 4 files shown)


hubblestack/extmods/modules/pulsar.py
```diff
@@ -958,20 +977,30 @@ def top(topfile='salt://hubblestack_pulsar/top.pulsar',
 
 
 def get_top_data(topfile):
+    '''
+    Cache the topfile and process the list of configs this host should use.
+    '''
+    # Get topdata from filesystem if we don't have them already
+    global TOP
+    global TOP_STALENESS
+    if TOP and TOP_STALENESS < 60:
```
fossam: You decide - if things break, go even less than once a minute? (like every 5 minutes) Not a huge deal. Just making sure it won't fill the logs too much.

basepi: As we discussed in slack, this is a very lightweight operation, and the only reason I even track staleness is to keep it from logging too often.

PR #363, opened 2018-04-23
Work on Dockerfiles

fossam:
The goal is to have working Dockerfiles for the release. 
Additionally, remove git dependency from from the packages.


fossam:
Need to test Amazon Linux packages.

fossam:
All working, even deb7.

PR #364, opened 2018-04-23
Dockerfile change

yassingh:
We are trying to use Dockerfiles for automation of hubble build creation. It would be better for us if we could specify the git repo/branch from which code needs to be pulled while build creation.

To achieve this, we can parametrize the Dockerfile so that git repo and branch can be passed as arguements in docker image creating step.

Let us know if these changes are ok. If ok, I will make similar changes to Dockerfiles of other OS as well.

One more alternative for this:
If we move the last few lines of the dockerfile ( related to fetching code from git repo ) to CMD section, then we can specify the git repo and git branch at docker "run" step instead of docker "build" step.  We can easily provide environment variables from outside while executing the docker run command.

### DIFF
--- a/pkg/amazonlinux2016.09/Dockerfile
+++ b/pkg/amazonlinux2016.09/Dockerfile
@@ -91,10 +91,12 @@ RUN yum install -y ruby ruby-devel rpmbuild rpm-build rubygems gcc make \
 #pyinstaller start
 #commands specified for ENTRYPOINT and CMD are executed when the container is run, not when the image is built
 #use the following variables to choose the version of hubble
-ENV HUBBLE_CHECKOUT=v2.3.0
+ARG GIT_PATH=https://github.com/hubblestack/hubble.git
+ARG GIT_BRANCH=v2.3.0
+ENV HUBBLE_CHECKOUT=${GIT_BRANCH}
 ENV HUBBLE_VERSION=2.3.0
 ENV HUBBLE_ITERATION=1
-ENV HUBBLE_GIT_URL=https://github.com/hubblestack/hubble.git
+ENV HUBBLE_GIT_URL=${GIT_PATH}
 ENV HUBBLE_SRC_PATH=/hubble_src
 ENV _HOOK_DIR="./pkg/"
 ENV _BINARY_LOG_LEVEL="INFO"
@@ -141,7 +143,7 @@ CMD [ "pyinstaller --onedir --noconfirm --log-level ${_BINARY_LOG_LEVEL} --addit
        --iteration ${HUBBLE_ITERATION} \
     #todo: get rid of the git dependency with static bin in the future
        -d 'git' \
-       --config-files --config-files /etc/osquery/osquery.conf \
+       --config-files /etc/osquery/osquery.conf \
        --after-install /hubble_build/conf/afterinstall.sh \
        --after-upgrade /hubble_build/conf/afterupgrade.sh \
        etc/hubble etc/osquery etc/init.d opt usr \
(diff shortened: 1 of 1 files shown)


pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -141,7 +144,7 @@ CMD [ "pyinstaller --onedir --noconfirm --log-level ${_BINARY_LOG_LEVEL} --addit
        --iteration ${HUBBLE_ITERATION} \
     #todo: get rid of the git dependency with static bin in the future
        -d 'git' \
-       --config-files --config-files /etc/osquery/osquery.conf \
+       --config-files /etc/osquery/osquery.conf \
```
yassingh: typo

pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -91,10 +91,12 @@ RUN yum install -y ruby ruby-devel rpmbuild rpm-build rubygems gcc make \
 #pyinstaller start
 #commands specified for ENTRYPOINT and CMD are executed when the container is run, not when the image is built
 #use the following variables to choose the version of hubble
-ENV HUBBLE_CHECKOUT=v2.3.0
+ARG GIT_PATH=https://github.com/hubblestack/hubble.git
```
yassingh: parametrizing GIT_PATH and GIT_BRANCH

basepi:
This is probably a good change and may obviate the need for the dev/ dockerfiles (except we'd have to do a little more work integrating `git describe` for versions not on a tag).

How does passing these parameters on the command line look? I'm new do docker build args, so I'm just curious how it works.

yassingh:
build args can be passed while running the docker build command like this : 
```
docker build --build-arg GIT_PATH=git_repo_url --build-arg GIT_BRANCH=git_repo_branch  -t <image_name> . 
```
Since, we have provided a default value to these arguments in Dockerfile, we can omit these  build args in the command. In other words, the existing command to build docker image should also work fine.


Moreover, if you want to substitute a custom "hubble version" and/or "iteration number"  to the hubble build, you can do so by injecting environment variable(s) in docker run command. Something like this:
```
export HUBBLE_ITERATION=12
export HUBBLE_VERSION=2.2.100
docker run -e HUBBLE_ITERATION -e HUBBLE_VERSION --rm -v `pwd`:/data $IMAGE_NAME
```
And again, you can omit these environment variables in the command as you have already provided a default value for these in the docker file.

Note that we can inject environment variable only in "docker run" step and not in "docker build" step.

--------------------------------------------------------------------------------

As I stated earlier, we can move the line ```RUN git clone "$HUBBLE_GIT_URL" "$HUBBLE_SRC_PATH``` to CMD part. By doing so, docker image created in "docker build" step will not contain hubble code, and the hubble code can be injected solely while running the "docker run" command ( via environment variables or using default value).
The benefit with this approach is that you need not create a new docker image whenever you make any change to your hubble code. All the dependencies, python, osquery ,etc are simply frozen in the docker image ( which you need to build just once ) . Afterwards, you can use hubble code from different repos with the same docker image.

fossam:
@yassingh Having pass-in arguments for build and run phase is a great idea! I wish I would have remembered this docker option earlier as it would have saved me from manually editing files to switch between git branches and repos for testing.
I have started to use the arguments in the other PR where I am finishing up the Dockerfiles for the upcoming release. In this release we are going to include the _git_ binary with Hubble to avoid having the install-time _git_ package dependency. Due to that, I converted the _HUBBLE_CHECKOUT_ and _HUBBLE_GIT_URL_ variables to ARGs to avoid possible confusion with the actual _git_ binary and its associated variables.

The primary goal of these Dockerfiles last year was to provide a way to have very well defined and repeatable packaging process for the releases. So all software and their versions that are included in the package have to be defined, and all the software included in the package are in the docker image (including Hubble) and its only output is the package.
Of course we can (and should) use much more flexible system for testing. The build-time and run-time arguments can allow us only build and test the parts we need. For example, for tests that don't need osquery, that part of the build can be disabled.

basepi:
Yeah, @fossam was in the process of major changes to the dockerfiles, so I had him implement your great idea. Thank you again, this will really clean up the dev dockerfiles as well.

PR #366, opened 2018-04-23
Add grain to read splunk config from hubble.d/*.conf

basepi:
Also set up the splunk returners to use this config over __opts__ if it
is present. This allows for splunk config to be updated without a hubble
restart. Note that these changes should have no effect on
splunklogging.py because config is only processed on startup, but I made
the changes anyway for consistency (for when we DRY later)


basepi:
This is tested and working.

PR #368, opened 2018-04-24
Bug fix and improvement in azurefs fallback mechanism

yassingh:
see 
https://github.com/hubblestack/hubble/pull/314 
for more context.

There is a problem with the implementation of fallback mechanism for azure containers. It's not working properly with stored access policy. So making some changes to improve it.

### DIFF
--- a/hubblestack/extmods/fileserver/azurefs.py
+++ b/hubblestack/extmods/fileserver/azurefs.py
@@ -264,20 +264,18 @@ def update():
                         continue
                     if '<Code>AuthenticationFailed</Code>' in str(exc) or '<Code>AuthorizationPermissionMismatch</Code>' in str(exc):
                         log.debug('Could not update the file "{0}" from azure blob.'.format(fname))
-                        container_cache_folder = _get_container_path(container) 
                         log.debug('Trying to delete the corrupt file "{0}"'.format(fname))
                         try:
                             if os.path.exists(fname):
                                 os.remove(fname)
-                                os.remove(lk_fn)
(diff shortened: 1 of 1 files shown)


hubblestack/extmods/fileserver/azurefs.py
```diff
@@ -201,9 +201,19 @@ def update():
         except Exception as exc:
             log.exception('Error occurred fetching blob list for azurefs')
 
-            if not __opts__['delete_inaccessible_azure_containers'] or not "<class 'azure.common.AzureHttpError'>" in str(type(exc)) :
+            if not __opts__['delete_inaccessible_azure_containers'] \
+               or ( not "<class 'azure.common.AzureHttpError'>" in str(type(exc)) \
+                    and \
+                    not "<class 'azure.common.AzureMissingResourceHttpError'>" in str(type(exc))
```
yassingh: If we delete a container, it falls under 'azure.common.AzureMissingResourceHttpError' error.

hubblestack/extmods/fileserver/azurefs.py
```diff
@@ -201,9 +201,19 @@ def update():
...
+                    ):
                 continue
-            if '<AuthenticationErrorDetail>Signature did not match.' in str(exc):
+
```
yassingh: '<AuthenticationErrorDetail>Signature did not match.' is not a good string. THere are different error strings for different scenarios. Specially when we use stored access policy with azurefs, there are many other error strings. These newly added 3 error codes should cover them all.

PR #402, opened 2018-06-04
Adding --buildinfo arguement to hubble

yassingh:
With these changes, we can add build info to a hubble build. We simply need to place file named "hubble_buildinfo " in the directory which is to be mounted as "/data"

The file should contain ```__buildinfo__``` object in standard python format. For example:
```
[root@ip-10-36-1-20 tmp]# cat hubble_buildinfo 
__buildinfo__ = { 'version' : '1.1.26',
                  'creator' : 'yash',
                  'date' : '06.04.2018',
                  'branch' : 'develop',
                 'flavor' : 'dme' }
```

If this file is found in /data , it's contents are appended to ```hubblestack/__init__.py``` file. The resultant file looks something like this:
```
[root@ip-10-36-1-20 centos7]# cat /opt/hubble/hubble-libs/hubblestack/__init__.py 
__version__ = "2.3.4-2-8-g6e5c919"

__buildinfo__ = { 'version' : '1.1.26',
                  'creator' : 'yash',
                  'date' : '06.04.2018',
                  'branch' : 'develop',
                 'flavor' : 'dme' }
[root@ip-10-36-1-20 centos7]# 
```

If everything goes right, this is the output of "hubble --buildinfo" command:
```
[root@ip-10-36-1-20 centos7]# hubble --buildinfo
{'date': '06.04.2018', 'flavor': 'dme', 'version': '1.1.26', 'branch': 'develop', 'creator': 'yash'}
```

In case you don't provide "hubble_buildinfo" file in /data, you get this:
```
[root@ip-10-36-1-20 centos7]# hubble --buildinfo
NOT SET
```

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -637,6 +637,9 @@ def parse_args():
     parser.add_argument('--version',
                         action='store_true',
                         help='Show version information')
+    parser.add_argument('--buildinfo',
+                        action='store_true',
+                        help='Show build information')
     parser.add_argument('function',
                         nargs='?',
                         default=None,
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -58,6 +58,15 @@ def run():
         clean_up_process(None, None)
         sys.exit(0)
 
+    if __opts__['buildinfo']:
+        try:
+            from hubblestack import __buildinfo__
+        except ImportError:
+            __buildinfo__ = 'NOT SET'
+        print(__buildinfo__)
+        clean_up_process(None, None)
+        sys.exit(0)
+
     if __opts__['daemonize']:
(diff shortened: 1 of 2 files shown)


yassingh:
This PR is not complete yet. Let me know if there are some more changes to be made. If you are okay with these changes, I will go on to make similar changes in Dockerfile of every OS ( in "pkg" directory as well as "pkg/dev" directory" )



>> Small problem:

As you can see, the ```__version__``` variable is still present in ```__init__.py``` .  Ideally,  ```__version__```  variable should reside in ```__buildinfo__``` object now. We can modify the code so as to completely overwrite the ```__init__.py``` file ( removing the original ```__version__``` variable )

I have said more about this on slack.

PR #405, opened 2018-06-15
Nebula per-query overrides

basepi:
@sumanmehta This is an important change for you to review, as you must make sure your nebula queries are migrated before the next version.

The reason for this change is we want the ability to override specific queries using nebula.top and the top.nebula file.

### DIFF
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -81,9 +81,9 @@ def queries(query_group,
     MAX_FILE_SIZE = 104857600
     if query_file is None:
         if salt.utils.platform.is_windows():
-            query_file = 'salt://hubblestack_nebula/hubblestack_nebula_win_queries.yaml'
+            query_file = 'salt://hubblestack_nebula_v2/hubblestack_nebula_win_queries.yaml'
         else:
-            query_file = 'salt://hubblestack_nebula/hubblestack_nebula_queries.yaml'
+            query_file = 'salt://hubblestack_nebula_v2/hubblestack_nebula_queries.yaml'
     if not isinstance(query_file, list):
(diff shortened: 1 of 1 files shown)


fossam:
In terms of managing hubblestack_data, what are the implications of this when an old version and new version of hubble need to work at the same time?

basepi:
@fossam The new config points to a new directory, hubblestack_nebula_v2. Shouldn't be any problems.

sumanmehta:
@basepi After this change, nebula profiles need to be present in "hubblestack_nebula_v2" rather than in "hubblestack_nebula" folder. It seems that format of nebula profile will also change. Is my understanding correct?
Could you also elaborate why are we doing this change? what is the usecase of overriding nebula queries.
Thanks

jettero:
@sumanmehta The layout of the profiles changed somewhat in the _v2 tree and the way hubble processes the profile data changed to match. So I think it'll be more a case of versions of hubble that read the new style config will read from v2 and versions that aren't upgraded yet will read the old configs. I don't know the plan for sure, but I suspect each time the config structure out has to change (for whatever reason), we'd stick another _v# on the end there.

We'll have to wait for @basepi's reply regards to use cases, but I can make a guess: normally you wouldn't want to override anything, but if something doesn't work right or you wish to check something in a slightly different way, it might make sense to do so --- say one of the checks takes up too much CPU time on your tiny cloud instances, or it's looking in the wrong place for something.

basepi:
The use case is for per-team overrides. For example, we have one team whose load balancers have issues with the network queries in osquery. We want to be able to modify those queries just for a specific team without having to maintain an entire complete list of queries for that team.

Previously, you could only add new queries via the top.nebula file. Now you can override existing queries. Does that make sense?

PR #409, opened 2018-06-21
Adding exception and timeout in custom grains and pillar module

anuragpaliwal80:


basepi:
Were these running for too long? I'm a little worried that this default might be too aggressive. For a minion with a busy master, it could take longer than 5 seconds to get everything from the master.

I would recommend a 30 second default, since it's configurable. Also, we don't generate grains very often, so 30 seconds every hour shouldn't be a problem.

anuragpaliwal80:
Then we might see hubble service time out issue when we try to start the hubble service. I think the default service timeout value is 90 sec (on linux) and if we add 30 sec timeout for each grain and pillar call, we might cross the total service start time out limit. And in that case we need to make changes in hubble.service file as well and add some higher service timeout value to 5 min or so.
Suggestions ?

basepi:
What happens if it times out? Does the service system kill the process? Or just report issues with starting the service?

We could potentially solve this issue by daemonizing in load_config() before doing any pillar/grains. That would make the service system happy without risking missing data.

All of that said, I don't think we're actually using this in DX, so it's probably fine as long as you're aware of the risks.

MoodyMudit:
Hi basepi,
The hubble service fails to start since the total time taken to fetch all the grains/pillars is greater than the hubble service timeout. This data is necessary on ITC systems and therefore required here. 

Can you please explain how we can solve the issue by daemonizing in load_config()?

basepi:
The service system is waiting for the command `hubble -d` to return, which happens when we run the daemonize function. Right now that happens after all of the grains and pillar are generated. But we don't need that data to decide to daemonize -- so if we do it after generating the config but before generating pillar and grains, then the service should start properly no matter how long it takes.

basepi:
I think #424 obviates the need for this PR. Can you please test that PR and make sure I am correct?

PR #413, opened 2018-06-28
graylog returner addition

zachsis:
This is a working version compatible with GELF HTTP Graylog inputs. Tested on Graylog 2.4


zachsis:
I used the logstash returner as a template and modified what was needed to be friendly with the GELF Specs outlined here: http://docs.graylog.org/en/2.4/pages/gelf.html#gelf-payload-specification, and refactored a few things so it was more 'grayloggy' than 'logstashy'

PR #415, opened 2018-07-11
attempt to fix various rare process hangups in netlink socket recv()

jettero:
If the hanging netlink socket recv() issue is only applicable to osquery, this is the fix. The condition is quite rare, so the occasional timeout is much preferable and will be logged as a transient error (probably along side other system busy statuses).

I later expanded the PR to cover other hangs in various salt.grains.core functions.


basepi:
Does hubble kill the osquery process when the timeout expires? I want to make sure it's killing it, as opposed to just abandoning it.

basepi:
I'm not too worried about that. Does salt's cmd.run use that timed subprocess class?

mew1033:
@basepi No, I think they implement it with threads on their own.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -163,7 +163,7 @@ def queries(query_group,
 
         cmd = [__grains__['osquerybinpath'], '--read_max', MAX_FILE_SIZE, '--json', query_sql]
         t0 = time.time()
-        res = __salt__['cmd.run_all'](cmd)
+        res = __salt__['cmd.run_all'](cmd, timeout=90)
```
mew1033: Let's change this to 10800 seconds (3 hours). We have some queries that run for a ridiculously long time, but do eventually finish. We want to continue to debug those queries, so we want them to finish and report their query run length into Splunk. We'll track them down and try to fix them.

This fix is just to kill queries that have become indefinitely hung.

fossam: Additionally, is there a way to know when the command ended due to going over the timeout (and log that)? That would give us indication when a query ran but took longer than our allowed time.

jettero: Yeah, when the timeout hits, it logs the stderr and some other things. It wouldn't be super obvious, but if you went looking for missing osquery results, you'd see a message about it and the stderr for the killed proc.

Changing to 10ksec on the nose. I like ksec as a time unit. I can make it precisely 3 hours if you need. Please, ksec!!

jettero:
@basepi salt's modules/cmdmod.py uses various methods to launch the proc, depending on the situation (shell=True, pty needed, etc); but it tracks the subproc and terminates it (if it doesn't exit due to timeout or whatever) and issues all the appropriate waitpid() to avoid zombies.

basepi:
Awesome. Should be a good change.

hubblestack/hangtime.py
```diff
@@ -1,23 +1,74 @@
...
+        ... code-block:: python
+            try:
+                with HangTime(timeout=1, id=1):
+                    do_things_that_may_timeout()
+            except HangTime as ht:
+                if ht.id == 1:
+                    log.error("something bad happened (code-1):", ht) # w/o traceback
+                    log.error(ht) # with traceback
+
+            :param int timeout: timeout in seconds, default 300s
```
basepi: since this is the end of the example it should be back to an 8-space indent.

hubblestack/hangtime.py
```diff
@@ -1,23 +1,74 @@
...
 class HangTime(Exception):
     # NOTE: this will break completely in multithreading
     # it should work just fine in multiprocessing
 
     prev = list()
 
     def __init__(self, msg="hang timeout detected", timeout=300, id=0, repeats=False, decay=1.0):
+        ''' HangTime wraps code via with block.
+
+        ... code-block:: python
```
basepi: I think it's one leading dot. Though I'm tempted to go with markdown instead of RST....though we should probably match salt, I just don't love RST.

hubblestack/hangtime.py
```diff
@@ -0,0 +1,85 @@
...
+    except ReferenceError:
+        return False
+
+class HangTime(Exception):
+    # NOTE: this will break completely in multithreading
+    # it should work just fine in multiprocessing
+
+    prev = list()
+
+    def __init__(self, msg="hang timeout detected", timeout=300, id=0, repeats=False, decay=1.0):
```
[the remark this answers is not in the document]
jettero: oh, oops. Shoulda linted.

hubblestack/daemon.py
```diff
@@ -571,7 +572,21 @@ def __init__(self, message, levelname, asctime, name):
...
+#
+# The grain will simply be missing, but the next refresh_grains will try to
+# pick it up again.  If the hang is transient, the grain will populate
+# normally.
+#
+# repeats=True meaning: restart the signal itimer after firing the timeout
+# exception, which salt catches. In this way, we can catch multiple hangs with
+# a single timer. Each timer restart is a new 600s timeout.
+#
+# id=1337 will appear in the logs to differentiate this from other
```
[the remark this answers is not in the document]
jettero: typo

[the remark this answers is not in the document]
jettero: Actually, there's no reason to restrict that field to integers anyway... it's only used for passing on to except blocks and logging.

PR #417, opened 2018-07-12
Add timeouts to the s3 requests

basepi:
Turns out the upstream saltstack s3 utils were not using timeouts for requests.request. This *might* be the source of the hard to reproduce hangs in hubble. In any case, a timeout is a good thing to define.

However, since we don't want to wait for the new version of salt containing this fix, I had to pull in the s3.py utils file. https://github.com/hubblestack/hubble/commit/8498b1588757bb6ecb0c57438fb9e380df8e99c2 shows the changes that I made from the upstream version.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -499,6 +499,9 @@ def load_config():
     fileserver_dirs = __opts__.get('fileserver_dirs', [])
     fileserver_dirs.append(os.path.join(os.path.dirname(__file__), 'extmods', 'fileserver'))
     __opts__['fileserver_dirs'] = fileserver_dirs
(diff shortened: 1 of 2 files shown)


mew1033:
Any reason why we should make the timeouts user configurable?

basepi:
I can't see a reason. If it's taking more than 5 minutes there are probably worse problems.

basepi:
The logging requested by @fossam is already present: https://github.com/hubblestack/hubble/blob/11c92a9faa4c9b57f68009295583ea2e2c5431a9/hubblestack/daemon.py#L139

This is ready for approval, I think.

PR #429, opened 2018-07-26
Mask nebula passwords

yassingh:
Currently, some passwords are also getting logged by hubble osqueries. 
This PR aims to mask such passwords by using mask.yaml file. See PR in hubblestack_data for more details:
https://github.com/hubblestack/hubblestack_data/pull/119

There are two ways in which passwords are being printed:
1) Some passwords are being passed as command line arguemnts. For example "command_line" column of 'running_procs' contains some passwords in some cases.
2) Some passwords are injected as environment variable, and hence logged under 'environment' column of 'running_procs' for example.

Passwords of type 1 can be handled by "blacklisted_strings" part of mask.yaml.
Passwords of type 2 can be handled by "blacklisted_objects" part of mask.yaml.

DO NOT MERGE. 
( still need to test a few scenarios )
But reviews are welcome.

### DIFF
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -339,7 +339,6 @@ def mask_passwords(unmasked_object):
                                     key == blacklisted_string['column']:
                                         for pattern in blacklisted_string['blacklisted_patterns']:
                                             value = re.sub(pattern + "()",r"\1" + mask_by + r"\3", value)
-                                        #print("string  masking : {0} : {1}".format(result[key],value))
                                         result[key] = value
                             else:
                                 # handle like [json]
@@ -365,7 +364,6 @@ def _recursively_mask_objects(object_to_mask, blacklisted_object, mask_by):
         for key in blacklisted_object['attributes_to_mask']:
             if key in object_to_mask:
                 object_to_mask[key] = mask_by
-                print(" masking : {0}".format(object_to_mask[key]))
 
 
 def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -324,7 +324,7 @@ def mask_passwords(unmasked_object):
     masked_object = copy.deepcopy(unmasked_object)
 
     try:
-        mask = yaml.load(open('mask.yaml'))
+        mask = yaml.load(open('salt://mask.yaml'))
         mask_by = mask.get('mask_by','******')
         for r in masked_object:
             for query_name, query_ret in r.iteritems():
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -227,7 +227,9 @@ def queries(query_group,
                         if value and isinstance(value, basestring) and value.startswith('__JSONIFY__'):
                             result[key] = json.loads(value[len('__JSONIFY__'):])
 
-    return mask_passwords(ret) if mask_passwords else ret
+    if mask_passwords:
+        mask_passwords_inplace(ret)
+    return ret
 
 
 def fields(*args):
@@ -322,7 +324,7 @@ def get_top_data(topfile):
     return ret
 
 
-def mask_passwords(object_to_be_masked):
+def mask_passwords_inplace(object_to_be_masked):
 
     try:
         mask_file  = __salt__['cp.cache_file']('salt://hubblestack_nebula_v2/mask.yaml')
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -361,7 +361,7 @@ def _recursively_mask_objects(object_to_mask, blacklisted_object, mask_by):
         for child in object_to_mask:
             _recursively_mask_objects(child, blacklisted_object, mask_by)
     elif blacklisted_object['attribute_to_check'] in object_to_mask and \
-         object_to_mask[blacklisted_object['attribute_to_check']] in blacklisted_object['balcklisted_patterns']:
+         object_to_mask[blacklisted_object['attribute_to_check']] in blacklisted_object['blacklisted_patterns']:
         for key in blacklisted_object['attributes_to_mask']:
             if key in object_to_mask:
                 object_to_mask[key] = mask_by
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -71,6 +71,10 @@ def queries(query_group,
         Defaults to False. If set to True, more information (such as the query
         which was run) will be included in the result.
 
+    mask_passwords
+        Defaults to False. If set to True, passwords mentioned in the 
+        return object are masked.
+        
     CLI Examples:
 
     .. code_block:: bash
@@ -325,7 +329,11 @@ def get_top_data(topfile):
 
 
 def mask_passwords_inplace(object_to_be_masked):
-
+    '''
+    It masks the passwords present in 'object_to_be_masked'. Uses "mask.yaml" as
+    a reference to find out the list of blacklisted strings or objects.
+    Note that this method alters "object_to_be_masked".
+    '''
     try:
         mask_file  = __salt__['cp.cache_file']('salt://hubblestack_nebula_v2/mask.yaml')
         if not mask_file:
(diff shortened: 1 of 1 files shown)


hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -318,6 +319,53 @@ def get_top_data(topfile):
     return ret
 
 
+def mask_passwords(unmasked_object):
+
+    masked_object = copy.deepcopy(unmasked_object)
+
+    try:
+        mask_file  = __salt__['cp.cache_file']('salt://mask.yaml')
```
basepi: Can we move this into hubblestack_nebula_v2/ please? It should also be a configuration option, overridable in the config file and also via an optional argument to nebula.queries or nebula.top.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -318,6 +319,53 @@ def get_top_data(topfile):
...
+        mask = yaml.safe_load(open(mask_file))
```
basepi: Rather than relying on the try:except:, can we check whether `mask_file` is False here and just return?

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -318,6 +319,53 @@ def get_top_data(topfile):
     return ret
 
 
+def mask_passwords(unmasked_object):
```
basepi: docstring for documentation, please?

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -318,6 +319,53 @@ def get_top_data(topfile):
...
+                                        _recursively_mask_objects(value, blacklisted_object, mask_by)
+
+        return masked_object
+
+    except Exception as e:
+        print("An error occured while masking the passwords: {}".format(e))
+        return unmasked_object
+
+
+def _recursively_mask_objects(object_to_mask, blacklisted_object, mask_by):
```
basepi: Another docstring needed

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -318,6 +319,53 @@ def get_top_data(topfile):
...
+    try:
+        mask_file  = __salt__['cp.cache_file']('salt://mask.yaml')
+        mask = yaml.safe_load(open(mask_file))
+        mask_by = mask.get('mask_by','******')
+        for r in masked_object:
+            for query_name, query_ret in r.iteritems():
+                if 'data' in query_ret:
+                    for result in query_ret['data']:
+                        for key, value in result.iteritems():
+                            if isinstance(value,basestring):
```
basepi: Can you put a space after the comma? We don't have pylint checking live, but it will complain about this.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -318,6 +319,53 @@ def get_top_data(topfile):
...
+                                # handle like string
+                                for blacklisted_string in mask.get("blacklisted_strings",[]):
```
basepi: another space after the comma

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -318,6 +319,53 @@ def get_top_data(topfile):
...
+                                    if blacklisted_string['query_name'] in ('*', query_name) and \
+                                    key == blacklisted_string['column']:
+                                        for pattern in blacklisted_string['blacklisted_patterns']:
+                                            value = re.sub(pattern + "()",r"\1" + mask_by + r"\3", value)
+                                        result[key] = value
+                            else:
+                                # handle like [json]
+                                for blacklisted_object in mask.get('blacklisted_objects',[]):
```
basepi: another space after the comma

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -318,6 +319,53 @@ def get_top_data(topfile):
...
+                                    if blacklisted_object['query_name'] in ('*', query_name) and \
+                                    key == blacklisted_object['column']:
+                                        _recursively_mask_objects(value, blacklisted_object, mask_by)
+
+        return masked_object
+
+    except Exception as e:
+        print("An error occured while masking the passwords: {}".format(e))
```
basepi: Please use log.exception instead of print

fossam:
I am not yet ready to review the code yet, but should we be concerned about the efficiency? It looks like we would enter loops in cases when later on it would turn out unnecessary. Some returns may be huge and blanket iteration might be costly.

yassingh:
Well, I tested the performance of my code by combining all sorts of osqueries together. I didn't see any  significant CPU spike or any significant increase in nebula's running time.
But perhaps that was because osqueries didn't return gargantuan results on my machine.

 I will try to make the code more efficient by leveraging the dict functionality instead of looping.
Will ping once I am done with code changes. Thanks for the reviews.

fossam:
The size of data about a single process can go as big as 28 megabytes. Some processes may have a lot of open files.

yassingh:
@fossam and @MoodyMudit 
I have made some changes to improve the performance:

1. Now I have reduced some for loops. Wherever possible, I have tried to directly fetch the data using dictionary functions. But still if I have missed some optimization(s), please let me know
2. Now I am not using the deepcopy function anymore. Even if the nebula return object is too heavy, I am just working with the object references. So hopefully, the code will not cause memory troubles.
3. Now the code with access only those osquery columns, which are actually meant to be changed. So, if we try to mask "command_line" using mask.yaml file, the code will stay away from the "open files" of the processes. So code will not slow down even if 1000s of files have been opened by the process. 

And there are negligible chances that a password might ever show in information related to file descriptors. Perhaps, we will need to mask "command_line" and "environment" columns only.
For those cases where there can be 100s or 1000s of (json) objects ( like "open-files" , "environment variables" ) one needs to provide the exact attribute name/value which is to be masked. By avoiding the use of regex in such osquery columns, we are further minimizing the chance of accidentally slowing the code.

I understand that besides all the aforementioned measures, there is still some chance that the code can be slow. So I have tested the code's performance with (very) big nebula return objects as well. The code worked fine for me. Didn't face any significant performance hit.

If we meticulously design our mask.yaml and stay away from unnecessarily complicated regexes, the code should work fine. 
[ Plus now the masking functionality is an opt in option. False be default. ]

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -335,32 +335,87 @@ def mask_passwords_inplace(object_to_be_masked):
     Note that this method alters "object_to_be_masked".
     '''
     try:
-        mask_file  = __salt__['cp.cache_file']('salt://hubblestack_nebula_v2/mask.yaml')
-        if not mask_file:
-          return
-        mask = yaml.safe_load(open(mask_file))
+
+        mask = {}
+        topfile = 'salt://hubblestack_nebula_v2/top.mask'
```
MoodyMudit: instead of writing file paths in code, these should be written in some other file that contains all constants. By looking at the path I feel there will, someday, be a **hubblestack_nebula_v3**, due to which changes in the function definition will be made. Hence increasing maintainability.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -317,6 +327,120 @@ def get_top_data(topfile):
...
+def mask_passwords_inplace(object_to_be_masked):
+    '''
+    It masks the passwords present in 'object_to_be_masked'. Uses "mask.yaml" as
+    a reference to find out the list of blacklisted strings or objects.
+    Note that this method alters "object_to_be_masked".
+    '''
+    try:
+
+        mask = {}
+        topfile = 'salt://hubblestack_nebula_v2/top.mask'
```
MoodyMudit: instead of writing file paths in function definitions, these should be written as constants in some file. By looking at the path, I feel there will, someday, be a hubblestack_nebula_v3, due to which you will have to make changes in code. This increases maintainability of the code.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -335,32 +335,87 @@ def mask_passwords_inplace(object_to_be_masked):
...
-          return
-        mask = yaml.safe_load(open(mask_file))
+
+        mask = {}
+        topfile = 'salt://hubblestack_nebula_v2/top.mask'
+        mask_files = get_top_data(topfile)
+        mask_files = ['salt://hubblestack_nebula_v2/' + mask_file.replace('.', '/') + '.yaml'
+                   for mask_file in mask_files]
+        if mask_files is None:
+            mask_files = 'salt://hubblestack_nebula_v2/mask.yaml'
```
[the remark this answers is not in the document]
yassingh: @MoodyMudit 
I have made some chages so that the top.mask can be passed as an argument. 
But still, the code contains some hardcodings. 
These hardcodings are present in the nebula code from a long time. I don't know the reason behind this.
Maybe @basepi  has some idea. 
But in a nutshell, if we ever move to hubblestack_nebula_v3, we must anyways make some other changes in the the nebula file. Hopefully, the masking code will also be changed accordingly at that time.

yassingh:
Changes are complete. The code is good to be merged ( from my side atleast ).
Hence removed the 'do not merge' label.

basepi:
I'd like to get a couple more approvals here. Perhaps from @sumanmehta and @MoodyMudit as well as @fossam

jettero:
I'm still trying to get this to work. I wanted to try it out before I clicked approve. I think it's a neat idea. I just want to see it go... I'm having trouble with the format of the top.mask ... I'm assuming I'll figure it out, but I wonder if there should be a brief example?

yassingh:
@jettero 
I have provided you a sample "mask.yaml" on slack.

jettero:
I was finally able to get this working. It's not exactly straight-forward, particularly because you have to know a  priori what fields to mask and for what. I was spelling `'cmdline'` as `'command_line'` and it _of course_ didn't match anything.

PR #430, opened 2018-07-26
Add optional retries to splunk returners

basepi:
Also brings all the returners in line for requests error handling

Defaults for the retries are 3 times with 15 seconds between. Retries
are on a per-batch basis.

### DIFF
--- a/conf/hubble
+++ b/conf/hubble
@@ -45,6 +45,7 @@ fileserver_backend:
 #    kwargs:
 #      verbose: True
 #    returner: splunk_nova_return
+#    returner_retry: True # only works on splunk returners for now
 #    run_on_start: False
 #  nebula_fifteen_min:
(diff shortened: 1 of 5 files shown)


basepi:

jettero:
This seems alright to me, but what happens if there's a network problem (perhaps intentionally caused by a redteam)? Does hubble get kinda stuck in this section? On fail, we should probably dump to a disk queue and come back around in a minute or something. Being single threaded this could be a problem.

basepi:
@jettero We only retry 3 times, so it's only a minute of lag, which is worth the wait to prevent us from losing daily data.

The on-disk caching is on the backlog as well but is a riskier feature that we didn't want to shove into a quick release. This fix is primarily to overcome very transient network issues, or splunk HEC overloading.

PR #439, opened 2018-08-03
Add before-remove to all packages to stop hubble

basepi:


conf/beforeremove.sh
```diff
@@ -0,0 +1,2 @@
+service hubble stop
+pkill hubble
```
fossam: Should we be more careful here?
Imo, blanket kill statement like this gets me worried when it's likely possible to be more precise.
There is some other software out there that likes to name itself _hubble_, and it's probably matter of time before it's going to bite us.

basepi: That's valid, but then we're back to relying on pidfiles, which historically has not been sufficient. I don't know what the solution is here.

fossam: Since this is before-remove script perhaps an acceptable alternative is to use `-f` and/or `-x` of the _pkill_ options.

jettero: Hrm. How does memcached do it? How about mysqld? what about salt?

IMO, service hubble stop or systemctl stop hubble aught to do the trick. If it doesn't, the SREs should figure out what they broke.

jettero: The more I think about it... in the init.d case, we're talking about pidfiles in fact. upstart tracks pids and forks via memory DB ... systemd creates a whole cgroup for the process. So if we're in systemd we definitely don't ever need the pkill.

basepi: @jettero Then why does systemd have a pidfile entry? Am I doing it wrong? https://github.com/hubblestack/hubble/blob/develop/pkg/hubble.service

I agree that it shouldn't be necessary, but it certainly seems like it doesn't work in every case and I still don't know why.

basepi:
@fossam and @jettero: I decided to be more conservative with the first pass of this, and just call the stop. Especially since `beforeremove` wouldn't even help with the restart issues that people deal with on upgrade, so there's no reason to add that additional risk.

basepi:
Please re-review and we can get this merged.

PR #440, opened 2018-08-07
Disable nebula queries with 'attach' or 'curl'

basepi:
These commands can be used to send sensitive data offsite or even change files on a system. We want hubblestack_data to be as secure as possible from operations of this type so we want to disable queries like this.


jettero:
The string match is a little heavy handed... it's going to match things that aren't curl or attach eventually... what about wget? Otherwise, this seems fine.

basepi:
Yeah if it needs to grow then that's one thing. But I don't think it will, much. We shall see.

basepi:
Thanks for the approvals, but I'm going to close this until I have replacement functionality for @fossam's curl needs (via FDG).

PR #444, opened 2018-08-11
Convert Vulners scanner to use SDK

### LINKED ISSUE #438: Vulners Python API support
Hi fellows!
May I suggest you refactoring Vulners Scanner to the public [Vulners Python API](https://github.com/vulnersCom/api).
It's completely supported and I think will make it easier to integrate.
It's also [available via Pypi](https://pypi.org/project/vulners/).

NerdsvilleCEO:
Resolves #438


basepi:
Sorry for my late reply here, I was out of the office (and out of cell service) for awhile.

This looks good -- can you please just add the vulners library to each of the pyinstaller-requirements.txt files inside of the `pkg` directory? As well as in the setup.py as well? That way our docker-built packages will include it, as will any source installs (that don't use pip).

And while you're at it, can you add `salt-ssh` to your requirements.txt file? If we're going to add an explicit requirements file, I want it to be complete.

NerdsvilleCEO:
@basepi I can definitely make these changes, I am re-setting up my development environment and can take care of this tomorrow at the latest

Thanks!

basepi:
No hurry! I appreciate these changes!

NerdsvilleCEO:
@basepi working on this now, had a quick question, is there a reason these all don't just share a single pyinstaller-requirements.txt?

basepi:
Mostly because we don't want to mount `../<something>` in the dockerfile. With the way we mount now it can only see the local directory, so we copy it to each location. It changes infrequently enough that it hasn't been a huge issue, though it still is an ugly solution.

NerdsvilleCEO:
@basepi could you please review the changes? Thanks!

vulnersCom:
Hi fellows! That's a really nice work, thank you!
May I ask you not to set vulners == 1.3.0 version?
Please, use latest, I'm checking backport compatibility.
There is a lot of performance tuning in 1.3.5.

NerdsvilleCEO:
I can make this change in a separate PR @vulnersCom thanks!

PR #445, opened 2018-08-16
Win dockerfile

madchills:
With a little help from @yassingh, we got this working!


yassingh:
Hubble is not  running as a service ( perhaps because PortableGit is not placed at right location ).
Other than that, the Dockerfile seems fine. Atleast, I am able to create hubble exe using this Dockerfile.

Let's merge the PR. 
I will keep making changes to the Dockerfile to make it more robust as I get more time.

PR #448, opened 2018-08-24
2.4.2 build broken

yassingh:
Line number 31 is not working. As a result nova module working is not working.
So I generated a PR with the code which worked for me.

If line no. 31 has some other intention, feel free to close my PR and fix it in the appropriate way.


basepi:
Hmm. I do need to look into this -- pretty sure I made that change so the docs system could import this file. Maybe I just need to gate it in some way instead.

PR #455, opened 2018-09-13
I sometimes desire to change the console log format.

jettero:
It makes sense to not allow this for the central logging, but I can't see the
harm in altering the console logging.


basepi:
This looks great. One note: you appear to have committed with an email not associated with your github account. You might want to fix that.

PR #458, opened 2018-09-19
Improving windows docker file

yassingh:
Following changes have been made in windows docker file:

1. Earlier, PortableGit was extracted in a wrong location. Fixed now.
2. Upgraded the salt version being used to 2018.11
3. If 'opt' folder is present in mounted location, it's contents are inserted into the build.
4. If 'hubble.conf' is present in the mounted location, it's content are inserted into the build.
5. SHA256 is calculated for the generated build.

### DIFF
--- a/pkg/windows/dockerfile
+++ b/pkg/windows/dockerfile
@@ -84,7 +84,7 @@ CMD powershell.exe -Command Push-Location C:/temp/hubble; \
 #Build the installer
   Push-Location 'C:/Program Files (x86)/NSIS'; \
   ./makensis.exe /DHubbleVersion="$env:HUBBLE_CHECKOUT" 'C:/temp/hubble/pkg/windows/hubble-Setup.nsi'; \
-  Get-FileHash -Path C:/temp/hubble/pkg/windows/Hubble*exe -Algorithm SHA256 | Select  Hash > hubble_windows.sha256; \
+  Get-FileHash -Path C:/temp/hubble/pkg/windows/Hubble*exe -Algorithm SHA256 > hubble_windows.sha256; \
   Copy-Item C:/temp/hubble/pkg/windows/Hubble*exe -Destination C:/data/; \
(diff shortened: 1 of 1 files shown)


pkg/windows/dockerfile
```diff
@@ -54,7 +54,7 @@ RUN powershell.exe -Command pip install -r pyinstaller-requirements.txt;
   
 #Move portable git to a new location
 RUN powershell.exe -Command New-Item C:/temp/hubble/PortableGit -ItemType Directory; \
-  Copy-Item -Path C:/tools/git -Destination C:/temp/hubble/PortableGit -Recurse;
```
madchills: In my tests you need to have a trailing / after PortableGit or it doesnt work.  This should be added here just in case

pkg/windows/dockerfile
```diff
@@ -54,7 +54,7 @@ RUN powershell.exe -Command pip install -r pyinstaller-requirements.txt;
...
+  Copy-Item -Path C:/tools/git/* -Destination C:/temp/hubble/PortableGit -Recurse;
```
madchills: In my tests you need to have a trailing / after PortableGit or it doesnt work. This should be added here just in case

pkg/windows/dockerfile
```diff
@@ -13,9 +13,9 @@ ENV CHOCO_URL=https://chocolatey.org/install.ps1
 #All the variables used for salt
 ENV SALT_SRC_PATH='C:/temp/salt/'
 ENV SALT_GIT_URL=https://github.com/saltstack/salt
-ENV SALT_CHECKOUT=v2018.3.0
+ENV SALT_CHECKOUT=v2018.11
```
madchills: Has this been tested to work?  I know in previous versions of salt (after 2018.3.0 but before 2018.11) they had changed the internal workings of their modules enough for it to not install through docker.

If you tested this, and it works good, then we don't need to worry

PR #459, opened 2018-09-19
a solution to timeouts in windows (lacking SIGALRM)

jettero:
The only way to actually do the timeouts in windows (near as I can tell) is to use a neighboring thread to ping the process and politely ask it to stop.

Rather than re-write hangtime to do all that magic and gracefully switch to threads for windows...

I looked at the package: stopit

It does all the magic things, is more standard in terms of os-systems programming, and is actually thread safe — for later, when we do multithreading.

I haven't tested this branch a lot, but I did test to make sure:

1. Nothing bad seems to happen if no timeouts trip during grains refreshes — though, I didn't test this over very long runs. I plan to keep looking for things, but long runs take a long time.

2. If a timeout occurs during some try: try: try: try: long_salt_operation() except: except: except ; stopit seems to correctly log the error and abort the decorated function (grains_refresh) as expected.

Once (2) worked and (1) didn't cause any problems on short runs, I posted the PR for consideration. The main question is: is stopit ok?


hubblestack/daemon.py
```diff
@@ -604,10 +604,6 @@ def __init__(self, message, levelname, asctime, name):
 # pick it up again.  If the hang is transient, the grain will populate
 # normally.
 #
-# repeats=True meaning: restart the signal itimer after firing the timeout
-# exception, which salt catches. In this way, we can catch multiple hangs with
-# a single timer. Each timer restart is a new 600s timeout.
-#
 # tag='hubble:rg' will appear in the logs to differentiate this from other
```
jettero: This is partially a lie. Normally the timeout exceptions will be caught by salt's try:except blocks internal to (eg) grains.core. But you still get the stopit log.warning() if nothing else.

basepi:
As discussed in our meeting today, I think we're just going to remove the timers in windows. I'll close this for now...

PR #472, opened 2018-10-09
updating systemd config to restart hubble on crash

praksinha:


praksinha:
@basepi Please review

PR #474, opened 2018-10-12
Hubble autostart feature

praksinha:
- Added control script to start/stop/restart hubble for servers where systemd is missing
- Scheduled control script using cron
- Fixed Dockerfiles for few flavours to make systemd conf work


conf/afterinstall.sh
```diff
@@ -1 +1,8 @@
+crontab -l > currentcrons
+# Try to start hubble every 5 mins. If hubble is already running, no action will be taken.
+echo "*/5 * * * * /bin/sh /etc/hubble/hubblectl.sh start" >> currentcrons
+# restart hubble once every day at 0700 hours. This is to handle scenarios where hubble goes into hang state
+echo "* 7 * * * /bin/sh /etc/hubble/hubblectl.sh restart" >> currentcrons
+crontab currentcrons
+rm currentcrons
```
basepi: I don't like editing the crontab by default with our packages. I think it's dangerous, and we also aren't removing the job on uninstall.

I think we'd be much better off using a cron.d/ file instead.

basepi:
Also, why are we not either using or updating https://github.com/hubblestack/hubble/blob/develop/pkg/hubble instead of creating a new script? Seems like an init script should be sufficient, and if it's not, maybe we should update the init script.

conf/afterinstall.sh
praksinha: Done

praksinha:
I didn't use /etc/init.d/ script since we will not be using init.d service to manage hubble in future. Due to this, there were chances that we remove/remove the script that we install under /etc/init.d which might break cron job. That script is indeed sufficient to start/restart hubble and can be used technically, we just have to make sure we do not remove it for flavours where systemd isn't present.

conf/afterinstall.sh
```diff
@@ -1 +1,5 @@
+echo "*/5 * * * * root /etc/init.d/hubble start" >> hubble-autostart
+# restart hubble once every day at 0700 hours
+echo "* 7 * * * root /etc/init.d/hubble restart" >> hubble-autostart
+mv hubble-autostart /etc/cron.d/
```
basepi: Rather than doing this in the afterinstall.sh, can we just create the hubble-autostart file and put it in the package? If we just include it in the package then we don't have to worry about the security implications of using a tempfile in this way (and it will be removed if the package is removed -- with this current implementation, `hubble-autostart` would remain in cron.d/ which is not acceptable).

praksinha:
Made changes as per review comments

PR #475, opened 2018-10-17
 Clean up nebula_osquery masking code

basepi:
* Better documentation
* Some linting
* Better error reporting

@yassingh This shouldn't affect you (as the only real change I made was switching to `mask_with` instead of `mask_by`, but I retained backwards compatibility) but it would be nice if you could review this anyway.


yassingh:
Changes seem ok to me.
Thanks for refactoring (my) code.

basepi:
This is pending discussion taking place on https://github.com/hubblestack/hubblestack_data/pull/127

Issue #453, opened 2018-09-07
Install on Windows 2012R2 failing


w500:
Hello,

When I use the powershell to create the windows package for hubblestack it gets partway through and then gives the following error. 

.3.0) (3.0.4)
Requirement already satisfied: pywin32>=223 in c:\python27\lib\site-packages (from pypiwin32; sys_platform == "win32"->C
herryPy==10.2.1->salt==2018.3.0) (223)
Requirement already satisfied: backports.functools-lru-cache in c:\python27\lib\site-packages (from cheroot>=5.2.0->Cher
ryPy==10.2.1->salt==2018.3.0) (1.5)
Requirement already satisfied: more-itertools>=2.6 in c:\python27\lib\site-packages (from cheroot>=5.2.0->CherryPy==10.2
.1->salt==2018.3.0) (4.3.0)
Requirement already satisfied: tempora>=1.8 in c:\python27\lib\site-packages (from portend>=1.6.1->CherryPy==10.2.1->sal
t==2018.3.0) (1.13)
Requirement already satisfied: smmap2>=2.0.0 in c:\python27\lib\site-packages (from gitdb2>=2.0.0->GitPython==2.1.7->sal
t==2018.3.0) (2.0.4)
Requirement already satisfied: pytz in c:\python27\lib\site-packages (from tempora>=1.8->portend>=1.6.1->CherryPy==10.2.
1->salt==2018.3.0) (2018.5)
Requirement already satisfied: jaraco.functools>=1.20 in c:\python27\lib\site-packages (from tempora>=1.8->portend>=1.6.
1->CherryPy==10.2.1->salt==2018.3.0) (1.20)
Installing collected packages: salt
  Running setup.py develop for salt
    Complete output from command c:\python27\python.exe -c "import setuptools, tokenize;__file__='C:\\temp\\salt\\setup.
py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file_
_, 'exec'))" develop --no-deps:
    2018.3.0
    running develop
    running install-pywin32-windows
    error: invalid command 'install-pywin32-windows'

    ----------------------------------------
Command "c:\python27\python.exe -c "import setuptools, tokenize;__file__='C:\\temp\\salt\\setup.py';f=getattr(tokenize,
'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" develop --n
[...]

Issue #406, opened 2018-06-18
Running `./create_build_env.ps1` to create windows package ends with `error: invalid command 'install-pywin32-windows'`


mcarpenterjr:
When creating the build environment to create the window's package the build ends/ fails with `error: invalid command 'install-pywin32-windows'` and prompts for a Repository and Full URL.

Exact Message -
```
Command "c:\python27\python.exe -c "import setuptools, tokenize;__file__='C:\\temp\\salt\\setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" develop --no-deps" failed with error code 1 in C:\temp\salt\
Enter a Repository (full URL only):
```

I'm building on a 64Bit system and scrolling back up to the begining it seems the install script recognizes this. Any help much appreciated.

mcarpenterjr:
Tried installing separately through `pip` and it seems to already be satisfied, looks to genuinely be a bad command inside the script. Was able to complete creating the build environment.

madchills:
I have talked to the guys at SaltStack, and they recognize this as a big in their code.  until they fix it I have found that instead of running line 64 "pip install -e .", you replace that with "python ./setup.py --quiet install --force;" and it seems to finish fine.  I will update this in the code

gjones623:
https://jira.corp.adobe.com/browse/CSTTEAM-44033

PR #478, opened 2018-10-23
Packaging ps1

madchills:


madchills:
This should fix #453 and #406

PR #480, opened 2018-10-31
emit some grains to syslog

fossam:
this helps to identify and correlate host syslog data with hubble data

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -36,6 +36,10 @@ from hubblestack.hangtime import hangtime_wrapper
 
 log = logging.getLogger(__name__)
 
+# Importing syslog fails on windows
+if not salt.utils.platform.is_windows():
+    import syslog
+
 __opts__ = {}
 # This should work fine until we go to multiprocessing
 SESSION_UUID = str(uuid.uuid4())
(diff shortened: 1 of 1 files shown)


hubblestack/daemon.py
```diff
@@ -690,6 +708,24 @@ def __init__(self, message, levelname, asctime, name):
...
+    '''
+    try:
+        # Avoid a syslog line to be longer than 1024 characters
+        # Build syslog message
+        syslog_list = []
+        syslog_list.append('hubble_syslog_message:')
+        for grain in grains_to_emit:
+            if grain in __grains__:
+                syslog_list.append('{0}={1}'.format(grain, __grains__[grain]))
+        syslog_message = ' '.join(syslog_list)
```
basepi: Some grains have spaces in them, which may make this hard to parse on the other end. I'd recommend using a different separator -- either a tab or a newline, probably.

fossam: In syslog we can't do newlines and tabs may be not be nice and tricky to parse or look at on the host syslog log file. On mine if I replace the space with tab it will look like this:
```
hubble.py[14194]: hubble_syslog_message:#011system_uuid=92C3F82D-CDD1-4170-904A-AC6459A94287#011hubble_uuid=954d118d-5cb2-4023-9fb3-6a89c330fa17#011session_uuid=...
```

basepi: After chatting with @fossam on slack we're going to leave this as is for now.

PR #482, opened 2018-11-01
FDG topfile support

basepi:
Topfile support for FDG. Also some other stuff:

* Support for initial chained values in FDG
* Small fixes to conf/hubble that I noticed during testing


basepi:
Added the fdg osquery module to this as well, since nobody has reviewed it yet.

basepi:
Ready for review again. Finished testing osquery.py

PR #483, opened 2018-11-02
log the actual error. why is this ever happening (fairly rarely)

jettero:


hubblestack/extmods/modules/pulsar.py
```diff
@@ -552,8 +552,8 @@ def _wrapped(val):
                 try:
                     c = re.compile(r)
                     the_list.append(re_wrapper(c))
-                except:
-                    log.warn('Failed to compile regex: {0}'.format(r))
+                except Exception as e:
+                    log.warn('Failed to compile regex "%s": %s', r,e)
```
basepi: I would just use `log.exception` instead, since it gives us the traceback:
```
                    log.exception('Failed to compile regex "%s"', r)
```
I think a regex not compiling is an ERROR-worthy issue.

jettero: I can do that. AM wanted _less_ logging though. I figured just the error itself would be enough. No?

basepi: Hmm. I mean, I guess you're right that we don't need the call stack in this case since the context of the exception is so small. Normally you want that context because try except blocks often include many lines, and you want to know exactly where things went wrong.

You've won me over, can you switch it back? xD

jettero: reverted

PR #485, opened 2018-11-02
FDG Curl module

basepi:
@fossam You'll have to tell me if this has any gaps around your use case. I left it pretty open (which I think is still ok from a security standpoint since I disabled fdg chaining for this module)


basepi:
Merging. I'll create a new PR for any additions we need as a result of @fossam's discussion with me.

PR #488, opened 2018-11-09
Add FDG module for reading data from files

basepi:
Adds the ability to pull json/yaml data from a file, including the ability to pull data from a subkey of a data structure. Also includes the ability to have nested subkeys with a custom separator. Finally, includes the ability to pull somewhat arbitrary data from "config-style" files.


basepi:

PR #506, opened 2018-11-30
Osquery daemon integration - DO NOT MERGE

praksinha:
Following are the requirements:
======================

1) Hubble should be able to update osqueryd configuration file and flag file.
2) Should be able to parse osqueryd log file(s) and send to splunk in efficient manner, probably cache file offset etc.
3) Should be able to rotate osqueryd's log file
4) Should have capabilities to start/stop/restart osquery daemon based on hubble profile flags.
5) osqueryd should be in running state, we can need to implement monitoring capability in hubble itself to take care of this.
6) There should be an option to enable/disbale events table in osqueryd. By enabling, I mean hubble should stop auditd in case it is already running and also make sure that it is kept in stopped state while hubble is in running state in order for osquery event tables to work. Although this can be handled using discovery queries in packs.
7) Hubble should have capabilities to insert custom fields in logs that are sent to splunk.
8) Design should be single process and should consume at-most 1 CPU core as SREs might not be ok otherwise. We may or may not chose make this multi-threaded depending on performance and python support etc.
9) Should support both osqueryi and osquery daemon mode.

Final High level design post discussion with team:
====================================

1) Conf and flag files in /var/cache/hubble directory
2) Hubble nebula module to have two separate functions scheduled at probably different frequency - osqueryd_monitor and osqueryd_log_parser
3) One function would be responsible for managing/monitoring osquery daemon like start/stop/restart. Should be responsible for keeping osquery daemon in running state.
4) Other function would be responsible for parsing osquery logs and sending to splunk
5) In case auditd is enabled, we can have following two approaches:
    a) Dynamically generate osquery flag file to disable events as they anyway won't work but continue to execute standalone queries like we have in osqueryi
    b) Create a separate pack for evented queries and execute them based on discovery query (conditional execution) which would check if auditd process is running or not. (As of now, opting for this approach)
6) Hubble would continue to support existing nebula functions to schedule queries through osqueryi. Depending on Hubble conf, users can make use of osqueryi or osqueryd to execute queries
7) There should be a way to distinguish sourcetype of osqueryi and osqueryd. Due to difference in logging format of osqueryi and osqueryd, new splunk returner is being created. This would allow users to specify custom sourcetype.
8) We'll keep at-least one backup log file of osquery (after logfile truncation). This number should be configurable and can be tweaked as per individual team's needs. Like if DME wants 5 log backups and DMA wants just 1 backup, it should be doable. Regarding caching offset on filesystem, we'll see if we can cover all corner cases without saving the offset to disk. If not, we'll probably make use of offset and save it to disk.
9) Since Hubble would be dealing with log file rotation, it would be good to have a feature where-in we can check disk usage. If disk/partition is full or is above a certain threshold then either do not make backups or alert users in some manner. We can discuss this further in more detail in subsequent meetings. Implemented a basic disk usage functionality, it's usage however is still to be finalized.


pkg/windows/dockerfile
```diff
@@ -81,6 +81,7 @@ CMD powershell.exe -Command Push-Location C:/temp/hubble; \
   Move-Item 'hubble/PortableGit' -Destination './hubble/dist/hubble/' -Force; \
   Move-Item 'C:/ProgramData/chocolatey/lib/NSSM/tools/nssm.exe' -Destination './hubble/dist/hubble/' -Force; \
   Move-Item 'C:/ProgramData/osquery/osqueryi.exe' -Destination './hubble/dist/hubble/' -Force; \
+  Move-Item 'C:/ProgramData/osquery/osqueryd.exe' -Destination './hubble/dist/hubble/' -Force; \
```
praksinha: Dockerfile changes are not yet finalized, any changes in dockerfiles just reflect what could potentially change.

pkg/dev/amazonlinux2016.09/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN mkdir -p /etc/osquery /var/log/osquery /etc/hubble/hubble.d /opt/hubble /opt
 #osquery should be built first since requirements for other packages can interfere with osquery dependencies
 #to build, osquery scripts want sudo and a user to sudo with.
 #to pin to a different version change the following envirnment variable
-ENV OSQUERY_SRC_VERSION=1ed050147ac88a7314c172e15e86f8886206dd06
+ENV OSQUERY_SRC_VERSION=0314871908273b2770129da9f5a7c3206f224f68
```
praksinha: Dockerfile changes are not yet finalized, any changes in dockerfiles just reflect what could potentially change.

conf/custom-dme-hubble.conf
```diff
@@ -0,0 +1,169 @@
+#################################
```
praksinha: This is just a sample config which will be removed eventually and actually sample config etc would be added in respective files.

This file is added to provide an example configuration that can be used to test osqueryd_monitor and osqueryd_log_parser functions.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -241,6 +247,213 @@ def queries(query_group,
     return ret
 
 
+def osqueryd_monitor(servicename='hubble_osqueryd',
```
praksinha: In case of any concerns w.r.t to monitor functionality, contact 'muagarwa@adobe.com'

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -241,6 +247,213 @@ def queries(query_group,
...
+        osqueryd_running = _osqueryd_running_status(pidfile, servicename)
+        if not osqueryd_running:
+            _start_osqueryd(pidfile, configfile, flagfile, logdir, databasepath, servicename)
+        osqueryd_restart = _osqueryd_restart_required(hashfile, flagfile)
+        if osqueryd_restart:
+            _restart_osqueryd(pidfile, configfile, flagfile, logdir, databasepath, hashfile, servicename)
+
+
+def osqueryd_log_parser(osqueryd_logdir=None, 
+                        backuplogdir=None,
```
praksinha: In case of any concerns with 'log_parsing' functionality, please contact 'prsinha@adobe.com'

PR #507, opened 2018-12-01
Osquery daemon integration in Hubble

praksinha:
Following are the requirements:
======================

1) Hubble should be able to update osqueryd configuration file and flag file.
2) Should be able to parse osqueryd log file(s) and send to splunk in efficient manner, probably cache file offset etc.
3) Should be able to rotate osqueryd's log file
4) Should have capabilities to start/stop/restart osquery daemon based on hubble profile flags.
5) osqueryd should be in running state, we can need to implement monitoring capability in hubble itself to take care of this.
6) There should be an option to enable/disbale events table in osqueryd. By enabling, I mean hubble should stop auditd in case it is already running and also make sure that it is kept in stopped state while hubble is in running state in order for osquery event tables to work. Although this can be handled using discovery queries in packs.
7) Hubble should have capabilities to insert custom fields in logs that are sent to splunk.
8) Design should be single process and should consume at-most 1 CPU core as SREs might not be ok otherwise. We may or may not chose make this multi-threaded depending on performance and python support etc.
9) Should support both osqueryi and osquery daemon mode.

Final High level design post discussion with team:
====================================

1) Conf and flag files in /var/cache/hubble directory
2) Hubble nebula module to have two separate functions scheduled at probably different frequency - osqueryd_monitor and osqueryd_log_parser
3) One function would be responsible for managing/monitoring osquery daemon like start/stop/restart. Should be responsible for keeping osquery daemon in running state.
4) Other function would be responsible for parsing osquery logs and sending to splunk
5) In case auditd is enabled, we can have following two approaches:
    a) Dynamically generate osquery flag file to disable events as they anyway won't work but continue to execute standalone queries like we have in osqueryi
    b) Create a separate pack for evented queries and execute them based on discovery query (conditional execution) which would check if auditd process is running or not. (As of now, opting for this approach)
6) Hubble would continue to support existing nebula functions to schedule queries through osqueryi. Depending on Hubble conf, users can make use of osqueryi or osqueryd to execute queries
7) There should be a way to distinguish sourcetype of osqueryi and osqueryd. Due to difference in logging format of osqueryi and osqueryd, new splunk returner is being created. This would allow users to specify custom sourcetype.
8) We'll keep at-least one backup log file of osquery (after logfile truncation). This number should be configurable and can be tweaked as per individual team's needs. Like if DME wants 5 log backups and DMA wants just 1 backup, it should be doable. Regarding caching offset on filesystem, we'll see if we can cover all corner cases without saving the offset to disk. If not, we'll probably make use of offset and save it to disk.
9) Since Hubble would be dealing with log file rotation, it would be good to have a feature where-in we can check disk usage. If disk/partition is full or is above a certain threshold then either do not make backups or alert users in some manner. We can discuss this further in more detail in subsequent meetings. Implemented a basic disk usage functionality, it's usage however is still to be finalized.

### DIFF
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -24,6 +24,7 @@ nebula_osquery:
 from __future__ import absolute_import
 
 import copy
+import glob
 import json
 import logging
 import os
@@ -33,14 +34,19 @@ import yaml
 import collections
 
 import salt.utils
+import salt.utils.files
 import salt.utils.platform
+
+from hashlib import md5
 from salt.exceptions import CommandExecutionError
+from os import path
 from hubblestack import __version__
 import hubblestack.splunklogging
 
 log = logging.getLogger(__name__)
 
 __virtualname__ = 'nebula'
+__RESULT_LOG_OFFSET__ = {}
 
 
 def __virtual__():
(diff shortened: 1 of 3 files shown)


praksinha:
@basepi Yes data format is different from what we currently have with osqueryi. Osquery daemon can produce two types of logs in general - differential and snapshot. These logs have few additional fields like 'action', 'epoch', 'counter' etc and also the attributes in event generated might differ.

For example, if we have a running_procs query output in differential format, it would have following format:

{
name: 'running_procs',
action: 'added',
epoch: 123,
counter: 456,
unixTime: 178624682,
columns: {} // Note this field has dict type. Contains actual row returned.
...more fields
}


Same query in snapshot format would have following event format:

{
name: 'running_procs',
action: 'snapshot',  // Note value here would always be snapshot
epoch: 123,
counter: 456,
unixTime: 178624682,
snapshot: [] // Note this field has list type. Contains actual row(s) returned.
...more fields
}

As seen from above, result logs have a field name 'columns' which is of type 'dict' whereas snapshot logs have field 'snapshot' which is of type 'list'.

Logs generated by osqueryi are of following format:

{
'running_procs' : {
                               'data' : [] //Contains actual rows returned
                              }
}


If we want to use same returner for both osqueryi and osqueryd events, then we'll have convert both snapshot logs and result logs into current format. This would be tedious, so I created a new returner instead of formatting each event read from osquery log files (which will eventually again be re-formatted in returner code before being sent to splunk in form of event_data/payload object).

basepi:
Perfect, thanks for the reply and explanation. I think you're right that a new returner makes sense.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -580,3 +794,334 @@ def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
...
+              except:
+                log.error("process's cmdline cannot be determined, attempting to start osqueryd")
+            else:
+              log.error("process folder not present, attempting to start osqueryd")
+          else:
+            log.error("pid cannot be determined, attempting to start osqueryd")
+      except:
+        log.error("unable to open pidfile, attempting to start osqueryd")
+    else:
+      cmd = ['pkill', 'osqueryd']
```
praksinha: pkill osqueryd might kill osquery daemon instances started by user outside/independent of Hubble. Is it the only way to kill osquery daemon ? Can process name be tweaked to something like 'hubble_osqueryd' and only pkill hubble_osqueryd ?

basepi:
@praksinha As I mentioned on slack, can we have globbing behind a flag? I don't want to hit your performance, but we want to optionally be able to re-enable it.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -246,6 +253,249 @@ def queries(query_group,
...
+    log.info('cached osqueryd files to cachedir')
+    cachedir = os.path.join(__opts__.get('cachedir'),'files',saltenv,'osqueryd')
+    base_path = cachedir
+    servicename = "hubble_osqueryd"
+    if not logdir:
+            logdir = __opts__.get('osquerylogpath')
+    if not databasepath:
+            databasepath = __opts__.get('osquery_dbpath')
+    if salt.utils.platform.is_windows():
+        log.info("System is windows")
```
basepi: We can probably get rid of these logs (this line and on the `else` as well). In general I don't think we're wondering whether a system is windows or linux when we're looking at logs, and `.info` should be reserved for more significant info.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -592,3 +902,293 @@ def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
...
+                      logdir, 
+                      databasepath, 
+                      hashfile, 
+                      servicename):
+    '''
+    This function will restart osqueryd
+    ''' 
+    log.info("osqueryd needs to be restarted, restarting now")
+
+    open_file = open(flagfile, 'r')
```
praksinha: Either use 'with' clause or explicitly close 'open_file' handle

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -592,3 +902,293 @@ def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
...
+    file_content = open_file.read().lower().rstrip('\n\r ').strip('\n\r')
+    hash_md5 = md5()
+    hash_md5.update(file_content.encode('ISO-8859-1'))
+    new_hash = hash_md5.hexdigest()
+
+    f = open(hashfile, "w")
```
praksinha: Either use 'with' clause or explicitly close 'f' file handle

hubblestack/extmods/modules/nebula_osquery.py
MoodyMudit: sure, I'll remove these logs

hubblestack/extmods/modules/nebula_osquery.py
MoodyMudit: ok, I'll do this.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -246,6 +253,254 @@ def queries(query_group,
...
+
+    '''
+    ret = []
+    if osqueryd_logdir:
+        log.info("Base directory for osquery daemon logs specified as {0}".format(osqueryd_logdir))
+        result_logfile = os.path.normpath(os.path.join(osqueryd_logdir, 'osqueryd.results.log'))
+        snapshot_logfile = os.path.normpath(os.path.join(osqueryd_logdir, 'osqueryd.snapshots.log'))
+    else:
+        osquery_base_logdir = __opts__.get('osquerylogpath')
+        log.info("Setting osquery daemon log file to default dir: {0}"
```
basepi: I think we can remove this log since we already have the debug logs below.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -246,6 +253,254 @@ def queries(query_group,
...
+        will be extracted
+
+    mask_passwords
+        Defaults to False. If set to True, passwords mentioned in the
+        return object are masked
+
+    '''
+    ret = []
+    if osqueryd_logdir:
+        log.info("Base directory for osquery daemon logs specified as {0}".format(osqueryd_logdir))
```
basepi: I think we can remove this log since we already have the debug logs below.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -246,6 +253,254 @@ def queries(query_group,
...
+                                                                                                per_used))
+        disk_stats = {'Total' : total,
+                      'Available' : avail,
+                      'Used' : used,
+                      'Use_percent' : per_used,
+                      'Path' : path
+                     }
+
+        if __salt__['config.get']('splunklogging', False):
+            log.info('Logging disk usage stats to splunk')
```
basepi: Can we move this do `debug` instead of `info`?

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -425,7 +662,10 @@ def _mask_object(object_to_be_masked, topfile):
     try:
         mask = {}
         if topfile is None:
-            topfile = 'salt://hubblestack_nebula_v2/top.mask'
+            # We will maintain backward compatibility by keeping two versions of top files and mask files for now
+            # Once all hubble servers are updated, we can remove old version of top file and mask file
+            # Similar to what we have for nebula and nebula_v2 for older versions and newer versions of profiles
+            topfile = 'salt://hubblestack_nebula_v2/top_v2.mask'
```
basepi: What backwards-incompatible change happened to the topfile? Can you clarify?

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -592,3 +907,295 @@ def _dict_update(dest, upd, recursive_update=True, merge_lists=False):
...
+        hash_md5 = md5()
+        hash_md5.update(file_content.encode('ISO-8859-1'))
+        new_hash = hash_md5.hexdigest()
+
+        if not os.path.isfile(hashfile):
+            f = open(hashfile, "w")
+            f.write(new_hash)
+            return False
+        else:
+            f = open(hashfile, "r")
```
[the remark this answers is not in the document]
MoodyMudit: Closed all the mentioned file pointers explicitly

hubblestack/extmods/modules/nebula_osquery.py
basepi: Sounds fine. I didn't catch the schema change in my cursory review, I'll look closer.

hubblestack/extmods/modules/nebula_osquery.py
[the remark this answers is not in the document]
basepi: I appreciate the addition of the explicit `.close()` calls, but instead can you use `with` blocks? Explicit `close()` calls don't get called if there is an exception for some reason, and we can leak file descriptors.

There are only two ways to properly deal with files in python:

```
try:
    f = open('file')
finally:
    f.close()
```
and
```
with open('file') as f:
    pass
```
The latter is cleaner, in my opinion, so please use `with` blocks for all open files.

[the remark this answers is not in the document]
basepi: Sorry, I missed your second commit where you fixed this. Thanks!

PR #508, opened 2018-12-01
add brief note about returners

### LINKED ISSUE #502: Add returner information to the docs
Just getting started, I had to dig around a bit to find some information on how returners worked so adding some information to the docs would be helpful for others I think. I wrote this up as a brief snippet if you think it would be helpful.

> It is quite possible you may want to send the NOVA audit results to a collection system. This can be accomplished by using returners, which are included with Hubble. Currently, there are returners available for splunk, logstash, graylog, and sqlite. Hubble supports using multiple returners from the same node so sending data to multiple collectors is possible. Each returner has distinct options that need to be set prior to their use, which you can find the general config options for each returner at the top of each file located here https://github.com/hubblestack/hubble/tree/develop/hubblestack/extmods/returners.
>
> Specify a returner when running Hubble from the command line like this:
    hubble hubble.audit --r RETURN, --return RETURN
>
> You can also define schedules for returners in the Hubble config file, which is typically located at /etc/hubble/hubble.conf.

alias454:
Fixes #502

### DIFF
--- a/doc/contents/auditing_nova.rst
+++ b/doc/contents/auditing_nova.rst
@@ -121,3 +121,20 @@ following to your ``top.nova`` file::
     nova:
       '*':
         - example_control.example
+
+It is quite possible you may want to send the NOVA audit results to a collection system. 
+This can be accomplished by using returners, which are included with Hubble. Currently, 
+there are returners available for **splunk**, **logstash**, **graylog**, and **sqlite**.
+Hubble supports using multiple returners from the same node so sending data to multiple 
+collectors is possible. Each individual returner has distinct options that need to be 
+set prior to their use. You can find general config options for each returner at the top 
+of each file located here
+https://github.com/hubblestack/hubble/tree/develop/hubblestack/extmods/returners.
+
+**Specify a returner when running Hubble from the command line:**::
+
+    hubble hubble.audit --r RETURN, --return RETURN
+
+You can also define schedules for returners in the Hubble config file, which is typically
+located at ``/etc/hubble/hubble.conf``.
+
--- a/doc/contents/auditing_nova.rst
+++ b/doc/contents/auditing_nova.rst
@@ -131,10 +131,13 @@ set prior to their use. You can find general config options for each returner at
 of each file located here
 https://github.com/hubblestack/hubble/tree/develop/hubblestack/extmods/returners.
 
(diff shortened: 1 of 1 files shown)


doc/contents/auditing_nova.rst
```diff
@@ -121,3 +121,20 @@ following to your ``top.nova`` file::
...
+set prior to their use. You can find general config options for each returner at the top 
+of each file located here
+https://github.com/hubblestack/hubble/tree/develop/hubblestack/extmods/returners.
+
+**Specify a returner when running Hubble from the command line:**::
+
+    hubble hubble.audit --r RETURN, --return RETURN
+
+You can also define schedules for returners in the Hubble config file, which is typically
+located at ``/etc/hubble/hubble.conf``.
```
basepi: `/etc/hubble/hubble`, no `.conf`

doc/contents/auditing_nova.rst
```diff
@@ -121,3 +121,20 @@ following to your ``top.nova`` file::
...
+there are returners available for **splunk**, **logstash**, **graylog**, and **sqlite**.
+Hubble supports using multiple returners from the same node so sending data to multiple 
+collectors is possible. Each individual returner has distinct options that need to be 
+set prior to their use. You can find general config options for each returner at the top 
+of each file located here
+https://github.com/hubblestack/hubble/tree/develop/hubblestack/extmods/returners.
+
+**Specify a returner when running Hubble from the command line:**::
+
+    hubble hubble.audit --r RETURN, --return RETURN
```
basepi: Can we split this into two lines and add a `code-block` stanza?

```
.. code-block::

    hubble hubble.audit --r RETURN
    hubble hubble.audit --return RETURN
```

doc/contents/auditing_nova.rst
alias454: Thanks. .conf removed from the path.

doc/contents/auditing_nova.rst
alias454: Lines have been split and inserted into the code block. Thanks for the feedback.

basepi: I actually missed the fact that you had the trailing `::` -- you actually had the code block piece right already! But this is more explicit, so I think it's still a good change. Thanks for making those changes!

PR #509, opened 2018-12-04
Use print() function in both Python 2 and Python 3

cclauss:
__print()__ is a function in Python 3.


cclauss:
I will keep pushing in changes so that you are not up against the clock: http://pythonclock.org

PR #515, opened 2018-12-13
FDG splunk returner

basepi:
The returner for FDG has to be much more flexible than other returners because we don't know in advance in what format the data will be.

Here are the design requirements:

* Data formed as a list should be put in multiple events
* All events should have an identifier, which is a normalized fdg file name
* fdg.top results should be handled as if fdg.fdg were called on the individual files. The splunk events will look the same either way

In the process I had to tweak the return structures for the fdg system, but I think it's better for it.

I also removed the empty field stripping from the `hubblestack.hec.obj.Payload` class -- it was stripping in such a way that there were side effects, and really we only want to strip at the top-level of the event anyway in most cases -- something which is already done in the individual returners that consume this shared code. We don't want to strip all empty fields at all levels of the data in every case, so instead we'll leave that as an exercise for the individual returners, as they can tailor the stripping to the individual use cases. @jettero FYI, since this is your code.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -437,10 +437,10 @@ def run_function():
             __returners__[returner](returner_ret)
 
     # TODO instantiate the salt outputter system?
-    if(__opts__['json_print']):
+    if __opts__['json_print']:
         print(json.dumps(ret))
     else:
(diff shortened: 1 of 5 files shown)


basepi:

PR #516, opened 2018-12-14
Add globbing support to object masking in nebula

basepi:
This should drastically reduce the number of entries we need to get most of the secrets masked.


basepi:

basepi:
I'm going to merge these changes here as well so they're not necessarily tied to such a big PR. You may have to rebase.

PR #520, opened 2018-12-19
add function to remove dict keys that have values of None

Tenebriso:


hubblestack/extmods/fdg/process.py
```diff
@@ -321,3 +321,70 @@ def _split(phrase,
...
+    seq
+        The input sequence to sterilize
+    '''
+    updated_seq = []
+    for element in seq:
+        if isinstance(element, dict):
+            updated_seq.append(_sterilize_dict(element))
+        elif isinstance(element, (list, set)):
+            updated_seq.append(_sterilize_seq(element))
+        elif element is not None:
```
basepi: We only want to remove dict values -- None values in lists should remain. Should be an easy fix.

PR #525, opened 2018-12-30
fix various bad error handlings in the hec.obj

jettero:
* the misc exception catch fails to log anything (but almost never comes up anyway)
* if the status is >= 400, the message was requeued regardless of the reply; fixed to discard if the message is malformatted (and log the fact), or to requeue on other errors unknown --- and in either case, we log as much of the reply as possible for later analysis
* the check for `if self.log_other_exceptions` is flawed (no such settting exists)

### DIFF
--- a/hubblestack/hec/obj.py
+++ b/hubblestack/hec/obj.py
@@ -291,19 +291,20 @@ class HEC(object):
                 r = self.pool_manager.request('POST', server.uri, body=data, headers=self.headers)
                 server.fails = 0
             except urllib3.exceptions.LocationParseError as e:
-                if self.log_other_exceptions:
-                    log.error('server uri parse error "{0}": {1}'.format(server.uri, e))
+                log.error('server uri parse error "{0}": {1}'.format(server.uri, e))
                 server.bad = True
                 continue
             except Exception as e:
+                log.error('misc exception: %s', e)
                 server.fails += 1
                 continue
             if r.status < 400:
                 return r
-
-        log.error('failed to send payload, queueing for later delivery')
-        self._requeue(payload)
-
+            elif r.status == 400 and r.reason.lower() == 'bad request':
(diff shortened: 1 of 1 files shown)


hubblestack/hec/obj.py
```diff
@@ -291,19 +291,20 @@ def _send(self, *payload):
...
                 server.fails += 1
                 continue
             if r.status < 400:
                 return r
-
-        log.error('failed to send payload, queueing for later delivery')
-        self._requeue(payload)
-
+            elif r.status == 400 and r.reason.lower() == 'bad request':
+                log.error('message not accepted (%d %s), dropping payload: %s', r.status, r.reason, r.data)
```
jettero: eg, `message not accepted (400 Bad Request), dropping payload: {"text":"Event field is required","code":12,"invalid-event-number":0}`

hubblestack/hec/obj.py
```diff
@@ -291,19 +291,20 @@ def _send(self, *payload):
...
+            else:
+                log.error('message not accepted (%d %s), requeueing: %s', r.status, r.reason, r.data)
```
jettero: this will come up occasionally when the hec disk pools are full or there's some other transient error the hec can reply to, but still can't accept the message.

It may possibly come up for other errors of which I'm not aware. The splunk documentation on these results is a little thin.

hubblestack/hec/obj.py
```diff
@@ -291,19 +291,20 @@ def _send(self, *payload):
...
+                log.error('misc exception: %s', e)
                 server.fails += 1
                 continue
             if r.status < 400:
                 return r
-
-        log.error('failed to send payload, queueing for later delivery')
-        self._requeue(payload)
-
+            elif r.status == 400 and r.reason.lower() == 'bad request':
```
basepi: Would it be worth just using `.raise_for_status()` rather than a series of `if` statements? I guess it doesn't really matter, these additions should catch anything >=400.

PR #526, opened 2018-12-30
message submission accounting

jettero:
I'm not totally sure if we should call this a work in progress, or if @fossam will want changes to how the counters submit; but this is totally working for me with the below query.

(NOTE: do not actually do the below if your dataset is at all bigger than about a few thousand events... that O(N²) map search will kill your Splunk)
```
index=hubble sourcetype=hubble_hec_counters 
| table index stype count start end 
| dedup stype
| map search="tstats count as actual where index=$index$ sourcetype=$stype$ earliest=$start$ latest=$end$ | eval reported=$count$ | eval stype=\"$stype$\"" maxsearches=100
| fields stype actual reported 
| eval diff=abs(actual-reported)
| sort -diff
```

### DIFF
--- a/hubblestack/extmods/modules/hstatus.py
+++ b/hubblestack/extmods/modules/hstatus.py
@@ -8,7 +8,7 @@ log = logging.getLogger(__name__)
 
 __virtualname__ = 'hstatus'
 
-SOURCETYPE = 'hubble_audit_summary'
+SOURCETYPE = 'hubble_hec_summary'
 MSG_COUNTS_PAT = r'hubblestack.hec.obj.input:(?P<stype>[^:]+)'
 
 def __virtual__():
--- a/hubblestack/extmods/modules/hstatus.py
+++ b/hubblestack/extmods/modules/hstatus.py
@@ -38,10 +38,11 @@ def msg_counts(pat=MSG_COUNTS_PAT, reset=True, emit_self=False, sourcetype=SOURC
             # a couple counts before we report/reset.
             if v['last_t'] <= v['first_t'] + 1:
                 continue
-            if v['event_count'] < 1:
+            if v['count'] <= 1:
                 continue
         except KeyError:
             continue
+
         m = pat.match(k)
         if m:
             try:
--- a/hubblestack/extmods/modules/hstatus.py
+++ b/hubblestack/extmods/modules/hstatus.py
@@ -25,13 +25,11 @@ def msg_counts(pat=r'hubblestack.hec.obj.input:(?P<stype>[^:]+)', reset=True):
             pat - the key matching algorithm is a simple regular expression
     '''
 
-    if reset:
-        log.error("TODO: reset timers")
-
     km = re.compile(pat)
     r = list()
     s = get()
     fudge_me = None
+    rst = set()
     for k,v in s.iteritems():
         try:
             if v['first_t'] == 0 or v['last_t'] == 0:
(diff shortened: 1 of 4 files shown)


hubblestack/extmods/modules/hstatus.py
```diff
@@ -11,12 +16,50 @@ def get():
...
+def msg_counts(pat=r'hubblestack.hec.obj.input:(?P<stype>[^:]+)', reset=True):
+    ''' returns counter data formatted for the splunk_generic_return returner
+
+        params:
+            pat - the key matching algorithm is a simple regular expression
+    '''
+
+    km = re.compile(pat)
+    r = list()
+    s = get()
```
basepi: Can we rename `get()` to `get_stats()` or something? I like to be explicit and self-documenting where possible.

hubblestack/extmods/modules/hstatus.py
```diff
@@ -11,12 +16,50 @@ def get():
 
         (probably only useful from other excution modules)
     '''
-    return hubblestack.status.HubbleStatus.as_json()
+    return hubblestack.status.HubbleStatus.stats()
+
+def msg_counts(pat=r'hubblestack.hec.obj.input:(?P<stype>[^:]+)', reset=True):
```
basepi: Lots of one-letter variable names in this one. It makes the code pretty hard to follow. `s` could be `stats`, `r` could be `ret`, `m` could be `match`, and `d` could be `matchdict` or something. It's a little more verbose, but brings some clarity.

hubblestack/extmods/modules/hstatus.py
```diff
@@ -11,12 +16,50 @@ def get():
...
+        return { 'sourcetype': 'hubble_hec_counters', 'time': min_time, 'events': r }
 
 def dump():
     ''' trigger a dump to the status.json file as described in hubblestack.status
 
         This is intended to be invoked from a daemon schedule and is probably
         not useful outside that context.
     '''
-    return hubblestack.status.HubbleStatus.dumpster_fire()
+    hubblestack.status.HubbleStatus.dumpster_fire()
```
basepi: I'm thinking we should return this data. At least that way it will show up in the debug logs -- just throwing it away is probably not the best move.

basepi:
I also definitely want @fossam to look at this before we merge it.

jettero:
I found a bunch of bugs, some where probably from conflict resolution and others were just things I never tested (like directly queueing events to disk before trying to send; yeah, it was bugged, but it probably won't get used either).

But one apparent bug seems problematic: the counts are suddenly wrong in spots. I'm waiting for the events to fall off the last 15 minute search so I can get a test without crashes in the status counters.

Failing that, I'll have to bisect to find the problem. *sigh*

jettero:
Yeah, false alarm... the counts from several hours are all super duper. They got pretty screwed up when some of the msg accounting events couldn't be formatted due to something like syntax errors.

hubblestack/extmods/modules/hstatus.py
```diff
@@ -1,17 +1,66 @@
...
+import re
+import logging
+import math
 import hubblestack.status
 
+log = logging.getLogger(__name__)
+
 __virtualname__ = 'hstatus'
 
+SOURCETYPE = 'hubble_audit_summary'
```
basepi: `hubble_hec_summary` instead please

Issue #464, opened 2018-09-24
hubble --version not working


yassingh:
I am facing an issue with the newer hubble builds.
Actually I have faced this issue in 2.4.1 also . Now I am facing this issue in 2.4.5 also.

Sometimes ( most of the times ), "hubble --version" and "hubble --buildinfo" command don't give any output.

"hubble --version -vvv" also doesn't give any output.

However, if I add any random string in front of these commands , i get perfect output.

For example:
"hubble --version xyz" and "hubble --buildinfo 123" work fine.

This is a random issue. Haven't found any pattern yet.

basepi:
We moved the daemonization of the process (and the check for existing hubble processes) earlier in the daemon. What's happening is hubble is aborting because the daemon is already running on the system. We need to add an explicit exception for `--version`, which should be pretty simple.

Note the reason it works if you add additional characters is because there *is* an exception already for single-function runs, such as `hubble test.ping` or `hubble hubble.top`. By adding the additional characters you make hubble think that's what's happening, even if it never gets that far because `--version` is executed first.

It's an easy fix, and I'll get it in later this week when I'm back from sick leave. (My appendix was removed yesterday)

PR #527, opened 2019-01-04
Distinguish between real and develop build

yassingh:
# CAN BE MERGED
This PR aims to provide a better way to identify if a build was created with a proper git tag, or if it was created with the code in develop branch.

All suggestions are welcome. I have given some explanations along the code.

My idea is :

To distinguish the build from inside ( i.e. after installing it ), we will use "hubble --buildinfo" command. It will provide a lot of info.

In order to know about the build before installing it, we can have a look at its iteration number. I good production build should have 1,2,3,... like iteration number. A huge and weird number indicates that it is not meant for production.

### DIFF
--- a/pkg/debian9/Dockerfile
+++ b/pkg/debian9/Dockerfile
@@ -153,8 +153,8 @@ RUN git clone "$HUBBLE_GIT_URL" "$HUBBLE_SRC_PATH" \
  && cd "$HUBBLE_SRC_PATH" \
  && git checkout "$HUBBLE_CHECKOUT" \
  && cp -rf "$HUBBLE_SRC_PATH" /hubble_build \
- && sed -i bak 's/BRANCH_NOT_SET/${HUBBLE_CHECKOUT}/g' /hubble_build/hubblestack/__init__.py \
- && sed -i bak 's/COMMIT_NOT_SET/TAGGED_BUILD/g' /hubble_build/hubblestack/__init__.py
+ && sed -i "s/BRANCH_NOT_SET/${HUBBLE_CHECKOUT}/g" /hubble_build/hubblestack/__init__.py \
+ && sed -i "s/COMMIT_NOT_SET/TAGGED_BUILD/g" /hubble_build/hubblestack/__init__.py
 RUN mkdir /data
 VOLUME /data
 WORKDIR /hubble_build
--- a/pkg/debian9/Dockerfile
+++ b/pkg/debian9/Dockerfile
@@ -157,7 +157,8 @@ RUN mkdir /data
 VOLUME /data
 WORKDIR /hubble_build
 ENTRYPOINT [ "/bin/bash", "-o", "xtrace", "-c" ]
-CMD [ "pyinstaller --onedir --noconfirm --log-level ${_BINARY_LOG_LEVEL} --additional-hooks-dir=${_HOOK_DIR} --runtime-hook=pkg/pyinstaller-runtimehooks/pathopthubble.py hubble.py \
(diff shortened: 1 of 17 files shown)


pkg/dev/debian9/Dockerfile
```diff
@@ -143,8 +143,8 @@ RUN apt-get install -y ruby ruby-dev rubygems gcc make \
 #use the following variables to choose the version of hubble
 ARG HUBBLE_CHECKOUT=develop
 ARG HUBBLE_GIT_URL=https://github.com/hubblestack/hubble.git
-ENV HUBBLE_VERSION=2.4.7_develop
-ENV HUBBLE_ITERATION=1
+ENV HUBBLE_VERSION=2.4.7
+ENV HUBBLE_ITERATION=50501
```
yassingh: I have prefixed the iteration number with "5050" in "pkg/dev" Dockerfiles ( since it looks like SOS; ideally one should not be running such builds on their prod envs )
I have assumed that iteration number will not go over 10.

pkg/debian9/Dockerfile
```diff
@@ -152,12 +152,15 @@ ENV LD_LIBRARY_PATH=/opt/hubble/lib:/lib:/lib64:/usr/lib:/usr/lib64:/usr/local/l
 RUN git clone "$HUBBLE_GIT_URL" "$HUBBLE_SRC_PATH" \
  && cd "$HUBBLE_SRC_PATH" \
  && git checkout "$HUBBLE_CHECKOUT" \
- && cp -rf "$HUBBLE_SRC_PATH" /hubble_build
+ && cp -rf "$HUBBLE_SRC_PATH" /hubble_build \
```
yassingh: Values of 'branch' and 'last_commit' are being updated here.
Since we don't need the commit when we are using a tag, I have used 'TAGGED_BUILD' string here.

basepi:
@yassingh Any ETA on when you can make these changes across all of the dockerfiles?

yassingh:
@basepi 
I was just waiting for @fossam  to give green signal to these few changes first. Once he does that, I should be able to make this change ( and test it ) for all Dockerfiles in one or two days.

( I have just pinged Andres for review in case he missed the notification )

yassingh:
As @fossam  correctly pointed out, version name in debian is creating problem only because of underscore. Things are working fine if we use hyphen in place on underscore. I have tested this thing.

But still, I think that the other changes related to --buildinfo will make our build cleaner. Let me know if they are to be reverted. 

I have tested all sorts of changes on debian9. Will be done with the testing of other OS by tomorrow.

basepi:
Yes, I think these changes are good even if we could have gotten to a working debian package just by dropping the underscore. I like the idea of being able to know what commit a build came from.

pkg/windows/dockerfile
```diff
@@ -32,7 +32,8 @@ COPY pyinstaller-requirements.txt c:/temp/
 COPY hubble.conf C:/temp/
 #install Chocolatey, then git and git.portable, and osquery
 RUN powershell.exe -Command Set-ExecutionPolicy Bypass -Scope Process -Force; iex ((New-Object System.Net.WebClient).DownloadString("$env:CHOCO_URL"))
-RUN powershell.exe -Command choco install git git.portable osquery nssm -y;
+RUN powershell.exe -Command choco install git osquery nssm -y;
+RUN powershell.exe -Command choco install git.portable --version 2.19.0 -y;
```
yassingh: There is some problem with the latest version of portable git.
It is not able to fetch profiles from git.
The older version is working fine.
Anyways, it's a good idea to keep our dependencies pinned to a specific version so that we get identical build every time we run docker.

yassingh:
@basepi 
I have done my testing.
It is ready to be merged from my side.

But "buildinfo" command will still face some problem because this issue is not completely resolved:
https://github.com/hubblestack/hubble/issues/464
( I just realized this )

The bug is still there for --buildinfo.

yassingh:
made the appropriate changes. 
It's good to go from my side.

PR #528, opened 2019-01-15
Unittests for readfile

Tenebriso:
Add tests for the rest of the functions in readfile.py


basepi:
We can never have too many tests! Thank you!

PR #537, opened 2019-01-29
Fix the hubblestack/cent7dev for caching

basepi:
Originally I was just going to turn off caching for the build altogether
and just do a full clone every time. But that seems a little silly. So
I added back in the ADD command even though we don't use the result. I
figure it's better to download the develop archive than to have to
install osquery and the other deps every time.

I think with these changes I can re-enable caching on the docker hub
build which will save time and resources overall even though we have the
wasted ADD command.

@mew1033 Any thoughts?


mew1033:
I like it. Leaving in the unused ADD is a good solution.

PR #543, opened 2019-02-13
Enhancing masking feature and adding build metadata logging

praksinha:
Following changes are present in this PR:

- There was a requirement from few teams in Adobe to enable masking based on a environment variable instead of a centralised list. Let's say if there is an environment variable '__hubble_mask__' set with value as comma separated blacklisted patterns, then masking will be performed based on value of this env variable instead of blacklisted patterns specified in mask.yaml. If '__hubble_mask__' is not set in environment, patterns specified in mask.yaml will be used for masking. This variable can be specified using 'custom_mask_key' in mask.yaml.
- Added a function that can be scheduled and will be responsible for logging versions of various modules of hubble and also buildinfo, if it is set. (similar to --buildinfo param usage)
- Few bug fixes


praksinha:
> The security team has requested that this masking be modified so that it will refuse to mask the `custom_mask_key` itself. This way we ensure that we have visibility into the masking configuration from splunk in case of issues.
> 
> Also, once we get this merged, can you do a `git fetch upstream && git reset --hard upstream/develop` on your develop branch? Your last couple of PRs have included hundreds of unrelated commits due to a messy git history. I want to make sure future PRs don't follow this pattern.

Made changes to not mask 'custom_mask_key' environment variable. Will do reset --hard once this PR is merged.

PR #545, opened 2019-02-14
add readfile_string to readfile.py

Tenebriso:


hubblestack/extmods/fdg/readfile.py
```diff
@@ -326,3 +328,38 @@ def _process_line(line, dictsep, valsep, subsep):
...
+        return False, None
+    try:
+        with open(path, 'r') as input_file:
+            ret = input_file.read()
+    except Exception as exc:
+        log.error('Error reading file {0}: {1}'.format(path, exc))
+        return False, None
+    status = bool(ret)
+    if encode_b64:
+        status, ret = encode_base64(ret)
```
basepi: Can you turn off format_chained in this call? Just in case there are any curly braces in the file.

Tenebriso: Nice catch! Thank you

PR #548, opened 2019-02-18
Fix vulners scanner

brannondorsey:
The current version of `vulners_scanner.py` fails silently as it seems that Vulners has updated its API and [no longer returns results wrapped in a `data` object](https://github.com/hubblestack/hubble/compare/develop...brannondorsey:vulners-scanner#diff-34467bb551bbb10b7a509ddbbce6179dL117). The existing behavior will silently fail, reporting a false negative for all scans. In other words, it will report that ALL packages are secure every time (e.g. `'Secure packages': '462 out of 462'`). This PR fixes this bug and now correctly reports the number of vulnerable packages installed on the machine being audited.

Additionally, the existing version of this scanner formats all packages [as if they are `rpm` packages](https://github.com/hubblestack/hubble/compare/develop...brannondorsey:vulners-scanner#diff-34467bb551bbb10b7a509ddbbce6179dL71) This PR updates the package format to also support Debian packages (presently Debian packages are also failing silently by saying all packages are secure).

### DIFF
--- a/hubblestack/files/hubblestack_nova/vulners_scanner.py
+++ b/hubblestack/files/hubblestack_nova/vulners_scanner.py
@@ -37,7 +37,7 @@ def audit(data_list, tags, labels, debug=False, **kwargs):
     os_version = __grains__.get('osmajorrelease')
 
     if debug:
-        log.debug("os_version: {0}, os_name{1}".format(os_version, os_name))
+        log.debug("os_version: {0}, os_name: {1}".format(os_version, os_name))
 
     ret = {'Success': [], 'Failure': [], 'Controlled': []}
 
@@ -48,8 +48,8 @@ def audit(data_list, tags, labels, debug=False, **kwargs):
             vulners_data = _vulners_query(local_packages, os=os_name, version=os_version, api_key=data['vulners_api_key'])
             if 'result' in vulners_data and vulners_data['result'] == 'ERROR':
                 log.error(vulners_data['data']['error'])
-            vulners_data = _process_vulners(_vulners_query(local_packages, os=os_name, version=os_version))
 
+            vulners_data = _process_vulners(vulners_data)
             total_packages = len(local_packages)
             secure_packages = total_packages - len(vulners_data)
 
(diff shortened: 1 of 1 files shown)


hubblestack/files/hubblestack_nova/vulners_scanner.py
```diff
@@ -48,8 +48,8 @@ def audit(data_list, tags, labels, debug=False, **kwargs):
             vulners_data = _vulners_query(local_packages, os=os_name, version=os_version, api_key=data['vulners_api_key'])
             if 'result' in vulners_data and vulners_data['result'] == 'ERROR':
                 log.error(vulners_data['data']['error'])
-            vulners_data = _process_vulners(_vulners_query(local_packages, os=os_name, version=os_version))
```
brannondorsey: This looks like it was wrong all along. `_vulners_query()` is run twice, back to back, the second time failing to provide the `api_key`. Instead we'll cache the first results saved to `vulners_data` and use them again.

hubblestack/files/hubblestack_nova/vulners_scanner.py
```diff
@@ -114,11 +133,11 @@ def _process_vulners(vulners):
     :return: A list of dictionaries as hubble.py swallows
     '''
 
-    packages = vulners.get('data', {}).get('packages')
+    packages = vulners.get('packages')
     if not packages:
         return []
 
-    return [{'tag': pkg,
+    return [{'tag': 'Vulnerable package: {0}'.format(pkg),
```
brannondorsey: The existing implementation reports secure packages like `'Secure packages': '462 out of 462'` but reports vulnerable packages simply using their package name  `{ 'firefox-60.2.2-1.el7.centos.x86_64': u'CESA-2019:0219, CESA-2018:3005, CESA-2018:3833'}`. 

This change aligns the two formats so that you can identify that the failure is because of a vulnerable package specifically (e.g. `{'Vulnerable package: firefox-60.2.2-1.el7.centos.x86_64': u'CESA-2019:0219, CESA-2018:3005, CESA-2018:3833'}`). In our setup, we are processing these failure reports downstream, so its very useful to include the string `"Vulnerable package"` so we can know _why_ the error was reported at all, instead of just having a package name.

hubblestack/files/hubblestack_nova/vulners_scanner.py
```diff
@@ -68,7 +68,26 @@ def _get_local_packages():
     '''
 
     local_packages = __salt__['pkg.list_pkgs']()
-    return ['{0}-{1}'.format(pkg, local_packages[pkg]) for pkg in local_packages]
```
brannondorsey: The existing implementation assumes packages are in RedHat format only. The new implementation uses salt grains to determine the `os_family` before deciding how packages should be formatted.

brannondorsey:
Thanks for the quick merge! We're using this module on some of our hubble-enrolled machines, so I'll let you know how it goes. I'll be sure to direct useful changes upstream if we have any more. Basic testing through Jenkins would be rad too!

basepi:
Yeah right now I have @jettero setting up linting and his test suite -- then we'll be tackling adding a lot more tests.

PR #558, opened 2019-03-06
Enhancing nebula to generate osquery conf from yaml files

praksinha:
Following changes are made:

- Added support to dynamically generate osquery conf in JSON format using topfile. One can now create host specific osquery conf using grains
- Added support to dynamically generate osquery flagfile using topfile. One can now create host specific flagfile using grains. This can be especially helpful if we want to enable/disable certain setting on certains host(s) or enable/disable extensions on few hosts.
- Added buildinfo in grains

### DIFF
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -294,10 +294,10 @@ def osqueryd_monitor(configfile=None,
     '''
     log.info("Starting osqueryd monitor")
     saltenv = __salt__['config.get']('hubblestack:nova:saltenv', 'base')
-    osqueryd_path = 'salt://osquery'
+    osqueryd_path = 'salt://hubblestack_nebula_v2'
     cached = __salt__['cp.cache_dir'](osqueryd_path, saltenv=saltenv)
-    log.info('Cached osqueryd files to cachedir')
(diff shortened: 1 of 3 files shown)


hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -615,6 +607,129 @@ def _get_top_data(topfile):
...
+                for key in flags_data:
+                    propdata = "--" + key + "=" + flags_data.get(key) + "\n"
+                    cf.write(propdata)
+        except Exception as e:
+            log.error("Failed to generate osquery flags file using topfile {0}".format(e))
+    
+    return flagfile
+
+
+def _get_osquery_top_data(topfile):
```
basepi: Can we re-use (might need slight modifications) the existing `_get_top_data()` function? Seems like the topfiles are similar enough we can make it work without so much duplication.

praksinha: Done.

praksinha:
> This looks good to me. Are we waiting on any other changes/approvals?

I'm just testing this code on few supported platforms. This should be completed by EOD today. Once done, we can merge this.

praksinha:
> This looks good to me. Are we waiting on any other changes/approvals?

Testing is complete on both linux and Windows. If there are no further review comments, we can merge this PR.

PR #559, opened 2019-03-06
More counter fields

jettero:
This seems to do the job @fossam was asking for (adding std host fields).

I attempted to split out the generation of the std host fields from splunk_pulsar_return; with the intention of literally standardizing those fields generation between the various splunk returners. (not totally sure that's a great idea, comments welcome).

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -29,6 +29,7 @@ import salt.utils.gitfs
 import salt.log.setup
 import hubblestack.splunklogging
 import hubblestack.hec.opt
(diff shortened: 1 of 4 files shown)


jettero:
I apparently branched from v3.0 ... rebased to develop

basepi:
Do we want to use this new import on the existing splunk returners which return similar standard information?

jettero:
That was my idea for sure. I'd like to DRY the whole concept of required host-id fields — at least partially so I don't miss it again, but also so it's the same for all the returners and data gatherers.

PR #562, opened 2019-03-11
should be TERM?

jettero:
Actual question... Pretty sure this should be TERM, not KILL.


fossam:
I agree. At least on Linux there is no point to catch SIG_KILL or SIG_STOP.

PR #565, opened 2019-03-12
Add meta_ to all index time extracted fields

mew1033:
I think the idea is that, for anything listed in the config file under {splunk_index_extracted_fields}, we find the value in the event, then stick it in {fields} with {meta_} prepended to the name. @fossam, that's what we were thinking, right?

I think this does it, but @jettero, you might want to double check I didn't miss anything.

### DIFF
--- a/hubblestack/extmods/returners/splunk_fdg_return.py
+++ b/hubblestack/extmods/returners/splunk_fdg_return.py
@@ -183,7 +183,7 @@ def returner(ret):
                         fields = {}
                         for item in index_extracted_fields:
                             if item in payload['event'] and not isinstance(payload['event'][item], (list, dict, tuple)):
-                                fields[item] = str(payload['event'][item])
+                                fields["meta_%s" % item] = str(payload['event'][item])
                         if fields:
                             payload.update({'fields': fields})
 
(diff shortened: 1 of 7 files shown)


hubblestack/extmods/returners/splunk_fdg_return.py
```diff
@@ -183,7 +183,7 @@ def returner(ret):
                         fields = {}
                         for item in index_extracted_fields:
                             if item in payload['event'] and not isinstance(payload['event'][item], (list, dict, tuple)):
-                                fields[item] = str(payload['event'][item])
+                                fields["meta_%s" % item] = str(payload['event'][item])
```
jettero: Why isn't this `fields[item] = payload['event'].pop(item)` again? It had something to do with archiving or something. Seems silly to duplicate the data (with a different name), but I vaguely recall there was a reason.

jettero:
This is the same change in 3-5 places … feels like something we should DRY eventually. Doesn't it?

basepi:
@jettero This can probably be rolled into my request to extend https://github.com/hubblestack/hubble/pull/559 for more DRY.

hubblestack/extmods/returners/splunk_fdg_return.py
mew1033: We want it in both so that you can still search for the data in standard event searches. If it's only in meta, it doesn't appear in the event when you're looking through the data.

PR #566, opened 2019-03-13
this broken in PR 565

jettero:
... and from the looks of it, never worked right anyway


jettero:
I do think this is the right place to do this (rather than the same code in the 4 returners); but this is broken. Remove.

PR #567, opened 2019-03-15
Changes to consider global as well as dynamic list for masking

praksinha:
Changes to cater following requests:

- For masking, both global blacklist (specified in mask.yaml) and dynamic blacklist (specified via process’ environment variable) will be used at same time. There is no OR condition here, we’ll use both. Also this not configurable as of now.
- Hubble would check for ENV_VAR_TO_REDACT environment variable to generate per process backlisted patterns. Values specified in this env var can be comma delimited, space delimited or both.
Masked value of variable that is redacted can be controlled by mask.yaml. As of now this value is like ‘xxxxx’ but we can tweak it to ‘REDACTED’ as suggested by Shannon in another email thread.
- To improve performance, I’m generating a single blacklist and performing masking using this list. This single blacklist would be specific to each process and contains only unique values after merging global and local/dynamic blacklist. This means, if we 1000 patterns in global blacklist and 100 patterns in ENV_VAR_TO_REDACT and out of these 100 patterns 95 are already present in global blacklist, then generated blacklist for that process would have total of 1005 patterns and will not have any duplicates.

### DIFF
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -51,6 +51,7 @@ hubble_status = HubbleStatus(__name__, 'top', 'queries', 'osqueryd_monitor', 'os
 
 __virtualname__ = 'nebula'
 __RESULT_LOG_OFFSET__ = {}
+__CUSTOM_MASK_KEY__ = 'ENV_VARS_TO_REDACT'
 
 
 def __virtual__():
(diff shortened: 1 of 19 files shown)


hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -51,6 +51,7 @@
 
 __virtualname__ = 'nebula'
 __RESULT_LOG_OFFSET__ = {}
+__CUSTOM_MASK_KEY__ = 'ENV_VARS_TO_REDACT'
```
fossam: Would it be possible to avoid hard coding it in the code?

Importing it somehow from profile would allow to change it more flexibly.

PR #571, opened 2019-03-20
updated counters with buckets

jettero:
There's a new setting "bucket_len" for "hubble_status". It defaults to 3600. The idea is to drop the counters into clock-aligned buckets so they're easier to line up with splunk counts by time. If the periods are predictable, then all the accounting is that much easier.

I think I increased the tech-debt a bit as this is pretty hard to read now; but it's worth it as it'll be a zillion times easier to line up with splunk event counts in searches.

I'm still testing this. (Should I add myself as a reviewer?)

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -797,6 +797,7 @@ def refresh_grains(initial=False):
     hubblestack.splunklogging.__opts__ = __opts__
 
     hubblestack.status.__opts__ = __opts__
+    hubblestack.status.__salt__ = __salt__
     hubble_status.start_sigusr1_signal_handler()
 
     if not initial and __salt__['config.get']('splunklogging', False):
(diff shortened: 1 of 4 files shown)


jettero:
OK. I think I got the memory leaks in this new code. There's others... I was variously using pympler and mem_top and `ps v $(pgrep hubble)`. There's still random increases I can't explain (even using the mentioned tools); but I don't think they're in the counters.

jettero:
"jettero dismissed basepi’s stale review via 4c11c77 13 minutes ago"

That's a little harsh github. I don't think I did that.

basepi:
Right now I have it dismissing reviews when new code is pushed -- since it hasn't been reviewed yet.

basepi:
Backported to 3.0 branch.

PR #575, opened 2019-03-25
Remove grains and more secrets from conf_publisher

basepi:
Also cleaned up some of the code


basepi:
Tested, working, and ready for review.

PR #576, opened 2019-03-26
Making global and local masking configurable

praksinha:
After discussion with team, made following changes:

- Introduced two new configuration flags in mask.yaml
  - enable_local_masking
  - enable_global_masking

Masking can be done in one of the following ways:
 - both local and global list to be considered if both the flags are set to 'true'
 - consider either local or global blacklist
- Use none of them (equivalent to disabling masking)


hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -810,6 +810,9 @@ def _mask_object(object_to_be_masked, topfile):
             column = blacklisted_object['column']  # Can be converted to list as well in future if need be
             # Name of column that stores environment variables
             custom_mask_column = blacklisted_object.get('custom_mask_column', '')
+            enable_local_masking = blacklisted_object.get('enable_local_masking', '').lower()
```
basepi: Can we please use the normal `None` default for these `.get()` calls? I'd like to use actual booleans where possible, since

```
if var is True:
```

is substantially faster than

```
if var == True:
```

and *way* faster than

```
if var == 'true':
```

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -820,7 +823,7 @@ def _mask_object(object_to_be_masked, topfile):
                         # This means data is generated by osquery interactive shell
                         for query_name, query_ret in r.iteritems():
                             for query_result in query_ret['data']:
-                                if custom_mask_column and custom_mask_column in query_result:
+                                if enable_local_masking == 'true' and custom_mask_column and custom_mask_column in query_result:
```
basepi: As with the comments above, let's use real `True` here.

```
if enable_local_masking is True
```

The rest of the instances should also be changed. We should never use `'true'`, it's much slower.

praksinha: Changes done and tested.

PR #584, opened 2019-03-29
Remove splunklogging.emit_data

basepi:
No need for these manual instances now that we have
hubblestack.log.emit_to_splunk

Note, these will now show up under different logger names in splunk --
rather than logging under the module name, they will log under the
new names `hubblestack.osquery_timing` and
`hubblestack.osquery_disk_usage`.

### DIFF
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -499,7 +499,7 @@ def check_disk_usage(path=None):
         if __salt__['config.get']('splunklogging', False):
             log.debug('Logging disk usage stats to splunk')
             stats = {'disk_stats': disk_stats, 'schedule_time': time.time()}
-            hubblestack.log.emit_to_splunk(stats, 'INFO', 'hubblestack.osquery_disk_usage')
+            hubblestack.log.emit_to_splunk(stats, 'INFO', 'hubblestack.disk_usage')
 
     return disk_stats
 
--- a/hubblestack/extmods/modules/nebula_osquery.py
+++ b/hubblestack/extmods/modules/nebula_osquery.py
@@ -42,7 +42,7 @@ import salt.utils.platform
 from hashlib import md5
 from salt.exceptions import CommandExecutionError
 from hubblestack import __version__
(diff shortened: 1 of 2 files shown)


basepi:
@praksinha You should probably review this as well -- since you haven't accepted the org invite I can't add you as a reviewer directly yet.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -502,12 +498,8 @@ def check_disk_usage(path=None):
 
         if __salt__['config.get']('splunklogging', False):
             log.debug('Logging disk usage stats to splunk')
-            hubblestack.splunklogging.__grains__ = __grains__
-            hubblestack.splunklogging.__salt__ = __salt__
-            hubblestack.splunklogging.__opts__ = __opts__
-            handler = hubblestack.splunklogging.SplunkHandler()
             stats = {'disk_stats': disk_stats, 'schedule_time': time.time()}
-            handler.emit_data(stats)
+            hubblestack.log.emit_to_splunk(stats, 'INFO', 'hubblestack.osquery_disk_usage')
```
fossam: In this case I would use a different name for the loggername, this data doesn't specifically report disk usage attributable only to osquery. Rather I think this is path specific.

basepi: So what would you like me to do? Technically you're correct, but it will only ever be called from osquery code (since it's in that module).

fossam: That is not explicit or obvious in back end logging system where these messages will end up and be consumed.

basepi: I disagree. Calling it `hubblestack.disk_usage` or something would be way less clear, since this is only used for osqueryd disk usage.

basepi: I'm open to alternatives, but you've given me none.

praksinha:
> @praksinha You should probably review this as well -- since you haven't accepted the org invite I can't add you as a reviewer directly yet.

Seems good to me.

PR #585, opened 2019-03-29
bump osquery version to 3.3.2 to dev

fossam:


basepi:
Should this be backported to 3.0 or do we want to test first?

PR #587, opened 2019-04-02
osquery Extensions Support

basepi:


basepi:
Still needs testing, so I opened this as a draft. Just wanted to get it up for high level review from @fossam.

basepi:
This is tested and working.

basepi:
Changes discussed above are in place and tested. This is ready for review.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -1484,3 +1495,168 @@ def query(query):
...
+        shutil.copyfile(local_path, dest)
+        ret = __salt__['file.check_perms'](name=local_path,
+                                           ret=None,
+                                           user=user,
+                                           group=group,
+                                           mode=mode)
+
+        return ret[0]['result']
+    except Exception as exc:
+        log.error('An error occurred getting file {0}: {1}'.format(path, exc))
```
fossam: Since most problems have been taken care with getting the file from cache would it make sense to have an error here about copying the file from `local_path` to `dest` (with the exception `exc`)?
We know that `local_path` and `dest` are very likely to exist as far as data values go, so the most likely sources for errors are issues with the action of copying the file. Having both the source and destination paths can be useful context in the error.

basepi: I don't think it's necessary -- purely because for any given salt:// path you can infer the local path easily by just prepending the cache dir. Additionally, it's theoretically possible that an exception could be thrown before `local_path` is defined so I'd prefer to keep it out of the error message.

basepi: @fossam Are we good to merge this?

PR #589, opened 2019-04-05
Fossam cloud details fix

fossam:
* get rid of gcp attributes, it can blow up the grains
* move extra details to their own try-except so we're guaranteed to get the limited details even if there's a problem with the return values of the extra details
* add azure extra details, prohibitive to correlate these details through other methods due to less than stellar api design.

### DIFF
--- a/hubblestack/extmods/grains/cloud_details.py
+++ b/hubblestack/extmods/grains/cloud_details.py
@@ -61,6 +61,10 @@ def _get_aws_details():
                                                           timeout=3).text
             aws_extra['cloud_private_hostname'] = requests.get('http://169.254.169.254/latest/meta-data/local-hostname',
                                                                timeout=3).text
+            for key in aws_extra.keys():
+                if not aws_extra[key]:
+                    aws_extra.pop(key)
+
         except (requests.exceptions.RequestException, ValueError):
             aws_extra = None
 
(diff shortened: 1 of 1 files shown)


fossam:
Tested on available aws, gcp, and azure instances.

hubblestack/extmods/grains/cloud_details.py
```diff
@@ -138,20 +148,47 @@ def _get_gcp_details():
...
+                gcp_extra[grain_name_mac] = value['mac']
+
+                grain_name_forwardedips = "cloud_interface_{0}_forwarded_ips".format(counter)
+                gcp_extra[grain_name_forwardedips] = ','.join(value['forwardedIps'])
+
+                grain_name_targetips = "cloud_interface_{0}_target_ips".format(counter)
+                gcp_extra[grain_name_targetips] = ','.join(value['targetInstanceIps'])
+
+                grain_name_accessconfig_external_ips = "cloud_interface_{0}_accessconfigs_external_ips".format(counter)
+                external_ips_list = [ dict['externalIp'] for dict in value['accessConfigs'] if 'externalIp' in dict ]
```
basepi: Can you please use a different variable name besides `dict`? I don't like shadowing built-ins.

PR #590, opened 2019-04-08
Fix typos in fdg module

mew1033:
Pretty sure these are typos


basepi:
Definitely. Good catches.

PR #593, opened 2019-04-10
Update/Fix fdg pieces

mew1033:
This PR does a few things:

1. FDG modules return a status along with the result. This status wasn't passed up as part of the overall fdg return, now it is. I also added an option to the fdg<-->nova connector that allows you to use this status instead of the real returned value.
2. I added the ability to pass in custom headers in the curl module, as well as a toggle for attempting to decode json from the result.
3. The splunk_fdg_return didn't specify a default sourcetype. I added it as `hubble_fdg`.
4. Fixed a small bug in xpipe. It was bassing in the full chained value instead of the iteration each time.

Comments welcome

### DIFF
--- a/hubblestack/extmods/fdg/curl.py
+++ b/hubblestack/extmods/fdg/curl.py
@@ -31,7 +31,8 @@ def request(url,
             password=None,
             timeout=9,
             verify=None,
-            chained=None):
+            chained=None,
+            chained_status=None):
     '''
     Given a series of arguments, make a request using ``requests``.
 
--- a/hubblestack/extmods/fdg/curl.py
+++ b/hubblestack/extmods/fdg/curl.py
@@ -27,10 +27,12 @@ def request(url,
             function='GET',
             params=None,
             data=None,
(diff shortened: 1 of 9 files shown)


mew1033:
The only place I actually needed it was in the curl module. If I curl against a host that doesn't exist, I don't know exactly what error is going to come back. The error is then converted to a string and returned as the result. When cast to a bool, you can't tell whether or not the request was successful. I figured that since the curl module was reporting both a true/false status as well as a result, that I could just use the status in my Nova check. But that meant I needed to return the status from the fdg module. I had to add it in to all the other module functions because it was going to be passed in from the _fdg_execute function. Technically I could have just made sure they all had **kwargs at the end and that would have been fine. But you explicitly pass in `chained`, so I figured I'd pass in the status variable too.

tl;dr I don't USE it anywhere, but I want to report the status outside of the fdg module. Coercing the result to a boolean (as is done in the fdg<-->nova connector wouldn't work in my case.

PR #594, opened 2019-04-11
Persist version/buildinfo grains and add hubble_mark_3 grain

basepi:
This one is a gnarly one. Since the hubble_version and buildinfo are on disk, and grains refresh *from* disk, an old version of the daemon could pick up changes to those thing on disk (perhaps from a new package install) and if you're using those grains as reporting tools, you lose insight into what is actually running on the host.

This PR makes those grains persist automatically across refreshes. But additionally it adds a new hardcoded grain `hubble_marker_3` so you can tell if a daemon is actually running the new persisting code or not. It's a hacky solution but I don't have a better one in mind.

Thoughts?

(Note, this is targeted directly at 3.0 and will hit develop when I merge forward)

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -583,9 +583,11 @@ def load_config():
     disable_modules.extend([
         'boto3_elasticache',
         'boto3_route53',
+        'boto3_sns',
         'boto_apigateway',
         'boto_asg',
         'boto_cfn',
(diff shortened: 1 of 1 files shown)


basepi:
I also threw another small fix that we noticed in new builds -- just updated the list of blacklisted boto modules from salt.

PR #595, opened 2019-04-11
Add support for account_contains in win_secedit.py

mew1033:
This will allow using the user/group name in checks where a user or group must be present in a list, but other items are allowed in the list.


basepi:
@mew1033 Is this a bugfix? Should it be backported or target 3.1?

mew1033:
@basepi Can we backport this to 3.0?

basepi:
Done.

PR #596, opened 2019-04-12
Add flag in nova profile for domain controller hosts

mew1033:
I think I did this right, but it's not tested yet.


basepi:
Yeah I have no qualification to review this PR, so I'll bow to @madchills' expertise.

mew1033:
Tested and it appears to work. @madchills ?

PR #600, opened 2019-04-15
random name resolver — is this janky or cool?

jettero:
I don't think this is the right way to do this; but it's sorta the right idea.

I ran into a problem in my home lab where the system resolver wasn't working (I have no idea why, it works for everything but hubble) — I added this as an experiment.

DNS Round Robin is a really common splunk load balancing technique. We should probably convert this section of code to loop over all the A and AAAA records and try each one before considering the name something.hec.wherever to have failed. Possibly this should be in a random order; regardless, this commit could be converted to whatever strategy is decided to be appropriate… probably.

Marked do-no-merge since I'm basically just thinking out loud.

### DIFF
--- a/hubblestack/hec/obj.py
+++ b/hubblestack/hec/obj.py
@@ -6,6 +6,7 @@ import time
 import copy
 import os
 import hashlib
+import random
 
 import certifi
 import urllib3
@@ -285,8 +286,18 @@ class HEC(object):
         possible_queue = False
         for server in sorted(servers, key=lambda u: u.fails):
             log.debug('trying to send %d octets to %s', len(data), server.uri)
+            the_uri = server.uri
             try:
-                r = self.pool_manager.request('POST', server.uri, body=data, headers=self.headers)
+                hname = urllib3.util.parse_url(server.uri).host
+                hip   = socket.gethostbyname_ex(hname)[2]
+                if hip:
+                    the_uri = server.uri.replace(hname, random.choice(hip))
+            except socket.gaierror:
+                pass
+            if server.uri != the_uri:
+                log.debug('rewrote %s as %s using random name resolution', server.uri, the_uri)
+            try:
+                r = self.pool_manager.request('POST', the_uri, body=data, headers=self.headers)
                 server.fails = 0
             except urllib3.exceptions.LocationParseError as e:
                 log.error('server uri parse error "%s": %s', server.uri, e)


basepi:
It's an interesting idea for sure. Can we not rely on the round robin DNS to effectively do this for us?

jettero:
I would have thought so. I suspect that python (or one of the libs under urllib3) tries to cache the DNS reply though.

jettero:
It took some messing around to get right; but my testing shows that iff you tell the pool to `clear()`; you get roughly the random ordering in my patch here. If you don't tell it to `clear()` then it persists (since the pool asks it to keep the connection open) and you get the same IP every time.

Note, in particular, that we don't exactly hit all the RR exhaustively before looping again in a new random order. 1 4 2 2 3 1 1 1 

```
In [38]: import urllib3
    ...: http = urllib3.PoolManager()
    ...: def gd():
    ...:     res = http.request('GET', 'http://cip/ip')
    ...:     http.clear()
    ...:     return res.data.strip().split()[-1].strip('.')
    ...: [ gd() for _ in range(10) ]
Out[38]: 
['10.1.1.1',
 '10.1.1.4',
 '10.1.1.2',
 '10.1.1.2',
 '10.1.1.3',
 '10.1.1.1',
 '10.1.1.1',
 '10.1.1.1',
 '10.1.1.3',
 '10.1.1.4']

```

It's going to be harder to test the `urllib3.util.retry.Retry` object, but I wonder if it can be taught to try all the RR before returning the failure. We actually use a Retry in the hec.obj; but it may as well not be there imo.

jettero:
After much playing around (see below), I've determined that this cannot be teased out of the `Retry` object even with rather extreme messing around. If the pool isn't told to `clear()` then it retries on the same connection -- which doesn't get us anywhere. If you trick it into clearing, urllib3 crashes because the connection isn't open.

```
In [22]: import logging
    ...: import urllib3
    ...: log = logging.getLogger('supz')
    ...: class MyRetry(urllib3.util.retry.Retry):
    ...:     def increment(self, *a, **kw):
    ...:         pm.clear()
    ...:         return super(MyRetry, self).increment(*a, **kw)
    ...: retries = MyRetry(total=10, status=10, status_forcelist=(404,))
    ...: pm = urllib3.PoolManager(retries=retries)
    ...: try:
    ...:     res  = pm.request('GET', 'http://cip/ip/nofile')
    ...: except urllib3.exceptions.MaxRetryError:
    ...:     pass
ClosedPoolError: HTTPConnectionPool(host='cip', port=80): Pool is closed.
```

basepi:
But the caching isn't really a problem, is it? Round robin is supposed to load balance across many hosts, not for multiple requests from the same host.

jettero:
I'm thinking about failure logic. If your endpoint is something.hec-region.whatever.com and it is itself an RR of 30 IPs; and pretend 40% of them are "down" for whatever reason.

Trying the single host for a given event should involve trying an alternate IP if the hec.obj encounters a 5xx error. (imo)

basepi:
Ah, not a bad idea for failure conditions.

basepi:
I'm going to close this for now -- it's not a bad idea but we need to find time to talk it out, and for now with the load-balancer handling the HEC endpoint we don't need this internally.

PR #602, opened 2019-04-16
Handling corner cases and improving logging

praksinha:


hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -1267,14 +1267,18 @@ def _start_osqueryd(pidfile,
     '''
     log.info("osqueryd is not running, attempting to start osqueryd")
     if salt.utils.platform.is_windows():
-        log.error("requesting service manager to start osqueryD")
+        log.warn("requesting service manager to start osqueryd")
```
basepi: Should this be `warn` or just `info`? Do we expect osqueryd to already be running via the service manager? Or does hubble always start osqueryd on windows?

praksinha: In ideal case case, this statement would be executed only the first time monitor function is called. In subsequent executions osqueryd service should be in running state already and control should not reach this statement. If it does, I thought of logging it as a warning. I can change this to INFO type as in most of the cases it would be executed only once and in that case INFO type would be more appropriate. Let me know if I should change that.

basepi: Let’s do info so that users are not concerned by a normal starting of the process the first time. If, in the future, we want to track whether it’s the first time the process was started, we can introduce a global variable.

praksinha: Fixed

PR #603, opened 2019-04-22
Salt log leak fix

jettero:
This seems to fix most of the memory problems with Hubble.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -178,10 +178,26 @@ def main():
 
         try:
             log.debug('Executing schedule')
-            schedule()
+            sf_count = schedule()
+            if sf_count > 0:
+                log.debug('Executed %d schedule item(s)', sf_count)
+                workaround_salt_log_handler_queues()
         except Exception as e:
             log.exception('Error executing schedule')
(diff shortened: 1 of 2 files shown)


basepi:
Since pulsar runs every second, this will run every second. Is that acceptable or should it be less frequent?

jettero:
I was wondering about that too. It seems fine. It deletes 4 or 8 records every sweep, sometimes a hundred after a nova or nebula... I can't see the downside of running it every time something fires.

hubblestack/daemon.py
```diff
@@ -178,11 +178,27 @@ def main():
...
-            schedule()
+            sf_count = schedule()
+            if sf_count > 0:
+                log.debug('Executed %d schedule item(s)', sf_count)
+                workaround_salt_log_handler_queues()
         except Exception as e:
             log.exception('Error executing schedule')
         time.sleep(__opts__.get('scheduler_sleep_frequency', 0.5))
 
+def workaround_salt_log_handler_queues():
```
basepi: Would you mind moving this function to [log.py](https://github.com/hubblestack/hubble/blob/develop/hubblestack/log.py)? I want to keep every bit of complexity out of daemon.py where it makes sense.

PR #604, opened 2019-04-23
Fix spelling in misc.py

mew1033:


hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -1021,6 +1021,12 @@ def _is_int(input):
         return False
     return True
 
+def check_sshd_parameters(*args, **kwargs):
+    '''
+    Fix spelling while also retaining backwards comparability
```
basepi: compatibility?

mew1033: ..... Oops.
Fixed.

PR #605, opened 2019-04-24
add timezone data to grains

Tenebriso:


hubblestack/extmods/returners/splunk_generic_return.py
```diff
@@ -62,6 +98,8 @@ def returner(retdata):
         idx = opts.get('index')
 
         for event in events:
+            event.update({'timezone': __grains__['timezone']})
+            event.update({'timezone_hours_offset': __grains__['timezone_hours_offset']})
```
basepi: Instead of adding this here (which will require it to be added manually to all of the returners) can you please add it to [stdrec.py](https://github.com/hubblestack/hubble/blob/develop/hubblestack/utils/stdrec.py) (inside of `std_info`) instead? I want to consolidate all of these always-included fields there so we don't have to have so much repeated code.

hubblestack/utils/stdrec.py
```diff
@@ -45,6 +45,8 @@ def std_info():
         'dest_ip': fqdn_ip4,
         'dest_fqdn': local_fqdn,
         'system_uuid': __grains__.get('system_uuid'),
+        'timezone': __grains__.get('timezone'),
```
fossam: Can we remove the these new fields so they won't be sent in with _every_ event. Since these are in grains they will be sent in at configurable intervals that are normally adequately frequent.

In addition, there is a naming collision with an existing timezone grain. The existing timezone grain for some reason is not populated every time, but it does exist occasionally.

basepi: Ah, good point. We don't need to put the new grains anywhere in the splunk returners since grains are already reported hourly.

Let's call the timezone `timezone_short` since it's the abbreviation anyway.

PR #606, opened 2019-05-01
add logs to queue conditions

jettero:
This seems to solve the problem of logging queue conditions. It's pretty janky though. For very good reasons, the hubblestack.hec.obj log messages are filtered from splunklogging.

Possible problems with this patch include:

* this doesn't check the status of the hubble logging options (should it??). I was about to add that and thought: these aren't typical log messages. They're events. Queueing to disk should be rare and it doesn't seem to be in the environments where this is needed.

* this doesn't make any attempt to measure remaining queue-space (remaining capacity); should it? Should it report when the queue is full? How would it do this?

### DIFF
--- a/hubblestack/hec/obj.py
+++ b/hubblestack/hec/obj.py
@@ -17,6 +17,7 @@ import hubblestack.status
 hubble_status = hubblestack.status.HubbleStatus(__name__)
 
 from . dq import DiskQueue, NoQueue, QueueCapacityError
+from hubblestack.utils.stdrec import update_payload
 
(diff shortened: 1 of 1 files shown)


basepi:
Do we need @fossam to review or should we just merge?

jettero:
@fossam  doesn't have to actually click the review buttons (imo); but I do want to check with him about what he needs to see to make it good. I think I still need to add a couple things.

basepi:
I'll hold off until I get @fossam's ok.

PR #609, opened 2019-05-06
production log bump

jettero:
Move a few logging messages up to ERROR level for production logging.

### DIFF
--- a/hubblestack/hec/obj.py
+++ b/hubblestack/hec/obj.py
@@ -228,7 +228,7 @@ class HEC(object):
             self._direct_send_msg('queue(start)')
         p = str(payload)
         # should be at info level; error for production logging:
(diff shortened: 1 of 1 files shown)


hubblestack/hec/obj.py
```diff
@@ -227,11 +227,13 @@ def _queue_event(self, payload):
         if self.queue.cn < 1:
             self._direct_send_msg('queue(start)')
         p = str(payload)
-        log.info('queueing %d octets to disk', len(p))
+        # should be at info level; error for production logging:
+        log.error('queueing %d octets to disk', len(p))
```
basepi: Could you make this more user-friendly? Perhaps something along the lines of

```
Sending to splunk failed. Queueing %d octets to disk
```

Since we're making these errors bubble up it's worth making it more obvious what's going on.

PR #610, opened 2019-05-06
Push large objects to bottom of json in Splunk returns

mew1033:
When you send in json events into Splunk, they are stored as raw strings, and the json is extracted at search time. This is great, all is well. However, Splunk will just stop extracting fields at 5k characters. This means that if the json event is stupid long, fields that fall after that point won't be extracted and _can't be searched on_.

The main thing this causes problems for is our identity fields: `system_uuid`, `cloud_instance_id`, etc. If they happen to fall at the bottom of a long event and don't get extracted, they won't be returned in a search based on that field.

This solution finds events that are > 5000 characters, finds the long ("long" is here defined as 500 characters) objects and pushes them to the bottom of the json object.

### DIFF
--- a/hubblestack/hec/obj.py
+++ b/hubblestack/hec/obj.py
@@ -89,7 +89,23 @@ class Payload(object):
         self.sourcetype = dat.get('sourcetype', 'hubble')
         self.time       = dat.get('time', now)
 
-        self.dat = json.dumps(dat)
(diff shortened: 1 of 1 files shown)


fossam:
Let's make sure Splunk agrees with these changes.

mew1033:
Looks like we found a better way of doing this.

PR #611, opened 2019-05-07
Removing unused directories from package and adding rpm spec metadata

praksinha:
Following are fixed in this:

- There were directories and files getting created and packaged with hubble installer that were not used by Hubble. Example: we include /etc/osquery/osquery.conf and /etc/osquery/osquery.flags file but never use it via hubble. Similar we create directories /var/log/osquery which is never used by Hubble. Removed all such directories.

- Added few RPM metadata fields like URL, Description and Summary. This would improve YUM repository support for rpm packages.

### DIFF
--- a/pkg/amazonlinux2016.09/Dockerfile
+++ b/pkg/amazonlinux2016.09/Dockerfile
@@ -187,7 +187,7 @@ CMD [ "if [ -f /data/hubble_buildinfo ] ; then echo \"\" >> /hubble_build/hubble
        --after-install /hubble_build/conf/afterinstall.sh \
        --after-upgrade /hubble_build/conf/afterupgrade.sh \
        --before-remove /hubble_build/conf/beforeremove.sh \
-       etc opt usr \
+       etc opt usr /var/log/hubble_osquery/backuplogs \
 #edit to change iteration number, if necessary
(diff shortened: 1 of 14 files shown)


pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -180,11 +181,14 @@ CMD [ "if [ -f /data/hubble_buildinfo ] ; then echo \"\" >> /hubble_build/hubble
        -n hubblestack \
        -v ${HUBBLE_VERSION} \
        --iteration ${HUBBLE_ITERATION} \
-       --config-files /etc/osquery/osquery.conf \
+       --url ${HUBBLE_URL} \
+       --description \"${HUBBLE_DESCRIPTION}\" \
+       --rpm-summary \"${HUBBLE_SUMMARY}\" \
+       --config-files /etc/hubble/hubble \
```
basepi: This used to be here but I explicitly removed it. The idea is that we have our users put their config in `/etc/hubble/hubble.d/hubble.conf` so that we can update `/etc/hubble/hubble` properly. If you add it as a `--config-files` entry, then if the user touches it, yum will stop updating it.

Can you please remove this from all of the dockerfiles?

praksinha: Sure, didn't know that. Fixed.

PR #612, opened 2019-05-13
attempt to fully fix dup hec summary messages

jettero:
Fixing the dup hec summary messages has the unintneded side effect of ensuring no repeat messages are ever submitted. So this change also submits intentional repeats (adjustable via `hubble_status:summary_repeat` in `__opts__`).


jettero:
This should be good to go.

PR #617, opened 2019-05-14
Disk queue set top of file

jettero:
look for `disk_queue*` at the top of config. It seems to work, and should be backwards compatible with other differing settings in the per-hec settings area(s)

### DIFF
--- a/hubblestack/hec/opt.py
+++ b/hubblestack/hec/opt.py
@@ -16,6 +16,11 @@
 #   list_of_opts = get_splunk_options()
 #
 # we just look in [config.get]('hubblestack:returner:splunk')
+#
+# Additionally, the defaults for disk_queue, disk_queue_size and
+# disk_queue_compression can be set in the top level configuration -- although,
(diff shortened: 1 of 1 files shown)


hubblestack/hec/opt.py
```diff
@@ -42,9 +49,10 @@ def _get_splunk_options(space, modality, **kw):
         'index_extracted_fields': [],
         'http_event_collector_ssl_verify': True,
         'add_query_to_sourcetype': True,
-        'disk_queue': False, # '/var/cache/hubble/dq',
-        'disk_queue_size': 10 * (1024 ** 2),
-        'disk_queue_compression': 5,
+        # disk_queue* can come from the top of the config
+        'disk_queue': confg('disk_queue', False),
+        'disk_queue_size': confg('disk_queue_size', 10 * (1024 ** 2)),
```
fossam: Since the feature is disabled by default, can we make the default size setting be 100MB anyway?

PR #620, opened 2019-05-15
Provision for building osquery with custom local code

yassingh:
#CAN BE MERGED
I am changing Dockerfiles so that one can build osquery from custom code as well.
This behavior is governed by the docker build argument "OSQUERY_BUILD_ENV"

By default, its value is "remote". In this case, docker will bring the osquery code from the GIT url specified in the Dockerfile.

If its value is "local", then docker will try to copy "osquery" folder into the image. So you should place your osquery code in a folder name "osquery" and this folder should be located in the same directory that contains Dockerfile.
On using "--build-arg=OSQUERY_BUILD_ENV=local", osquery is built from the custom code and packed into the hubble build.

This change uses "multi-stage" feature of docker, so this dockerfile will run on docker 17.05 or later version.
Refer this :
https://docs.docker.com/develop/develop-images/multistage-build/

I have already tested the dockerfiles around my changes.
But still, I want this code to be merged only after I have created a successful build from all these Dockerfiles.
Will let you know after I complete my testing ( perhaps by tomorrow).

@basepi @fossam

### DIFF
--- a/pkg/dev/amazonlinux2016.09/Dockerfile
+++ b/pkg/dev/amazonlinux2016.09/Dockerfile
@@ -5,7 +5,35 @@
 # The resulting image is ready to run the pyinstaller on container start and drop hubble<version>-coreos.tar.gz
 # in the /data directory. Mount /data volume into a directory on the host to access the package.
 # To run the container:  docker run -it --rm -v `pwd`:/data <image_name>
+# Requires docker 17.05 or higher
 
+# Set this arguement to "local" if you want to build osquery for local code.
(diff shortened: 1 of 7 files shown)


yassingh:
I have tried these Dockerfiles. 
Seem to be working fine .
PR can be merged now.

@basepi 
I don't have permission to remove the label.
Please ignore and remove the label.

basepi:
Sorry about that permissions issue, @yassingh. You were in the org but in the wrong team in the org. You should have labeling access now.

PR #622, opened 2019-05-16
Using ltsc2016 tag in win dockerfile

yassingh:
This dockerfile does not work any more  because mircosoft has depercated the "latest" tag which we were using to pull windowservercore.

See the bottom part:
https://techcommunity.microsoft.com/t5/Containers/Windows-Server-2019-Now-Available/ba-p/382430

In other words, you can't make any new windows build using current Dockerfile ( unless you have already cached windowsserver in the past ).

@basepi 
Let me know if I am supposed to raise PR for develop branch also.


basepi:
This will merge forward to the develop branch, so a separate PR is unnecessary.

PR #624, opened 2019-05-16
Add Sumologic returners && fixed Graylog returner bug

zachsis:
Hello! 

I've tested these Sumologic returners and validated they all work. Also while putting these together, I noticed a bug in the graylog returners (`graylog_nebula_return.py` and `graylog_nova_return.py`) that was created likely with a find/replace that went wrong. 

Thanks!


zachsis:
CLA is signed now. thx!

basepi:
Awesome! Thanks for jumping through that extra hoop for us. Legal, amirite? ;)

PR #625, opened 2019-05-17
add time check module against NTP servers

Tenebriso:


fossam:
What do we do if the `ntpdate` command is not available? It is my experience many times it does not come default.

In some distros `ntpdate` may have its own configuration, separate from the normal `ntp` package settings. How would we handle this situation?

Finally, in what percentage of systems this approach would not work? More finally, what can be done so that either the coverage would be larger or it would be possible to only run this on hosts where it has a high chance of success.

Apologies for all the questions. My experience with `ntpdate` prompted these.

basepi:
I wonder if it would be worth trying to parse the ntp configuration directly? That way we don't have to add a dependency or rely on system tools.

That said, assuming `ntplib` doesn't pull in a bunch of dependencies of its own, I don't have a huge problem with adding a new dep.

basepi:
Syncing up my comments on slack to this PR: Let's go with `ntplib` and `w32tm` (for windows), assuming `w32tm` is available by default. Remember that the `ntplib` dep needs to be added to the requirements.txt file, setup.py, and all the pyinstaller-requirements files throughtout `pkg/`. (We still need to make that less terrible -- sorry!)

basepi:
Oh, also this is conflicting due to @jettero's changes in another PR. Can you rebase and fix the conflict?

test-requirements.txt
```diff
@@ -21,5 +21,6 @@ smmap2
 tornado
 urllib3
 vulners
+ntplib
 pytest-pythonpath
-pycrypto
+pycrypto
```
basepi: This is odd. I can't detect any change but it still shows a diff on this line. 🤔

PR #630, opened 2019-05-22
add default_index to hec.obj

jettero:
the mysterious default indexing ... is a null index field.

I think there's still a couple returners (including my own generic return) that don't use `make_hec_args()`; the other is fdg.

Arguably the hec.obj should check for missing index and set it to the new `self.default_index` when missing.

### DIFF
--- a/hubblestack/extmods/returners/splunk_generic_return.py
+++ b/hubblestack/extmods/returners/splunk_generic_return.py
@@ -53,6 +53,7 @@ def _build_hec(opts):
     http_event_collector_host = opts['indexer']
     http_event_collector_port = opts['port']
     hec_ssl = opts['http_event_server_ssl']
+    http_event_index = opts['http_event_index']
     proxy = opts['proxy']
     timeout = opts['timeout']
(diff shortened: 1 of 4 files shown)


hubblestack/extmods/returners/splunk_generic_return.py
```diff
@@ -53,6 +53,7 @@ def _build_hec(opts):
     http_event_collector_host = opts['indexer']
     http_event_collector_port = opts['port']
     hec_ssl = opts['http_event_server_ssl']
+    idx     = opts['http_event_index']
```
basepi: s/idx/http_event_index/

hubblestack/hec/obj.py
```diff
@@ -131,6 +131,7 @@ def __str__(self):
 
 
     def __init__(self, token, http_event_server, host='', http_event_port='8088',
+                 http_event_index='default',
```
basepi: Let's move http_event_index next to token and don't give it a `default`. It should always be defined.

PR #635, opened 2019-05-23
Remove StartLimit directives in systemd unit files

basepi:
This change has two driving reasons. The first is that the StartLimit
directives have changed in recent versions of systemd. In some versions
it's expected in the `[Service]` section, in others in the `[Unit]`
section. Additionally, it's also been renamed to `StartLimitInvervalSec`
in recent versions of systemd.

Rather than deal with these compatibility issues, I propose that we
remove them. I have also bumped the `RestartSec` to 300. Basically, if
hubble ever crashes, we do want it to restart, but we're not in a hurry.
So I think using `RestartSec=300`, the impact of trying to restart
hubble forever is negligible, and we save ourselves the hassle of trying
to deal with systemd compatibility.

Thoughts? I think @praksinha added these in.

### DIFF
--- a/pkg/hubble.service
+++ b/pkg/hubble.service
@@ -2,15 +2,13 @@
 Description=Hubblestack
 Requires=network-online.target
 After=network-online.target
-StartLimitInterval=350
-StartLimitBurst=10
 
 [Service]
 Type=forking
 PIDFile=/var/run/hubble.pid
 ExecStart=/opt/hubble/hubble -d
 Restart=always
-RestartSec=30
+RestartSec=300
 MemoryAccounting=true
 
 [Install]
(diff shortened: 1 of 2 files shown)


praksinha:
Yeah I remember these directives not being compatible with older versions of systemd when I was adding them. We can remove them if them if they are creating too much noise on non compatible systems.

These are definitely useful but it seems they'll do more more harm then good in our use-case (assuming hubble wouldn't crash too often :) )

basepi:
Yeah honestly I'm not sure I've ever seen hubble crash unless there was an error in the config YAML. Most of the issues we've seen are stalls, and systemd isn't aware of those.

PR #637, opened 2019-05-27
Fixing osquery permissions for Windows server 2019

praksinha:
Following changes:
- Pinned osquery version to v3.3.2. This was required since facebook has released a new 3.4.0 version of osquery specifically for Windows.
- Fixed permissions of osquery for Windows server 2019. New permissions (explicit DENY) apply to all supported Windows platforms.


basepi:
👍

PR #639, opened 2019-05-30
Fix s3fs to use cached metadata and files during network outage

basepi:
Still needs to be tested, but I wanted @jettero's review before he left for the day.


basepi:
Working great, ready for review/merge.

PR #638, opened 2019-05-28
Provision of custom osquery build in Dockerfiles

yassingh:
Making changes in Dockerfiles similar to https://github.com/hubblestack/hubble/commit/890463de97766349977466e0e6b13109a9f3d5a1

After this change, one can build hubble with custom osquery code.

@basepi I have made these changes in just centos7 for now. Please answer these questions so that I can proceed to make these changes for all OS:

1. Is '3.0' the correct branch to make these changes ? I am assuming that the changes made here will propagate to 'master' branch Dockerfiles so that we can build hubble 3.0 using master Dockerfiles.

2. What should be the value of OSQUERY_SRC_VERSION ? ( see the comment in code for more details )

3. I have just focused on bringing the changes from 'dev' to 'pkg' dockerfiles related to the aforementioned commit for now. If there are any more changes which should copy from 'dev' to 'pkg' Dockerfiles in this PR, let me know.

PR #642, opened 2019-05-31
Renaming debian builds

yassingh:
Replacing underscore with dash to make debian build name similar to other os name.

I should have pushed this change in https://github.com/hubblestack/hubble/pull/638
But I caught this discrepancy in build names later.


fossam:
Debian didn't like underscores in their version strings (https://www.debian.org/doc/debian-policy/ch-controlfields.html#version). We had this in #527. But issue here is different because it's about package file names. Naming standard for this aspect is at https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_debian_package_file_names .

yassingh:
+1 with @fossam 
There was an issue because of using underscore in "version" string of debian packages.
I hope that @basepi is also referring to this issue only.

Moreover, even the current debian builds are named in the same way as I am proposing ( debian build names are also using dashes here https://github.com/hubblestack/hubble/releases/tag/v2.4.7 ). 

Earlier I used to rename the debain builds while uploading them.
With this PR, the debian builds will be renamed as soon as they are created inside the docker container.
So, this path change has already been tested by all of us.

PR #649, opened 2019-06-04
Revert "Avoiding last commit info in master builds "

basepi:
Reverts hubblestack/hubble#646

@yassingh After chatting with @fossam he thinks it's harder to work with the last_commit field in splunk when it's a special string like this.

I also did some testing, and if you're on a tag, `git describe` just returns the tag itself:

```
basepi@mustafar:~/src/work/hubble$ git checkout v3.0.1
Note: checking out 'v3.0.1'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 39c7491 Rev to v3.0.1
basepi@mustafar:~/src/work/hubble$ git describe
v3.0.1
```

I think we should leave the `git describe` in all cases. More info is better, and it's still very clean on the tag itself.

Thoughts?

### DIFF
--- a/pkg/amazonlinux2016.09/Dockerfile
+++ b/pkg/amazonlinux2016.09/Dockerfile
@@ -168,7 +168,7 @@ RUN git clone "$HUBBLE_GIT_URL" "$HUBBLE_SRC_PATH" \
  && git checkout "$HUBBLE_CHECKOUT" \
  && cp -rf "$HUBBLE_SRC_PATH" /hubble_build \
  && sed -i "s/BRANCH_NOT_SET/${HUBBLE_CHECKOUT}/g" /hubble_build/hubblestack/__init__.py \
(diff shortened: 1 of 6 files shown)


yassingh:
I have no problem with this. I was also somewhat inclined towards using 'git describe' as it adds more info. Please feel free to use 'git describe'

@basepi  is it possible to make this change is "debian9" Dockerfile as well in this PR ?? It would be nice to have similar behaviour for all linux OS. ( I changed debian9's Dockerfile in an older PR, so it is not showing up in this reverted PR ).

basepi:
I'll make a separate PR with the Debian 9 change, since this was generated by Github.

PR #651, opened 2019-06-06
Fix some formatting and organization issues

basepi:
This PR does three things:

* Copies the pyinstaller-requirements.txt file to the root requirements.txt file (not sure why they were mismatched)
* Fixes an instance of multiple imports on the same line
* Changes all single-quoted docstrings to double-quoted docstrings, to align with the recommendation in the docstring pip


fossam:
`pyinstaller-requirements.txt` files can't be symlinks. Docker does not provide that option.

Jenkins means that for those who check out the repo, docker build might fail right off the bat.

Separate pyinstaller requirements files were used to pin deps to different versions for different distros.

I am not saying it's a great system, just stating why things are the way they are. To really take building to the next level the whole build process may need to be redesigned to something like BuildKit. This is a bigger undertaking and probably not for now.

basepi:
Fixed the conflict. Please review at your leisure @jettero

basepi:
I also dislike the double quoted docstrings. But someone pointed out to me that the pep for docstrings actually does have a preference. So I'm going to hope I get used to it.

PR #654, opened 2019-06-10
Local CVE Scanner Concept for Hubble

buddwm:
A local CVE scanner that can be adapted for Hubble's Nova module.  Things to consider:

 - The script relies on a few python libraries, such as request
 - Might be a better way to approach OS info (packages, versions, etc) via osquery
 - Scanner has the potential to be split into modules/classes
 - Do we call out kernel packages that are installed but not active?
 - How can we refactor parser options as a configuration option within Hubble
 - Currently no tests have been written for this script

This is a script to provide a way to carry out local CVE scans against a local OVAL definition file.  It works on either CentOS or Ubuntu.  It builds a dictionary from an OVAL source, maps a CVE/RHSA to a package/version, pulls local package information from the system, and does version comparisons to call out any potential vulnerabilities.


basepi:
This is awesome! I won't have a chance to review it thoroughly until next week, but we'll definitely get this in! Thanks!

buddwm:
I have adapted the scanner for use with Hubble.  Please review and let me know your thoughts.

hubblestack/files/hubblestack_nova/oval_scanner.py
```diff
@@ -0,0 +1,428 @@
...
+
+def __virtual__():
+    return not sys.platform.startswith('win')
+
+
+def audit(data_list, tags, labels, debug=False, **kwargs):
+    """Hubble audit function"""
+    ret = {'Success': [], 'Failure': []}
+    for profile, data in data_list:
+        if 'oval' in profile and 'oval_scanner' in data:
```
basepi: Any particular reason you are only accepting `oval.yaml` as the profile name? Theoretically only the `'oval_scanner' in data` piece matters, and you might want the ability to name these files differently. Unless I'm missing something?

hubblestack/files/hubblestack_nova/oval_scanner.py
```diff
@@ -0,0 +1,428 @@
...
+
+import xml.etree.ElementTree as ET
+import json
+import sys
+import requests
+import logging
+
+
+def __virtual__():
+    return not sys.platform.startswith('win')
```
basepi: Could you switch to using [`salt.utils.platform.is_windows()`](https://github.com/hubblestack/hubble/blob/9849c4cb342acf1dc08f5b4b4d57216a010caee8/hubblestack/files/hubblestack_nova/win_firewall.py#L19) just for consistency across the codebase? Obviously if we end up removing the salt import it will need to be changed, but it's easier of all of these checks look the same.

hubblestack/files/hubblestack_nova/oval_scanner.py
buddwm: I'll go ahead and make these adjustments - I wasn't sure if we wanted to pin down the yaml file name or not.

hubblestack/files/hubblestack_nova/oval_scanner.py
buddwm: I'll implement these changes.

buddwm:
I've merged the requested changes @basepi .

basepi:
Whooo! I'm excited we have another CVE scanning alternative! Vulners just didn't cut it for a lot of people. Thanks again!

PR #660, opened 2019-06-25
Using sc to register Hubble as a service on WIndows

praksinha:
We have observed that if Hubble process is terminated, Hubble service still shows as running.
This is due to the fact that we register Hubble service via nssm. Actual windows service monitors nssm process which in turn is responsible for monitoring Hubble process.

Switching to 'sc' to check if we can register Hubble as a service with all configuration options that we have with 'nssm'.


fossam:
any update with regards to tests with using _sc_?

praksinha:
Created new PR as we still plan to use nssm but in a more efficient manner.

PR #661, opened 2019-06-26
fix linting errors for s3fs.py

Tenebriso:
Fixed most linting errors except for _Undefined variable '__opts__' /__utils__/___ , the TODO warning and _Catching too general exception_

### DIFF
--- a/hubblestack/extmods/fileserver/s3fs.py
+++ b/hubblestack/extmods/fileserver/s3fs.py
@@ -180,9 +180,9 @@ def find_file(path, saltenv='base', **kwargs):
     try:
         # jit load the file from S3 if it's not in the cache or it's old
         _get_file_from_s3(metadata, saltenv, fnd['bucket'], path, cached_file_path)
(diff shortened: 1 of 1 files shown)


hubblestack/extmods/fileserver/s3fs.py
```diff
@@ -182,9 +180,9 @@ def find_file(path, saltenv='base', **kwargs):
     try:
         # jit load the file from S3 if it's not in the cache or it's old
         _get_file_from_s3(metadata, saltenv, fnd['bucket'], path, cached_file_path)
-    except:
+    except Exception:
         if not os.path.isfile(cached_file_path):
-            raise
+            raise IOError
```
basepi: This is potentially a logic change -- does pylint not like it when you raise the enclosing exception with a simple `raise`?

Perhaps this would appease it?

```
except Exception as e:
    if not os.path.isfile(cached_file_path):
        raise e
```

PR #666, opened 2019-07-01
fix linting errors for grains/systemuuid.py

Tenebriso:
Still warns for _undefined variable_ and _broad exception_


fossam:
lgtm

PR #671, opened 2019-07-10
Fix lint fdg

Tenebriso:
Fix linting for fdg files

### DIFF
--- a/hubblestack/extmods/fdg/curl.py
+++ b/hubblestack/extmods/fdg/curl.py
@@ -14,20 +14,17 @@ a system, we don't want an attacker to be able to send that data to arbitrary
 endpoints.
 """
 from __future__ import absolute_import
-import json
 import logging
 import requests
 
-from salt.exceptions import CommandExecutionError
 
-log = logging.getLogger(__name__)
+LOG = logging.getLogger(__name__)
 
 
 def request(url,
             function='GET',
             params=None,
             data=None,
-            headers=None,
             username=None,
             password=None,
             timeout=9,
@@ -75,9 +72,6 @@ def request(url,
     data
         Payload to include for POST/PUT
 
-    headers
-        A dict of custom headers in the form {"user-agent": "hubble"}
-
     username
         Used for auth
 
@@ -96,9 +90,12 @@ def request(url,
 
     chained
         Ignored
+
+    chained_status
+        Ignored
     """
-    if chained:
-        log.warn('Chained value detected in curl.request module. Chained '
+    if chained or chained_status:
+        LOG.warn('Chained value detected in curl.request module. Chained '
                  'values are unsupported in the curl fdg module.')
 
     # Data validation and preparation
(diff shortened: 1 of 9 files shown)


hubblestack/extmods/fdg/curl.py
```diff
@@ -96,9 +90,12 @@ def request(url,
 
     chained
         Ignored
+
+    chained_status
+        Ignored
     """
-    if chained:
-        log.warn('Chained value detected in curl.request module. Chained '
+    if chained or chained_status:
```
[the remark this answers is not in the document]
Tenebriso: I was thinking if `chained_status` is true it would be the equivalent of having `chained` passed in, as it would mean another module called it, it was successfully executed and is trying to pass in a value. On a second thought now, `chained` would already narrow down that possibility so I'll remove it in the next commit

hubblestack/extmods/fdg/curl.py
```diff
@@ -14,20 +14,17 @@
...
 
-log = logging.getLogger(__name__)
+LOG = logging.getLogger(__name__)
 
 
 def request(url,
             function='GET',
             params=None,
             data=None,
-            headers=None,
```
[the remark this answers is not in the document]
Tenebriso: Oh, alright, sorry, I thought the functionality changed and since it was not used I thought there's no need for it anymore. I'll rebase with the complete function

hubblestack/extmods/fdg/curl.py
```diff
@@ -111,36 +112,58 @@ def request(url,
...
+
+    # Status in the return is based on http status
+    try:
+        response.raise_for_status()
+        return True, ret
+    except requests.exceptions.HTTPError:
+        return False, ret
+
+
+def __make_request(function, url, **kwargs):
```
basepi: Can we make this with a single leading underscore? I prefer the convention of single underscores to the name mangling of double underscore, unless we actually need the mangling. [more info](https://dbader.org/blog/meaning-of-underscores-in-python)

hubblestack/extmods/fdg/grep.py
```diff
@@ -24,23 +24,29 @@ def file(path, pattern, grep_args=None, format_chained=True, chained=None, chain
...
     The first return value (status) will be True if the pattern is found, and
     False othewise. The second argument will be the output of the ``grep``
     command.
 
     ``grep_args`` can be used to pass in arguments to grep.
     """
     if format_chained:
-        pattern = pattern.format(chained)
-        path = path.format(chained)
+        if chained_status:
```
basepi: Can you please remove these? This is actually a logic change -- I anticipate it's due to the fact that we're not using those arguments. But that's not necessarily a problem -- because fdg functions have a shared interface, then many of the functions may not use one or more of the parameters in that shared interface.

I want to keep logic changes out of lint fixes as much as possible.

(Also please remove all of these similar additions, I'm not going to leave a comment for each one.)

tests/unittests/test_process.py
```diff
@@ -131,11 +131,11 @@ def test_filterDict_invalidDict_emptyReturn(self):
         """
         expected_status, expected_ret = False, None
         status, ret = hubblestack.extmods.fdg.process.filter_dict(
-            starting_dict=[1, 2, 3], chained={1: 'a', 2: 'b'})
+            starting_dict=[1, 2, 3], chained={1: 'a', 2: 'b'}, chained_status=True)
```
basepi: You can also probably remove all of these `chained_status=True` with the changes I requested above.

hubblestack/extmods/fdg/grep.py
Tenebriso: I will clean them up, the `chained_status` argument was added afterwards and I thought this was the intended use for it. Even so, I agree that it make more sense to keep the logic changes out of the lint fixes and perhaps submit a different PR at some point

basepi: If I remember correctly (relying on my memory is dangerous!) the chained_status was just there in case a module wanted the ability to change something about its logic based on the true/false status of the previous item in the chain.

It may be that we only want to `format_chained` if the `chained_status` was True, but I just don't want to make that decision in a lint fix, as discussed. So I think leaving it out is the correct move for now. Thanks for making that change!

PR #675, opened 2019-07-15
Let pylint fail

jettero:
1. fix some unusual/rare build problems by removing venv/ and vlib/ before build setup
2. instruct jenkins to let pylint fail without reporting blocked-merge to github


jettero:
I think this is obviated by my new PR #680.

PR #677, opened 2019-07-16
Re-adding the ability to read splunk config from hubble.d

yassingh:
Somewhere along the way to 3.0.1 build, hubble lost its ability to pickup splunk configuration dynmaically.
As a result, hubble needs to be restarted if we intend to update the splunk config.

Some of our functionalities are dependent of this feature.
Making some changes in this PR to add that functionality back to hubble.

In this PR, I am recreating the file which was deleted in https://github.com/hubblestack/hubble/pull/580

Moreover, I am changing the hec/opt.py file so that first it looks in grains,get for splunk config ( and then fallback to config.get ),


yassingh:
Opening this PR in 3.0. Closing here.

PR #678, opened 2019-07-16
Override "hostname" grain to handle FQDN error

yassingh:
HI,

On some instances (especially in Tokyo and Sydney region of AWS ), a lot of instances are not able to run hubble.

This is the error which is received:
```
[root@ip-10-36-1-225 ec2-user]# hubble audit
An Error occurred while loading the config: [Errno 16] Device or resource busy
[ERROR   ] An un-handled exception was caught by salt's global exception handler:
error: [Errno 16] Device or resource busy
  File "hubble.py", line 5, in <module>
  File "hubblestack/daemon.py", line 49, in run
  File "hubblestack/daemon.py", line 585, in load_config
  File "hubblestack/hangtime/linux_itimers.py", line 159, in _frobnicator
  File "hubblestack/daemon.py", line 648, in refresh_grains
  File "salt/loader.py", line 737, in grains
  File "/opt/hubble/hubble-libs/salt/grains/core.py", line 1990, in hostname
    __FQDN__ = salt.utils.network.get_fqhostname()
  File "salt/utils/network.py", line 208, in get_fqhostname
error: [Errno 16] Device or resource busy
  File "hubble.py", line 5, in <module>
  File "hubblestack/daemon.py", line 49, in run
  File "hubblestack/daemon.py", line 585, in load_config
  File "hubblestack/hangtime/linux_itimers.py", line 159, in _frobnicator
  File "hubblestack/daemon.py", line 648, in refresh_grains
  File "salt/loader.py", line 737, in grains
  File "/opt/hubble/hubble-libs/salt/grains/core.py", line 1990, in hostname
    __FQDN__ = salt.utils.network.get_fqhostname()
  File "salt/utils/network.py", line 208, in get_fqhostname
socket.error: [Errno 16] Device or resource busy
[3269] Failed to execute script hubble
```

It turns out this error comes when the dns server is not able to resolve the hostname of the host instance. 

This error has been handled in this PR : 
https://github.com/saltstack/salt/pull/49726
But this salt will be release in August.

Unfortunately, this is a blocker bug, hubble does not run at all if dns is not able to resolve the hostname.

My PR is one solution for this problem:
I am overriding the definition of "hostname" grain. I have copied this file from /opt/hubble/hubble-libs/salt/grains/core.py
I have not changed any other grain. 
I have simply added some exception handling in "hostname" grain function.

Unfortunately, I can't do this exception handling directly in hubble's code because refreshing of grains is kinda atomic in "daemon.py".  Even if I catch exception if daemon.py's "refresh_grains" method, there are good changes that other grains won't be properly loaded because of this error.

@basepi 
Anyways, let me know if there are some other solutions for this problem.

This is supposed to be temporary fix only ( until the next version of salt is released )


hubblestack/extmods/grains/core.py
```diff
@@ -0,0 +1,2640 @@
...
+                                grains['lsb_distrib_codename'] = re.sub(r'\(.+\)', '', line).strip()
+                            elif 'version' in line.lower():
+                                version = re.sub(r'[^0-9]', '', line)
+                            elif 'patchlevel' in line.lower():
+                                patch = re.sub(r'[^0-9]', '', line)
+                    grains['lsb_distrib_release'] = version
+                    if patch:
+                        grains['lsb_distrib_release'] += '.' + patch
+                        patchstr = 'SP' + patch
+                        if grains[
```
yassingh: Earlier, this error used to come in the form of "socket.gaierror"

BUt not it comes as "socket.error".

That's the main reason behind this issue.

yassingh: I have edited only this grain.
Others are untouched.

yassingh:
I have completed my tested and it is ready to be merged.

I would also like @fossam 's input here.
The output on splunk will be slighly changed because of this ( only in case of faulty instances ).
I have give you more info on slack.

jettero:
I don't think this is probably the best way to do this. Until we remove salt from the hubble mainline, ... rather than copy this file and eventually expect it to get out of sync with salt mainline ... I think a monkeypatch would be a better option.  E.g.:

        import salt.grains.core
        salt.grains.core.offending_function = our_offending_replacement_function

I get the motivation for sure, and because we plan to copy parts of salt into hubble and remove the salt dependency anyway, perhaps my opinion is totally wrongheaded here. Still, as the PR is written, I can't tell which parts of grains.core is copied directly and which parts are the same. Another way to do it might be to supply a patch file that we could apply to our installed copy of salt.grains.core? Maybe?

I don't have any really good solution, just reservations about copying this file into extmods.  I can't wait to see what @basepi thinks.

yassingh:
Thanks, @jettero  and @basepi 
Your solution is far neater and easier than mine.
(Unfortunatley, such Pythonc tricks always elude my java-trained mind)

And please feel free to point out any other issue/suggestion with the code.

yassingh:
I have made the requested changes.
I have completed my testing, and the fix seems to be working fine.

jettero:
I wonder now too ... ex post facto ... if there should be a patch to the hangtime.py or in refresh_grains or something. It looks like the hangtime linux itimer is what's causing the crash in grains core. I'm assuming that's the one started by the decorator on refresh_grains... presumably before the daemon starts up.

basepi:
@jettero might be worth some testing, creating custom grains to throw errors in odd ways and seeing if we handle them gracefully. But we still want this fix so that's independent.

PR #680, opened 2019-07-16
coverage and profile

jettero:
* turned the output of pytest into an html report
* turned the output of pylint into an html report
* added bandit output
* added profiling to pytest to show a heatmap of time spent during pytest — we can later expand this to show memory problems and areas that need performance fixes

Also, 
* make sure pylint doesn't cause the "build" to fail and block github merge
* same for bandit ... the output shouldn't block anything on github until we decide if we even want its output

There should be a new nav link on the left that says Test Reports which leads to this:

https://jenkins.hubblestack.io/job/hubble/view/change-requests/job/PR-680/Test_20Reports/


jettero:
> My only concern with this is that now github says all checks have passed from jenkins. But if pylint fails we want it to show as a failure in the PR.  
> Does that make sense?

That absolutely makes sense. I consider the fail-but-pass situation completely temporary for both pylint and bandit. Until we intend to actually require those to pass, we shouldn't show them as preventing the merge (imo). It's pretty trivial to make pylint block the merge once we get to that stage.

No?

basepi:
OK, that sounds fine. We'll have them pass until we trust upstream to pass. Thanks!

PR #681, opened 2019-07-16
Automatically populate the splunkindex grain

basepi:
Some users were still on the old-style splunk returner config
and were missing this grain. We can fix that in the config, but we
decided we wanted to make sure this grain was bulletproof. If the config
exists, we populate the grain.


basepi:
Also updated this PR to use the grains loaded by splunkconfig.py first, falling back to opts otherwise, and to set another grain if grains and opts differ (so we know if the fallback mechanism has occurred.)

PR #682, opened 2019-07-17
Avoiding splunk token from getting logged among grains 

yassingh:
After this PR : https://github.com/hubblestack/hubble/pull/679 , splunk config loaded via hubble.d/*.conf will become a part of __grains__.

As a result, splunk related details will be published on spunk, including the token.
So I am trying to omit token ( and other such sensitive details ) from splunk logs.

Instead of writing completely new logic, I am reusing some functions which were present in extmods/modules/conf_publisher.py 

I have made more comments in the PR for explanation.

### DIFF
--- a/hubblestack/extmods/modules/conf_publisher.py
+++ b/hubblestack/extmods/modules/conf_publisher.py
@@ -41,7 +41,7 @@ def publish(report_directly_to_splunk=True, remove_dots=True, *args):
             if arg in  __opts__:
                 opts_to_log[arg] = __opts__[arg]
 
-    filtered_conf = _filter_config(opts_to_log, remove_dots=remove_dots)
+    filtered_conf = hubblestack.log.filter_logs(opts_to_log, remove_dots=remove_dots)
 
     if report_directly_to_splunk:
         hubblestack.log.emit_to_splunk(filtered_conf, 'INFO', 'hubblestack.hubble_config')
@@ -49,30 +49,3 @@ def publish(report_directly_to_splunk=True, remove_dots=True, *args):
 
     return filtered_conf
 
-
-def _filter_config(opts_to_log, remove_dots=True):
-    '''
-    Filters out keys containing certain patterns to avoid sensitive information being sent to splunk
-    '''
-    patterns_to_filter = ["password", "token", "passphrase", "privkey", "keyid", "s3.key"]
-    filtered_conf = _remove_sensitive_info(opts_to_log, patterns_to_filter)
-    if remove_dots:
-        for key in filtered_conf.keys():
-            if '.' in key:
-                filtered_conf[key.replace('.', '_')] = filtered_conf.pop(key)
-    return filtered_conf
-
-
-def _remove_sensitive_info(obj, patterns_to_filter):
-    '''
-    Filter known sensitive info
-    '''
-    if isinstance(obj, dict):
-         obj = {
-             key: _remove_sensitive_info(value, patterns_to_filter)
-             for key, value in obj.iteritems()
-             if not any(patt in key for patt in patterns_to_filter)}
-    elif isinstance(obj, list):
-         obj = [_remove_sensitive_info(item, patterns_to_filter)
-                    for item in obj]
-    return obj
--- a/hubblestack/log.py
+++ b/hubblestack/log.py
@@ -7,8 +7,12 @@ from __future__ import print_function
 
 import logging
 import time
+import copy
 
 import hubblestack.splunklogging
+
+# These patterns will not be logged by "conf_publisher" module and emit_to_splunk
+PATTERNS_TO_FILTER = ["password", "token", "passphrase", "privkey", "keyid", "s3.key"]
 
 # While hubble doesn't use these, salt modules can, so let's define them anyway
 SPLUNK = logging.SPLUNK = 25
@@ -154,10 +158,17 @@ def setup_splunk_logger():
     SPLUNK_HANDLER = handler
 
 
-def emit_to_splunk(message, level, name):
+def emit_to_splunk(message, level, name, filter_logs=False):
     '''
     Emit a single message to splunk
     '''
+
+    if filter_logs:
+        # copying the original message so that it does not
+        # get modified while filtering
+        copy_of_message = copy.deepcopy(message)
+        message = filter_logs(copy_of_message, remove_dots=True)
+
     if SPLUNK_HANDLER is None:
         return False
     handler = SPLUNK_HANDLER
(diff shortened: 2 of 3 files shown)


hubblestack/log.py
```diff
@@ -177,3 +188,31 @@ def handle(self, record):
     sls.LOGGING_NULL_HANDLER.sync_with_handlers([flh])
     # if flh.count > 0:
     #     log.info("pretended to handle %d logging record(s) for salt.log.setup.LOGGING_*_HANDLER", flh.count)
+
```
yassingh: moving these functions to hubblestack/log.py

hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -41,38 +41,11 @@ def publish(report_directly_to_splunk=True, remove_dots=True, *args):
...
-    filtered_conf = _filter_config(opts_to_log, remove_dots=remove_dots)
+    filtered_conf = hubblestack.log.filter_logs(opts_to_log, remove_dots=remove_dots)
 
     if report_directly_to_splunk:
         hubblestack.log.emit_to_splunk(filtered_conf, 'INFO', 'hubblestack.hubble_config')
         log.debug('Published config to splunk')
 
     return filtered_conf
 
-
```
yassingh: moving these functions to hubblestack/log.py

hubblestack/log.py
```diff
@@ -154,10 +158,17 @@ def setup_splunk_logger():
     SPLUNK_HANDLER = handler
 
 
-def emit_to_splunk(message, level, name):
+def emit_to_splunk(message, level, name, remove_sensitive_logs=False):
     '''
     Emit a single message to splunk
     '''
+
+    if remove_sensitive_logs:
```
yassingh: filter_logs is trying to modify the "opts_to_log" object ( I might be wrong here ). I don't want it to edit the __grains__ so I am making a deepcopy here.

Though this is a computattionally expensive operation, I think we can tolerate this much expense since 'hubblestack.grains_report' is NOT generated very frequently.

@basepi  @jettero , please review my code from "memory leak" perspecitive also because __grain__ might be a huge object in some cases.

basepi: I don't think this is necessary. `_remove_sensitive_info` uses list/dictionary comprehensions, which construct new objects. So by the time we get to the `remove_dots` modifications in `filter_logs` we are already in a copy.

@jettero would you mind confirming this for me? I want a second set of eyes.

hubblestack/extmods/modules/conf_publisher.py
```diff
@@ -41,38 +41,11 @@ def publish(report_directly_to_splunk=True, remove_dots=True, *args):
             if arg in  __opts__:
                 opts_to_log[arg] = __opts__[arg]
 
-    filtered_conf = _filter_config(opts_to_log, remove_dots=remove_dots)
+    filtered_conf = hubblestack.log.filter_logs(opts_to_log, remove_dots=remove_dots)
```
basepi: Looks like we do an `emit_to_splunk` on line 47 -- I think we can eliminate this step and just add `remove_sensitive_logs=True` to line 47, right? Or am I missing something?

hubblestack/log.py
jettero: That seems right, except that if the obj passed to _remove_sensitive_info isn't a list or a dict (eg, a tuple, or a namedtuple) then it wouldn't be filtered or copied. Probably this isn't normally going to be the case, but I wonder if that section (isinstance(list) or isinstance(dict)) should have an else: raise ValueError("trying to filter !list/!dict, doh") or something like that.

basepi: Except that we need to be able to emit things that aren't lists or dictionaries.

I recommend that we remove this deepcopy, but surround the `filter_logs()` call below with an `if isinstance(message, (list, dict)):` to make sure we only filter on lists/dicts.

Thanks @yassingh!

yassingh: Thanks for suggestions.
It's done 👍

hubblestack/log.py
```diff
@@ -154,10 +158,14 @@ def setup_splunk_logger():
...
 
 
-def emit_to_splunk(message, level, name):
+def emit_to_splunk(message, level, name, remove_sensitive_logs=False):
     '''
     Emit a single message to splunk
     '''
+
+    if isinstance(message, (list, dict)):
+        message = filter_logs(message, remove_dots=False)
```
yassingh: Note that I have changed the default value of  "remove_dots" argument to False.

Firstly, I don't want to remove dots from “hubblestack.grains_report” logs. I fear that it could break people's splunk search queries .

Plus I had some other doubts:

Q. Why didn't I add another parameter in the signature of "emit_to_splunk" for "remove_dots"?
A. I don't want to complicate the signature of emit_to_splunk further.  "remove_dots" won't make any sense if "remove_sensitive_logs" is False.

Q. What if someone want's to remove_dots also ?
A. IN that case he should call "filter_logs" first ( with  remove_dots=True ). Then he can call "emit_to_splunk" (  with remove_sensitive_logs=False as the logs have already been filtered )

Q.  Before this commit ( when remove_dots was set to True ) , could we cover the scenario where someone wants to filter the logs but don't want to remove dots ?
A. Nope. That's why I changed it to remove_dots=False. Now one can cover all the scenarios by mix and match of "emit_to_splunk" and "filter_logs"


I know, not very strong reasoning. But I had to go with atleast one of the choices. 
Let me know if someone thinks otherwise.

hubblestack/extmods/modules/conf_publisher.py
yassingh: Line 47 ( where we emit to splunk ) may or may NOT get executed depending upon the 'report_directly_to_splunk' flag.

Secondly, I have not exposed any parameter in "emit_to_splunk" function to pass the value of 'remove_dots' .

Because of these reasons, I have not deleted this step.

hubblestack/log.py
basepi: Our reasoning is for conf_publisher, not grains report. There are some settings with dots in the config -- the ones that come to mind are for s3 credentials (s3.key, s3.keyid, etc). The problem with this is that searching nested data structures in splunk is dot-based. So if you want to search for those keys, it becomes very difficult and you have to deal with escaping and it makes life hard.

I am going to review this further and maybe make it a conf_publisher argument instead.

hubblestack/extmods/modules/conf_publisher.py
basepi: You're right, of course. I was being silly. Thanks for catching that for me!

hubblestack/log.py
basepi: Oh, we already do that. We're good! This should be fine.

PR #693, opened 2019-07-29
Fixing windows service management

praksinha:
Hubble service management on windows is not optimal.
We have noticed if Hubble process is terminated or killed, Hubble service still shows as running.


pkg/windows/pyinstaller-requirements.txt
```diff
@@ -7,7 +7,7 @@ pprint
 daemon
 boto3
 botocore
-pygit2
+pygit2==0.27.3
```
praksinha: https://github.com/saltstack/salt/issues/51664
Pinning to version 0.27.3 as salt's gitfs throws error with pygit2 version 0.27.4 and above. Until new salt version is released in August 2019, we need to pin this.

basepi: Do we need to change the dockerfiles which build libgit2 to also pin it? I can't remember if libgit2 guarantees backwards compatibility with old pygit2 versions.

praksinha: So I tested the changes of this PR without making any changes in Dockerfile and it worked. So I guess libgit2 is compatible with this version of pygit2. Plus we are not pinning to very old version of pygit2 and would probably un-pin this in next 1-2 months, once new version of salt is released.

praksinha: I think yes, since we are pinning to 0.27.x and latest released is 0.28.x. So we need to keep libgit2 also like 0.27.x. I'm still testing these changes and will update dependencies wherever required.

praksinha: As per our discussion, I'm not pinning to older version of pygit2 as there were other issues introduced like compatibility with libgit2 etc.
When new version of salt is released this month, this should automatically get fixed. So any Hubble version released in Sept 2019 or later should work fine.

praksinha:
Tested on windows server 2016 and 2019. Will also test on windows server 2012.

praksinha:
Tested on windows server 2012 as well.

PR #702, opened 2019-08-15
defaults fix

jettero:
I have mixed feelings about this. The problem is clearly with how `_nick` functions and the docs that surround it. But rather than fix that, which is hard, I just fixed the usage of it in the returners that use it.

In particular: we want to make sure if `sourcetype_nebula` is undefined that it gets the intended default.


basepi:
This will need to be backported.

PR #704, opened 2019-08-16
fix for cloud_details grain

fossam:
Bypass proxies when accessing metadata services.

This is not tested!

Let me know if the commit needs to be against 3.0 branch instead.

### DIFF
--- a/hubblestack/extmods/grains/cloud_details.py
+++ b/hubblestack/extmods/grains/cloud_details.py
@@ -33,9 +33,11 @@ def _get_aws_details():
     aws['cloud_account_id'] = None
     aws['cloud_type'] = 'aws'
 
+    proxies = { 'http': None }
+
     try:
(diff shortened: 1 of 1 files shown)


basepi:
We can just backport it, thanks for adding the label. Do we have a good way to test this? Wasn't it @praksinha that saw this issue?

praksinha:
@fossam I have tested these changes and it fixed the issue where users were setting an environment variable like HTTP_PROXY. However, I'm still testing this and making sure we haven't left any other corner case.

I would like to add some logic to set cloud_details fields to something like 'Not Available' if metadata urls don't work instead of 'None' or a weird HTML text. Since while refreshing grains, we persist cloud_details, I'm afraid it would persist the value of these grains to invalid strings or HTML even if later on those URLs started working. So, while refreshing grains we should only persist cloud_details if it is not set to 'Not Available' which I think is still better than showing an AWS instance as 'gcp' instance. No information is better than incorrect information. Thoughts ?

PR #705, opened 2019-08-20
from hubblestack.extmods.fdg.process import encode_base64 issues …

jettero:
from hubblestack.extmods.fdg.process import encode_base64 is fine if you're
running a source install but if the live readfile.py is in /var/cache, this
causes an import problem.

Let's try using local import from process (which should be near to readfile.py
either in /var/cache or in $LIBDIR/hubblestack/extmods/fdg

### DIFF
--- a/hubblestack/extmods/fdg/process.py
+++ b/hubblestack/extmods/fdg/process.py
@@ -8,11 +8,10 @@ the data outputted by a module to serve it to another module.
 """
 from __future__ import absolute_import
 
-import base64
 import logging
 import re
 
-import salt.ext.six as six
+from hubblestack.utils.encoding import encode_base64 as utils_encode_base64
 from salt.exceptions import ArgumentValueError
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


jettero:
I think this addresses the immediate problem, but it'd probably better to create a utility function that we can import with absolute import.

basepi:
Yeah I would probably make something in hubblestack.utils that does the functionality, and have both process.py and readfile.py import that util. Cleaner and more explicit.

jettero:
I'm inclined to leave this is a temporary fix so I can hurry and get it into 3.0.7 though. That's apparently pretty rushed at the moment.

Tenebriso:
That look great and clean, nice! I think we need to change the unittests as well since we moved the function they are going to fail, maybe move the ones for encode_base64 from test_process.py to a test_encoding.py?

jettero:
> I think we need to change the unittests as well since we moved the function …

Well, I moved it, but I also kept it there. I wouldn't think it'd make much difference for the tests. Possibly we need a new test_encoding set, but it'd be the same-ish, no?

Oh, I see the problem. forgot to import logging. Good job tests! \o/

PR #706, opened 2019-08-21
Adding single proxy override variable support

praksinha:
Currently if user has to configure 'proxy' in Hubble configuration, he needs to update both fileserver block and splunk config block to specify the proxy.
With this change, users will now have an option to configure HTTPS proxy using single parameter `https_proxy`. Current behaviour has been kept intact and will take preference over global proxy settings.


hubblestack/hec/obj.py
```diff
@@ -147,6 +147,7 @@ def __init__(self, token, index, http_event_server, host='', http_event_port='80
         self.currentByteLength = 0
         self.server_uri = []
 
+        proxy =  proxy or __opts__.get('https_proxy', None)
```
fossam: Is calling it `https_proxy` correct here or would it be applied also in case if http scheme is used?

praksinha: @fossam I intended to name it as 'proxy' but it raised a conflict in salt's loader.py. Moreover as of now we are only using this variable in azurefs and splunk which both run on port 443. If in future there's a need for http only proxy, we can add a parameter like 'http_proxy'.

fossam: Thank you. 
Can the setting and the caveats be documented in some form, say in example configuration file?

praksinha: Sure, I'll add a sample config block and document it as well.

jettero:
Ignoring reviews is nice.

hubblestack/hec/obj.py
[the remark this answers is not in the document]
praksinha: Will make changes in PR #709

PR #708, opened 2019-08-22
pr705 backport to 3.0

jettero:


fossam:
thank you for changing to a title that is descriptive.

PR #709, opened 2019-08-22
Fixing global opts reference

praksinha:
Fixing global __opts__ reference which was missed in PR #706


jettero:
No, just no. __opts__ shouldn't be used in either hec/obj.py nor hec/opt.py ... you should use `salt['config.get']` instead and only in hec/opt.py. That is the purpose of those.

PR #717, opened 2019-09-02
Adding support for overriding splunk configuration via global parameters

arshiyaaggarwal:
Initially the user had to provide the whole block of parameters in /etc/hubble/hubble.d/user.config with a format similar to the default config block in /etc/hubble/hubble. Now the user can specify only the index, token and port in /etc/hubble/hubble using the global parameters splunk_token, splunk_index and splunk_port. These options will be given priority over the hubblestack:returner:splunk block. Index and token must be specified together or not at all.

### DIFF
--- a/hubblestack/hec/opt.py
+++ b/hubblestack/hec/opt.py
@@ -36,11 +36,22 @@ def _get_splunk_options(space, modality, **kw):
 
     confg = __salt__['config.get']
 
+    # both index and token must be specified if at all overriding in /etc/hubble/hubble
+    # is taking place using the variables splunk_token and splunk_index
+
+    if bool(confg('splunk_token', None)) != bool(confg('splunk_index', None)):
+        raise Exception('Both index and token must be specified together or not '
+                        'specified at all in case of overriding')
+
+    # additionally overriding can be carried out using 'splunk_index', 'splunk_token' and 'splunk_port'
+    # in the /etc/hubble/hubble file
+    # these are given priority over the data in 'hubblestack:returner:splunk' block in the default config
+
     base_opts = {
-        'token': REQUIRED,
+        'token': confg('splunk_token', REQUIRED),
         'indexer': REQUIRED,
-        'index': REQUIRED,
-        'port': '8088',
+        'index': confg('splunk_index', REQUIRED),
+        'port': confg('splunk_port', '8088'),
         'custom_fields': [],
         'sourcetype': 'hubble_log',
         'http_event_server_ssl': True,
@@ -69,7 +80,14 @@ def _get_splunk_options(space, modality, **kw):
             for k in opt:
                 j = nicknames.get(k, k)
                 if j in final_opts:
-                    final_opts[j] = opt[k]
+                    # if j is one of the args that can be overridden and has been provided then do not update it
+                    if j in {'token', 'index', 'port'}:
+                        if not bool(confg("splunk_" + j, None)):
+                            final_opts[j] = opt[k]
+                            # print final_opts
+                    else:
+                        final_opts[j] = opt[k]
+
             if REQUIRED in final_opts.values():
                 raise Exception('{0} must be specified in the {1} configs!'.format(req, space))
             ret.append(final_opts)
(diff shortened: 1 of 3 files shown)


hubblestack/hec/opt.py
```diff
@@ -69,7 +82,13 @@ def _get_splunk_options(space, modality, **kw):
             for k in opt:
                 j = nicknames.get(k, k)
                 if j in final_opts:
-                    final_opts[j] = opt[k]
+                    # if j is one of the args that can be overridden and has been provided then do not update it
+                    if j in {'token', 'index', 'port'}:
```
praksinha: You have not used 'options_for_grains_config' dict defined at the beginning. I guess you intended to use it here ?

praksinha:
@arshiyaaggarwal We should also make sure 'splunk_token' is not logged into splunk. I guess in current state, it would get loaded in __opts__ and eventually get logged into splunk by conf_publisher. We should handle this.

hubblestack/hec/opt.py
```diff
@@ -30,17 +30,38 @@ class Required(object):
...
+        raise Exception('Both index and token must be specified together or not '
+                        'specified at all in case of overriding')
+
+    if confg('splunk_token', None) and not str(confg('splunk_token', None)).strip():
+        raise Exception('splunk_token cannot be an empty field')
+
+    if confg('splunk_index', None) and not str(confg('splunk_index', None)).strip():
+        raise Exception('splunk_index cannot be an empty field')
+
+    if confg('splunk_port', None) and not str(confg('splunk_port', None)).strip():
```
praksinha: You should also check if 'splunk_port' is integer value. It should not be a random string.

jettero: I'm not sure checking integer or even str(config()) is necessary, and certainly not strip().

The salt yaml loader does most of that heavy lifting for us (str vs int and strip()). And I think if someone puts "dog" for the port number, it won't be a very different error than if they just use the wrong port. You'd get a network stack error and think: hrm, 'dog' might not be the right port number.

jettero: (Also, if you were to put "splunk 8080" at the end of /etc/protocols, then "splunk" might just be a valid port number.)

praksinha: That's good info that salt's loader would take care of empty strings, didn't know that.
Regarding /etc/protocols entry, that would be a valid scenario not sure if someone would ever use it with Hubble.

For integer check I was thinking of handling typos in config (especially hubble.d/) w.r.t. port field. Salt wouldn't know if that variable should have integer or string values which we could handle and fail fast instead of giving weird connection errors. 

For most of Adobe's use-cases/deployments it wouldn't matter, it's more of 'good to have' sort of check.

jettero: I've used text based port numbers plenty... both with things like socket.open('hostname:portname') and in config files. Some programs can deal with it and others can't because people check for digits.

I think it's fine to check that the port number is really a number here, so I won't make any further argument about it, but it'd make me smirk if I was an SRE and noticed it was so inflexible.

Futher notes on the yaml conversions... I think it wouldn't take care of empty strings if you told it not to do so. Definitely the salt code avoids checking types and `strip()`s in most places though. I think it'd make sense to stay consistent with the rest of salt and hubble and assume it's stripped.

But that said, if you used quotes to force the leading spaces, they'd survive into the `__opts__`.

```
blah:
   intentionally_spacey: '   leading spaces, and trailing ones   '
   autostripped:          this won't have leading spaces regardless, just part of yaml
```

Also, on the number conversions, when converting from yaml, salt is going to automatically convert all-number fields to int and leave strings alone... which is the source of some common errors in salt, particularly around octal.

```
config:
   some_integer: 755
   wrong_file_mode: 0755
   right_file_mode: '0755'
```

When a field is all number characters, the salt-yaml conversion makes it `int(755)`, so some_integer is fine... it's `int(755)` automatically. 

wrong_file_mode will actually give `int(755)==int(0o1363)` (sometimes) though. That's probably not what the salt user wanted. The salt-yaml conversion seems to not know about leading 0's in some implementations, does know about it in others. The salt best practices (and file.manged notes) therefore indicate that you should mark file modes as strings so they're handed to `chmod` correctly. That means using the quote markers to avoid the auto conversions: `'0755'`.

praksinha: ok. Let's not check for integer port numbers then.

praksinha:
@fossam Do you have any comments ? As of now, only index, token and port are made configurable via global parameters. Also making sure token doesn't get logged into splunk.

PR #719, opened 2019-09-16
Clean linting for graylog returners

Tenebriso:


jettero:
this seems to have failed pr-merge testing ... Looks like it maybe didn't pick up the new change to `list(set(dict()))` dedup (maybe).

PR #721, opened 2019-09-16
clean slack returner

Tenebriso:


hubblestack/extmods/returners/slack_pulsar_returner.py
```diff
@@ -62,18 +62,37 @@
...
-from salt.ext.six.moves.urllib.parse import urljoin as _urljoin  # pylint: disable=import-error,no-name-in-module
+from salt.ext.six.moves.urllib.parse import urljoin as _urljoin
+# pylint: disable=import-error,no-name-in-module
 import salt.ext.six.moves.http_client
 # pylint: enable=import-error,no-name-in-module,redefined-builtin
 
 # Import Salt Libs
 import salt.returners
 
-log = logging.getLogger(__name__)
```
jettero: could we keep this one as `log` (rather than `LOG`) also please?

hubblestack/extmods/returners/slack_pulsar_returner.py
```diff
@@ -261,33 +261,33 @@ def returner(ret):
...
 
     if not username:
-        log.error('slack_pulsar.username not defined in salt config')
-        return
+        LOG.error('slack_pulsar.username not defined in salt config')
+        return None
 
     if not as_user:
-        log.error('slack_pulsar.as_user not defined in salt config')
-        return
```
jettero: `return None` is something the linter recommends?

hubblestack/extmods/returners/slack_pulsar_returner.py
Tenebriso: yeah, I was planning to go over all the LOG changes and get them back to log, thanks for the heads up

hubblestack/extmods/returners/slack_pulsar_returner.py
Tenebriso: It doesn't like that some return statements return something and others don't, it says all of them should return something or none of them should. I figured returning None would be the fix for that, should I have it return something else instead?

jettero: Oh, right. I've seen that before now that you say it like that.  No, it's definitely right the way it is, it just seemed odd to me.

hubblestack/extmods/returners/slack_pulsar_returner.py
jettero: This was just the first one I saw...

PR #722, opened 2019-09-17
`log = logging.getLogger(__name__)`

jettero:
I know all the globals should be allcaps, but as long as I've been writing programs, allcaps has meant "constant" not "global".

I've seen `log = logging.getLogger(__name__)` as a global a billion times. I've never seen it as `LOG`. I think that's just a side effect of the pylint defaults.

I'm wondering if there's other things we should consider changing in the pylintrc.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -637,7 +637,9 @@ good-names=i,
            k,
            ex,
            Run,
-           _
+           _,
+           log
+
 
 # Include a hint for the correct naming format with invalid-name
 include-naming-hint=no


Tenebriso:
Oooooh, right. Good, I'll get it back to `log`. I wasn't aware that's a standard name like the `i` in fors, I wouldn't have changed it, I guess my lack of experience was showing.

jettero:
To be fair, I'm speaking more from anecdote and opinion than "standard." Arguably it shouldn't be global at all. But if nobody objects, I think the pylintrc route is the better one by far.

PR #723, opened 2019-09-17
Clean splunk returners

Tenebriso:


hubblestack/extmods/returners/splunk_osqueryd_return.py
```diff
@@ -40,176 +40,180 @@
...
-            # Get cloud details
-            cloud_details = __grains__.get('cloud_details', {})
-
-            if not data:
-                return
-            else:
-                for query_results in data:
-                    event = {}
-                    query_name = query_results['name']
-                    event.update({'query': query_name})
```
jettero: I wonder if we could use `hubblestack.utils.stdrec.update_payload()` here? If it can't do the this section, perhaps it could be taught to do so? Then the 4 or 6 PRs that all seem to have this section could be using the exact same setup.

Though, I think the non-splunk returners might need somewhat different time fields.

PR #725, opened 2019-09-25
clean linux_itimers

Tenebriso:


fossam:
Just as a comment - I for one _really_ like the more clear variable naming. Thank you!

hubblestack/hangtime/linux_itimers.py
```diff
@@ -90,21 +95,22 @@ def restore(self, ended=False):
...
+                signal.setitimer(signal.ITIMER_REAL, time_remaining)
             else:
-                p.restore()
+                prev.restore()
         else:
             log.debug("timer stack empty after %s, resetting signals/itimers", repr(self))
             signal.setitimer(signal.ITIMER_REAL, 0)
             signal.signal(signal.SIGALRM, signal.SIG_DFL)
 
-    def fire_timer(self, *sig_param):
```
jettero: I don't think this will actually work. when signal() fires, it passes two arguments to the callback function. without either the two args or a star-args, it'll likely crash during the signal. Have you tested this?

PR #726, opened 2019-09-25
clean sqlite returner

Tenebriso:


hubblestack/extmods/returners/sqlite.py
```diff
@@ -4,21 +4,24 @@
...
 import logging
 import os
-import json
+import salt.returners
+
 try:
     import sqlite3
+
     GOT_SQLI = True
-except:
```
jettero: I'm surprised except Exception passes where except: doesn't... I wonder if we should select the exception for "no such module" instead.

jettero: on second thought... who cares. this module is only for testing anyway, and if we ever use it again, we can fix it then.

PR #730, opened 2019-10-08
CPU improvements in diskqueue pop()/get()/put()

jettero:
I was experimenting with semi-large queues and discovered that the O(n²) operation of counting the disk usage on every `pop()` is (gasp) expensive. Probably estimating the disk queue size on a `pop()` will rarely ever matter, and probably saves us a ton of CPU time.

In fact, we can do roughly the same thing to `get()` and `put()`.

We leave counting disk space completely to `__init__()` and `getz()` — `getz()` gets as many messages as will fit in the splunk max message size.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,3 +119,7 @@ relevant-files.txt
 
 # PyCharm project
 .idea/
+
+# sometimes when I stop docker with a sigquit, hubble drops core (??)
+core
+core.*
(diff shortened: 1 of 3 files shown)


jettero:
I marked this as do-no-merge until I complete some basic testing.

jettero:
I checked this like crazy. There *are* discrepancies between the estimator method and the full `_count()` method; but they balance out over time. I didn't dig into why there's discrepancies, but if I abort hubble during a discrepancy and issue `dq._count()` from ipython; it matches the estimated values. I think it relates to internal python file write caching (8192k buffer by default).

jettero:
rebasing this to develop was a bummer after already changing this code in the behavior PR

PR #732, opened 2019-10-22
change syntax to make it compatible with python 3

Tenebriso:
At the moment the main point of failure is disk queue, with the transition to python3 a few things changed: b'foo' is now of type bytes instead of str and writing to file in binary mode ('wb') does not allow str type (expected bytes) which mainly breaks hec/dq.py. A possible fix would be changing `ok_types` to `bytes` instead of `str` and convert input that is of type str (`str.encode()`)
Comparing int to str is no longer possible so sorting breaks if lists contain items of different types, a naive fix would be converting items to string before comparing but we should definitely look closer and perhaps rewrite the logic behind.
Another issue I have is that I need to run hubble with `-v` because NoneType cannot be compared to int (only "==" and "!=" are allowed) which breaks load_config() in daemon.py


jettero:
this has been in the queue forever. @Tenebriso, IMO, you're the merge captain for the py3 branch. I can't see any reason to not merge this.

Tenebriso:
py2 doesn't mind because it allows for comparisons between `NoneType` and `int`, it just returns `False`, whereas py3 raises a TypeError so we need extra guardrails

jettero:
Oh, I guess I apprehended that, but never comprehended it.

fossam:
76 files - currently, this should not be `To Backport`, should it?

jettero:
> 76 files - currently, this should not be `To Backport`, should it?

Ah, no. definitely not. oops.

PR #734, opened 2019-10-23
Adding fdg module for certificate discovery

MoodyMudit:
This module is capable of connecting to any port and fetching certificate details for a certificate attached on it. The module is supposed to work in conjunction with fdg's osquery module. The osquery module would supply information about open ports after which this module extracts certificate information.


MoodyMudit:
@fossam I have updated the module to reflect your comments. Let me know if you have more questions/comments. Thanks.

jettero:
I'm waiting for futher comment from @fossam before I click approve, but this looks good to me.

fossam:
@jettero related and important bug that should be addressed before 3.0.8:
when testing this module I observed a the following sourcetype:
```sourcetype = hubble_fdg_salt://fdg/fdg_cert_test.fdg```
This was tested on 3.0.7 but with added fdg cert file function.

**Comment by fossam** [comment]
@MoodyMudit Thank you for adding the comments. Also, thanks for adding the additional fields per our Slack discussion.

I was able to test successfully the very latest version with the following fdg file:
```
main:
  module: cert_discovery.get_cert_details
  args:
    - host_ip: 'google.com'
    - host_port: '443'
```
However, I was not able to run test that uses xpipe (which worked earlier) - my file looks the following:
```
main:
  module: osquery.query
  args:
    - "select 'google.com' as host_ip, '443' as host_port ;"
  xpipe:
    do_certs

do_certs:
  module: cert_discovery.get_cert_details
```
Perhaps I am doing something wrong but this worked before adjusting to the currently expected field names.

To shortcut this - can you include verified fdg test case(s) with the PR.

**Comment by MoodyMudit** [comment]
@fossam Thanks for pointing that out. It was a difference between whether we want an fdg profile to have default values or not. I had changed the code to work with the following. **Notice the blank values passed to cert_discovery.get_cert_details**
`main:
  module: osquery.query
  args:
    - "select 'google.com' as host_ip, '443' as host_port ;"
  xpipe:
    do_certs

do_certs:
  module: cert_discovery.get_cert_details
  args:
    - host_ip: ''
    - host_port: '' `
But yes, it makes sense to not pass blank values to the function when the first module already passes some values.
TL;DR - Fixed the issue.
I plan to add test cases as soon as possible. I was hoping that we remove as many bugs and code flaws as possible, hence the 'DoNotMerge' label :).

**Comment by jettero** [comment]
@fossam any idea how that sourcetype was produced? do you see it each time you run this PR?

**Comment by fossam** [comment]
@jettero i don't know why it is happening. I suspect the sourcetype being like that is not connected to the behavior of this module, but that's just a guess. But I doubt sourcetypes like that are a good practice in Splunk.

**Comment by jettero** [comment]
got it (I think): see #738

**Comment by fossam** [comment]
@MoodyMudit Can you remove leading space from the list in Subject Alternative Names list?
Currently the modules outputs the list something to the effect of: 
`
... " DNS:*.google.co.in", " DNS:*.google.co.jp", " DNS:*.google.co.uk", ...
`

I think parsing the list would be easier without the leading space.

File: hubblestack/extmods/fdg/cert_discovery.py [hunk 1]
```diff
@@ -0,0 +1,168 @@
...
+
+    chained
+        The value chained from the previous call.
+
+    chained_status
+        The status returned by the chained call.
+    """
+    if host_ip == "":
+        host = chained['host_ip']
+    else:
```
praksinha: These if else can be combined into something like: `host = chained.get('host_ip', host_ip)` 
You should also verify before hand that `chained` arg is not `None` before invoking `.get`.

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,168 @@
...
+        host = host_ip['host_ip']
+    if host_port == "":
+        port = chained['host_port']
+    else:
+        port = host_port['host_port']
```
praksinha: Same optimization for `port` if else

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,168 @@
...
+    cert = load_certificate(host, port)
```
praksinha: Verify for `host, port` to be non empty before invoking `load_certificate` as even in that function you are directly utilising these params and not validating them.

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,168 @@
...
+    if not cert['result']:
```
praksinha: Better to use `cert.get('result', None)`expression as above expr can throw exception if key not found

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,168 @@
...
+    return san
+
+def load_certificate(ip, port):
+    """
+    fetch server certificate details and return Json with the first value being the
+    status of ssl.get_server_certificate function and second value being the actual
+    certificate data.
+    """
+    try:
+        log.info("FDG's cert_discovery is checking for ssl cert on {0}:{1}".format(ip,port))
```
praksinha: I guess we can change this to `log.debug` to reduce noise in logs as you are already logging certificate found message in case of a valid cert.

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,168 @@
...
+        hostport = (str(ip), int(port))
+        cert_details = ssl.get_server_certificate(hostport)
+    except Exception as e:
+        message = "FDG's cert_discovery Couldn't get cert: {0}".format(e)
+        log.info(message)
```
praksinha: Change this to `log.error`

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,168 @@
...
+    else:
+        host = host_ip['host_ip']
+    if host_port == "":
+        port = chained['host_port']
+    else:
+        port = host_port['host_port']
+    cert = load_certificate(host, port)
+    if not cert['result']:
+        message = "FDG's cert_discovery - cert details not found"
+        log.info(message)
```
praksinha: Change this to `log.warn`

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,168 @@
...
+    2. Alternately, execute hubble fdg.fdg <path to fdg profile> to run this module via cmd.
+"""
+import OpenSSL
+import ssl
+from socket import setdefaulttimeout
+import logging
+from datetime import datetime
+log = logging.getLogger(__name__)
+
+setdefaulttimeout(3)
```
praksinha: Default timeout should be made configurable either via profile or config. Adding this option in profile would make it more flexible. If this flag is not set explicitly anywhere in profile or config and then we can set it to some default like `3`.

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,168 @@
...
+Flexible Data Gathering: cert_discovery
+=============================
+Intention -
+This fdg module allows connecting to open ports on a system and retrieving
+certificate details that might be attached on those ports.
+
+Testing -
+    1. Configure this module through Hubble's schedule using the following configuration
+        fdg_cert_discovery:
+        function: fdg.fdg
```
praksinha: Is indentation of this example block correct ?

MoodyMudit: Fixed

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: Actually this is not a user supplied value. This is more a preventive action on our side to exit the code if it gets stuck. We don't want the user to either supply or edit the value.

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: done

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: Done

hubblestack/extmods/fdg/cert_discovery.py
praksinha: User would never specify timeout duration. It would be Admins who manage Hubble profiles who should have flexibility to override default timeout.

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: Done

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: I meant to give out an information here. Will change the text to sound more like an info.

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: Handled this.

hubblestack/extmods/fdg/cert_discovery.py
```diff
@@ -0,0 +1,188 @@
...
+    chained
+        The value chained from the previous call.
+
+    chained_status
+        The status returned by the chained call.
+    """
+    if host_ip == "":
+        host = str(chained.get('host_ip', ''))
+    else:
+        host = str(host_ip.get('host_ip', ''))
```
praksinha: You still have if else block :)  This can be done in one line expression:  `host = str(chained.get('host_ip', host_ip.get('host_ip', '')))`

Same for port

hubblestack/extmods/fdg/cert_discovery.py
jettero: Remember too that Hubble is single threaded, so if you're checking 10 hosts ... and for argument's sake, let's say they're all super busy; and your timeout is say 15s ...

Then during the schedule item (fdg) we'd check the 10 hosts at 15s/check giving 150s for just that one schedule item.

(For reference, we have this problem in the HEC too... There we decided several failures in a row should flag an outage condition and simply queue to disk.)

MoodyMudit:
@fossam Removed leading spaces in SANs.

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: Restructured the inputs

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: Restructured the inputs

hubblestack/extmods/fdg/cert_discovery.py
MoodyMudit: Configured timeout via the corresponding fdg profile.

fossam:
@MoodyMudit thanks for fixing the blank space, that looks a lot better.

And thank you for including the tests also. 

Is calling the module directly from the `main` supported (or not)? I think it would be helpful if this is specified in the doc strings somewhere.

fossam:
@MoodyMudit make sure to pull my changes to avoid conflicts.

Also, I noticed that the unit test still reference the module name that was used before you renamed the file. Those also should be fixed.

tests/unittests/test_cert_discovery.py
```diff
@@ -0,0 +1,132 @@
+# coding: utf-8
```
praksinha: It would be great if could rename this test case file as well to avoid any confusion in future.

jettero:
I clicked approve, but I just noticed integration testing failed...

```
04:16:19  ==================================== ERRORS ====================================
04:16:19  ___________ ERROR collecting tests/unittests/test_cert_discovery.py ____________
04:16:19  ImportError while importing test module '/opt/jenkins/workspace/hubble_PR-734/tests/unittests/test_cert_discovery.py'.
04:16:19  Hint: make sure your test modules/packages have valid Python names.
04:16:19  Traceback:
04:16:19  tests/unittests/test_cert_discovery.py:8: in <module>
04:16:19      import hubblestack.extmods.fdg.ssl_certificate
04:16:19  hubblestack/extmods/fdg/ssl_certificate.py:48: in <module>
04:16:19      import OpenSSL
04:16:19  E   ImportError: No module named OpenSSL
04:16:19  - generated html file: file:///opt/jenkins/workspace/hubble_PR-734/tests/unittests/output/pytest.html -
04:16:19  
04:16:19  ---------- coverage: platform linux2, python 2.7.16-final-0 ----------
04:16:19  Coverage HTML written to dir tests/unittests/output/coverage
04:16:19  
04:16:19  !!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
04:16:19  =========================== 1 error in 5.09 seconds ============================
```

I wonder if package builds will need an openssl kick in the pants.

PR #738, opened 2019-11-08
fix: sourcetype = 'hubble_fdg_' + fdg_file

jettero:
Apparently we do want to extend the sourcetype name from `hubble_fdg` to `hubble_fdg_filename` but let's choose to do that without the `salt://` protocol and without the `.fdg` file extension; while we're at it, make sure to replace all non alphanumerics with an underscore.

### DIFF
--- a/hubblestack/extmods/returners/splunk_fdg_return.py
+++ b/hubblestack/extmods/returners/splunk_fdg_return.py
@@ -177,7 +177,7 @@ def _build_args(ret):
 
     return args
 
-def _file_url_to_sourcetype_tag(filename):
+def _file_url_to_sourcetype(filename, base='hubble_fdg'):
     """ attempt to turn a file URL into a sourcetype extension description
         e.g.:
         'salt://fdg/interesting.operation.fdg'
@@ -186,13 +186,21 @@ def _file_url_to_sourcetype_tag(filename):
         (intended for internal use by _generate_payload() to append to the
         default sourcetype)
     """
-    if filename.startswith('salt://'):
-        filename = filename[7:]
-    if filename.startswith('fdg/'):
-        filename = filename[4:]
-    if filename.endswith('.fdg'):
+    if re.search(r'^\w+://', filename):
+        filename = filename.split('://', 1)[1]
+    if re.search(r'\.fdg$', filename):
         filename = filename[:-4]
-    return re.sub(r'[^\w\d]+', '_', filename)
+    def _no_dups(x):
+        sf = re.split(r'[^a-zA-Z0-9]+', x)
+        last = None
+        for item in sf:
+            if item == last:
+                continue
+            if not item:
+                continue
+            yield item
+            last = item
+    return '_'.join( _no_dups(base + '_' + filename) )
 
 def _generate_payload(args, fdg_args, cloud_details, opts, index_extracted_fields):
     """
@@ -203,9 +211,7 @@ def _generate_payload(args, fdg_args, cloud_details, opts, index_extracted_field
     payload = {'host': args['fqdn'], 'index': opts['index']}
     if opts['add_query_to_sourcetype']:
 
-        payload.update({'sourcetype': "{sourcetype}_{extension}".format(
-            sourcetype=opts['sourcetype'],
-            extension=_file_url_to_sourcetype_tag(fdg_file))})
+        payload.update({'sourcetype': _file_url_to_sourcetype(fdg_file, opts['sourcetype'])})
     else:
         payload.update({'sourcetype': opts['sourcetype']})
 
--- a/hubblestack/extmods/returners/splunk_fdg_return.py
+++ b/hubblestack/extmods/returners/splunk_fdg_return.py
@@ -43,7 +43,7 @@ gateway is not defined.
               - product_group
 """
 import socket
-
+import re
 import json
 import logging
 from hubblestack.hec import http_event_collector, get_splunk_options, make_hec_args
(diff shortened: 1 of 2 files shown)


hubblestack/extmods/returners/splunk_fdg_return.py
```diff
@@ -177,6 +177,14 @@ def _build_args(ret):
 
     return args
 
+def _file_url_to_sourcetype_tag(filename):
```
Tenebriso: Could this function have a short docstring explaining its purpose? It's clear what it does, it would be a small quality of life thing

jettero: fair point. pushed patch to that effect

fossam:
@jettero Thanks for quick turn around on this:
There is one corner case that isn't covered - when fdg is invoked from a command line it adds an additional underscore to sourcetype:

For example, if my fdg file had a following path: `/opt/tmpfdg_test/2fdg_cert_test.fdg`
then sourcetype would become `hubble_fdg__opt_tmpfdg_test_2fdg_cert_test`

Invoking fdg from command line is not widely used so if it is overly cumbersome to fix I'd be ok to leave it as it is.

jettero:
@fossam well, that's the question. I specifically trimmed `salt://fdg/blah.fdg` to `blah` because that's the most likely use case, and I figured if we invoked something from the command line, we wouldn't want to use the same sourcetype; because it really may not be the same source.

So the question isn't whether it's a lot of trouble, the question is what's the right result?

I could pop off that trailing filename, remove the extension and just roll with that, or we could come up with a long gruesome strategy.

fossam:
The "right" result is when both options result in the same outcome - have paths that start with `salt://fdg/` work as they are at this point, but when we have an absolute path then instead of the double underscore just have a single one. 

Say the same file is referenced by `salt://fdg/blah.fdg` or `/tmp/blah.fdg`. They would respectively become `hubble_fdg_blah` and `hubble_fdg_tmp_blah`.

jettero:
I don't understand why `salt://fdg/some_file.fdg` should work out to be the same as `/home/jettero/.fdg/some_file.py` -- but at least if I say the sourcetype extension is the filename minus decorations it'll be consistent.

fossam:
This is clever. 
Here's what happened: I had made an fdg profile file to test the certificate module, but couldn't think imaginatively enough and named it `fdg_cert_test.fdg`. In Splunk the sourcetype became `hubble_fdg_cert_test`. After a bit I figured out what had happened and I started to think back on conversations on Slack and now I am unsure what to say.

Should my ability to step right into the first surprising outcome be ignored or are other common cases where the outcome could as surprising - like `.../certs/certs_tests.fdg` or `.../testing/testing_certs.fdg`?

jettero:
I don't think it matters very much. I think it's a strange quirk of Hubble that it alters the source type based on the query anyway. I think with most types of injest, for the most part, the sourcetype would be `hubble_fdg` (or `hubble_nova` or whatever) and there'd be some kind of subfield that'd differentiate the data.  So your search would be `index=hubble_something sourcetype=hubble_fdg subtype_thing=differentiator` 

Normally the sourcetype-append-text is going to come from something in a signed data repo anyway and no matter what we do to this sourcetype guess generator, someone will find a way to make it screwy from the commandline.

fossam:
> Normally the sourcetype-append-text is going to come from something in a signed data repo anyway and no matter what we do to this sourcetype guess generator, someone will find a way to make it screwy from the commandline.

I agree about the command line being a bit corner case. Just to clarify my most recent test - those filenames were not invoked from a command line, they were from a top file and run from a schedule just like they'd be used in most often. But that still opens up a case where one wouldn't necessarily expecting folder name and filename having this issue..

fossam:
> I just can't tollerate hubble_fdg_fdg_test as a sourcetype that we _normally_ generate from fdg.

For those unaware, can you explain the issue with this?

jettero:
It's inconsistent with the other modules that do this silly renaming thing.
You don't see any `hubble_nebula_var_cache_hubble_hubblestack_nebula_v2_fifteen_minute` sourcetypes.

Nebula just grabs a name for a dict key in the file and rolls with it -- no matter where the file is located. The only reason we're having this discussion is because the fdg functions aren't named, so we're using the filename as a kindof placeholder for a janky sourcetype renaming (that we arguably shouldn't be doing anyway).

I dislike `hubble_fdg_fdg_test` because it's ugly and (most likely) every single fdg file is going to be in that directory in the repo so the extra fdg in the name isn't adding anything useful it's just embarrassing to look at. I'm also convinced if we really want to reveal the full fdg filepath for the event, we should certainly do so, but it should be more like this:

`index=hubble_whatever sourcetype=hubble_fdg fdg_file=fdg/test.fdg`

... that's really the only logical thing to do in splunk imo.

But nevermind all that. We already rename sourcetypes all over the place, so we may as well continue. But I see absolutely no reason we should reveal the full repo filepath in the sourcetype. We can add a field for that if we really need it. This is just a nickname we're appending to the sourcetype afterall.

If we really really needed this to be 100% accurate and reveal the full path of the fdg file; then surely we also have to go back to nebula and teach it to reveal the full filepath of the osquery files too? right? Cuz those nicknames don't have any path info in them at all!! \o/

fossam:
There are two issues at play: 
* nebula and fdg (I don't think I can compare the two because as you pointed out they use different naming conventions.)
* Splunk - out of the fields from fdg, only `sourcetype` is an index time extracted field. Using that to select different events is important for search performance.

As far as ugliness goes, I agree that it's not pretty. But the issue is that when folders come to play and the fdg profile files are separated to folders then it will be quite a surprising result. Even the very first examples of fdg that we had used that way or organizing the files and this would have caught that. I'll share the link to that example with you over Slack.

fossam:
Tested it and passed on affected configurations.

PR #740, opened 2019-11-18
simple patch. The default of s3.https_enable should be True

jettero:


fossam:
Great catch.

What would be a good test to make sure it comes out positive in the end of the package building process?

fossam:
Question:
In the same vein should: https://github.com/hubblestack/hubble/blob/bb5027df89af944a2b418ae195bd0cb56c605ffb/hubblestack/extmods/fileserver/s3fs.py#L332-L334 also be True?

fossam:
Looks good. Before we move on, can you go over the details of how you tested these two items working? PR comment is fine, for this one.

jettero:
I added a test before these patches that match the old behavior, then modified the tests to show the new behavior. I did not actually test the s3fs; but we did have that helpful helper that tested the `s3.https_enable` option for us.

hubblestack/extmods/fileserver/s3fs.py
```diff
@@ -324,29 +324,23 @@ def _get_s3_key():
...
+        'service_url': None,
+        'keyid': None,
+        'key': None,
+    }
+    ret = dict()
+    for k in defaults:
+        s3k = 's3.' + k
+        ret[k] = __opts__.get(s3k, defaults[k])
+
+    ret['kms_keyid'] = __opts__.get('aws.kmw.keyid')
```
praksinha: @jettero Shouldn't this be `aws.kms.keyid` instead of `aws.km**x**,keyid` ? I understand this is how it would referenced before as well but even that seems wrong, just have a look at `if` statement.

praksinha:
Apart from that minor comment, changes look good to me.

PR #735, opened 2019-10-24
Updating osqueryd safe permission script fix a small bug that doesn't let osqueryd start in windows

MoodyMudit:

PR #743, opened 2019-11-20
attempting to backport features for 3.0.8 release

jettero:
There's still a few more of these "To Backport" items I'd like to make "Backport Complete" ; a couple aren't even approved yet.

So far, this is:

* PR #714
* PR #717
* PR #727
* PR #729
* PR #730
* PR #731
* PR #738
* PR #734
* PR #735
* PR #740


fossam:
@jettero I don't know if intentional, but the diff items with fdg's certificate changes look wrong - the field names are not right, etc.
If this is intentional and will go away then disregard.

hubblestack/extmods/fdg/osquery.py
```diff
@@ -51,6 +51,7 @@ def query(query_sql, osquery_args=None, osquery_path=None, format_chained=True,
         osquery_args = []
     return _osquery(query_sql, args=osquery_args, osquery_path=osquery_path)
 
+    return _osquery(query_sql, args=osquery_args, osquery_path=osquery_path)
```
praksinha: Seems like a duplicate return statement. @jettero  Can you verify this ?

MoodyMudit:
I don't think this contains the latest commit for PR - https://github.com/hubblestack/hubble/pull/734

jettero:
> I don't think this contains the latest commit for PR - #734

Agreed. I think I fixed it. Please double check.

hubblestack/extmods/fdg/osquery.py
jettero: Fixed, thanks. Fallout from the very complicated backport. The forward port is going to be just as bad...

hubblestack/extmods/returners/splunk_fdg_return.py
```diff
@@ -41,33 +41,45 @@
...
+HTTP_EVENT_COLLECTOR_DEBUG = False
 
 log = logging.getLogger(__name__)
 
 
 def returner(ret):
+    """
+    Get fdg data and post it to Splunk
+    """
+    data = ret['return']
```
MoodyMudit: If the 'ret' variable is null, this might throw a keyError.

hubblestack/hec/obj.py
```diff
@@ -311,9 +387,21 @@ def _send(self, *payload):
...
+                else:
+                    log.info('flagged as having an outage -- but it is time for a recheck')
+                    server.outage.checking()
             try:
+                # Remember that we tried to send this
+                meta_data['send_attempts'] += 1
                 r = self.pool_manager.request('POST', server.uri, body=data, headers=self.headers)
                 server.fails = 0
+                if server.outage:
+                    server.outage = False
```
MoodyMudit: The outcome of the above 2 statements is that the value of server.outage is set to False here, no matter what the initial value was. Was this the intent?

PR #746, opened 2019-11-22
3.0 fix

MoodyMudit:
Pinning azure dependencies because latest dependencies are not compatible with each other.


MoodyMudit:
I agree.  We should just check the latest current release of all dependencies and pin them.

PR #748, opened 2019-11-26
Clean unittest

Tenebriso:
I think the main requirement in terms of style for unittests should be readability and the guidelines that I think would help write better unittests are:
- tests should be named `test_<function_name>_<does_this>_<when_this>`
- one concept per test, the _one assert per test_ rule is too restrictive but modularity would work beautifully
- be explicit: although `assert status` would work just fine, we expect status to be True so the test should be verbose about it
- the `expected_ret` should be the first part of a comparison, for better readability in case of failure
- an error message would be great for asserts but no need to go overboard
They're up for debate and feedback is more than welcome. 

Something I am unsure of:

-  Does it make more sense to have an `expected_ret` and test `assert expected_ret == ret` instead of a plain `assert ret is True`? 

The tests in test_process are fairly basic and easy, more complicated tests I think will call for more complex guidelines.

### DIFF
--- a/tests/unittests/test_process.py
+++ b/tests/unittests/test_process.py
@@ -17,55 +17,79 @@ class TestProcess():
     Class used to test the functions in ``process.py``
     """
 
-    def test__compare_invalidComp_raiseException(self):
+    def test__compare_raises_exception_if_arguments_have_invalid_type(self):
         """
         Test that given invalid ``comp``,
         the function raises an ArgumentValueError exception
         """
-        with pytest.raises(ArgumentValueError) as e_info:
+        with pytest.raises(ArgumentValueError):
             hubblestack.extmods.fdg.process._compare('foo', 1, 2)
 
-    def test__compare_geCompt_validReturn(self):
+    def test__compare_returns_correctly_with_ge_comparator(self):
         """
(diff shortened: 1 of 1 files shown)


jettero:
> Does it make more sense to have an expected_ret and test assert expected_ret == ret instead of a plain assert ret is True?

Depends on the context. If it matters that `ret` is `True` rather than simply `bool(ret) is True`, then probably `assert ret is True` is better. If it doesn't matter, then it doesn't matter. I realize this is tautological, but seriously though. It only matters when it matters.

Or, if the question is about having an explicit `expected_ret` var? I submit that probably adds extra lines that the parser optimizer will simplify away anyway. It could subtract from readability and make things more tedious to maintain and write. On the other hand, if it was structured just right (eg, expect returns table dictionary or something), it could make everything easier.

I don't think we should have a formal rule on this. It really depends on the situation.

Tenebriso:
> These are great ideas.
> 
> An aside though: I'd hate to accidentally make testing even less fun. If someone submits tests that aren't formatted quite right, we should be pretty forgiving about that.

Definitely, I don't think we should enforce any of these, they're here to help when you want to do some testing and don't know where to start

PR #752, opened 2019-12-04
platform.linux_distribution() & platform.dist() replaced with distro.linux_distribution()

daveed12:
Python 3.5 deprecated platform.linux_distribution() and the distro package is the recommended replacement. Its predecessor function platform.dist() was already deprecated since Python 2.6 and both will be removed in Python 3.8. More information can be found [here](https://bugs.python.org/issue1322)


daveed12:
Resubmitting PR after being added to Adobe Github org

PR #753, opened 2019-12-05
Adding support for aws IMDSv2

MoodyMudit:
IMDSv2 requires a session-based token to authenticate to the metadata service. Hubble will now generate a token and use it to fetch metadata values.


hubblestack/extmods/grains/cloud_details.py
```diff
@@ -35,8 +35,13 @@ def _get_aws_details():
     proxies = {'http': None}
 
     try:
+        ttl_header = {'X-aws-ec2-metadata-token-ttl-seconds': '21600'}
```
sumanmehta: I think we can reduce TTL value to 5*60 seconds.

MoodyMudit: Done

hubblestack/extmods/grains/cloud_details.py
```diff
@@ -35,8 +35,13 @@ def _get_aws_details():
     proxies = {'http': None}
 
     try:
+        ttl_header = {'X-aws-ec2-metadata-token-ttl-seconds': '300'}
+        token_url = 'http://169.254.169.254/latest/api/token'
+        token_request = requests.put(token_url, headers=ttl_header)
```
praksinha: You should specify `proxies=proxies` arg as well like we do in `requests.get` call and also specify a timeout similar to .get call. This is required as users might set proxy globally using environment variable. Also timeout helps in safeguarding against corner cases where this URL may not respond.

hubblestack/extmods/grains/cloud_details.py
```diff
@@ -35,8 +35,13 @@ def _get_aws_details():
...
+        token = token_request.text
+        aws_token_header = {'X-aws-ec2-metadata-token': token}
```
praksinha: This is new feature released by AWS and although I trust this will almost always work like current metadata APIs but we should still handle scenarios where `token` might be empty due to any reason (why I'm not sue of yet). Also, if token is empty then do we need to switch to older API which works without authentication headers ?

hubblestack/extmods/grains/cloud_details.py
MoodyMudit: Done.

hubblestack/extmods/grains/cloud_details.py
MoodyMudit: I checked this. If the IMDSv1 is enabled then both a valid token and ''(blank) token works. If the IMDSv1 is disabled, then only a valid token will work. So in a way, AWS has put a fallback for empty token value.

MoodyMudit:
Will raise this PR for 3.0 branch instead of develop. Closing from here.

PR #756, opened 2019-12-11
adding support for IMDSv2 for aws

MoodyMudit:
IMDSv2 requires a session-based token to authenticate to the metadata service. Hubble will now generate a token and use it to fetch metadata values.

### DIFF
--- a/hubblestack/extmods/grains/cloud_details.py
+++ b/hubblestack/extmods/grains/cloud_details.py
@@ -36,8 +36,13 @@ def _get_aws_details():
     proxies = { 'http': None }
 
     try:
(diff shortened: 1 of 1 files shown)


MoodyMudit:
I think it is fine to get stuck(for 3 seconds) during testing, rather than get stuck in production indefinitely. We wouldn't want to remove the timeout completely. Reducing it will only save us a second maybe.

jettero:
But my point is that on any bare metal host you're going to get two 3 second timeouts -- I guess just one since they're both in the same try/except. I wonder if before we hit network we should at least look for /var/lib/cloud/data or something.

I'm just thinking outloud though. I don't mean to hold up merging this.

PR #759, opened 2019-12-13
3.0.8 s3 fix

jettero:
The problem here seems to be that I failed to backport a cleanup of this code where the return type of get_s3_key changed. But I did backport a change that relied on that previous cleanup.


fossam:
Since the problem seems to be return values this might not be applicable but I'll ask anyway - would there be a log for certificate failure even if cache exists? This is to avoid the silent failure scenario.

Also, how can we find out which ca certificate file the requests library is using when running under pyinstaller?

jettero:
Yeah, before I merge, I think I'll add that. It should really have been generating an error before this.

PR #760, opened 2019-12-18
pinning salt-ssh to 2019.2.0

MoodyMudit:


jettero:
This is probably a good idea, but I wonder if we should pin to 2018.x.x somewhere. 2019.x.x is where the problems really start.

MoodyMudit:
I have tested this. This is working as expected. Explanation for pinning this version is as follows:
The latest version of salt has fixed a bug that reads a file's permission(which had permission 000) and returns '0000' , whereas earlier it returned <int>0. We want the hubble code to behave the same as before right now.

PR #776, opened 2020-01-08
fix bug in hubble.py

Tenebriso:
Python 3 doesn't allow for a dict's size to be modified during iteration. This PR fixes this issue in the hubble.py file.

### DIFF
--- a/hubblestack/extmods/modules/hubble.py
+++ b/hubblestack/extmods/modules/hubble.py
@@ -333,9 +333,12 @@ def _run_audit(configs, tags, debug, labels, **kwargs):
     for failure_index in reversed(sorted(set(failures_to_remove))):
         results['Failure'].pop(failure_index)
 
+    to_remove = set()
     for key in results.keys():
         if not results[key]:
-            results.pop(key)
+            to_remove.add(key)
+    for key_to_remove in to_remove:
+        results.pop(key_to_remove)
 
     return results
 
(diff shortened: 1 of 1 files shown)


daveed12:
Did a little research like you asked @Tenebriso . It seems that the keys method in Python 3 returns an iterator instead of a list. Instead of using a set we could use the list constructor to get a copy of the keys. Let me know what you think.
```
for key in list(results):
        if not results[key]:
             results.pop(key)
```

Tenebriso:
> Did a little research like you asked @Tenebriso . It seems that the keys method in Python 3 returns an iterator instead of a list. Instead of using a set we could use the list constructor to get a copy of the keys. Let me know what you think.
> 
> ```
> for key in list(results):
>         if not results[key]:
>              results.pop(key)
> ```

That does look cleaner, thanks for looking it up

Tenebriso:
> Did a little research like you asked @Tenebriso . It seems that the keys method in Python 3 returns an iterator instead of a list. Instead of using a set we could use the list constructor to get a copy of the keys. Let me know what you think.
> 
> ```
> for key in list(results):
>         if not results[key]:
>              results.pop(key)
> ```

Thank you for looking it up, that does look clean!

Tenebriso:
The build on this is failing, I will close it and reopen it on top of the newest changes to fix it. I have tried pulling the newest changes but I can't push them because of an issue with my local git config.

PR #780, opened 2020-01-15
data sign [py3 rebase] and some additional merge forward

jettero:
I rebased the data-sign PR onto py3 branch to make sure it didn't cause conflicts (rather than having to merge develop forward again). It didn't actually require any changes, but; it seemed easier to PR a clean history.


hubblestack/utils/signing.py
```diff
@@ -0,0 +1,474 @@
...
+            # we mark it with the same status as the MANIFEST it self --
+            # presumably it's signed (STATUS.VERIFIED); but perhaps it's only
+            # UNKNOWN or even FAIL.
+            ret[vfname] = ret[mfname]
+        else:
+            # We do have a MANIFEST entry and it doesn't match: FAIL with or
+            # without a matching SIGNATURE
+            ret[vfname] = STATUS.FAIL
+    # fix any normalized names so the caller gets back their specified targets
+    for k,v in xlate.iteritems():
```
Tenebriso: I think this fails in python3, would need to be `.items()`

jettero: Yeah, I have to bring in other changes too. I'll get this tomorrow prolly.

jettero:
doh

I rebased to develop this morning before I regenerated this branch. So this is rebased to develop, then rebased to py3, which is causing a merge of sorts.

The additional py3 changes were annoying to write, so hopefully I can keep them as is. Also, I had to fix various things in the pulsar test -- I did wonder why that was showing up in this rebase but didn't reason out the problem was a previous accidental merge.

We'll want to merge these things forward. It's a little silly to combine it into this PR, but I'm fine with it if you are @Tenebriso.

PR #783, opened 2020-01-22
S3 fix limit

daveed12:
S3 cache expire can now be configured through the hubble stack config('s3.cache_expire'). Default cache expire time increased from 30 seconds to 1800 seconds(30 minutes). Also handles s3.query return DELETE request properly.

### DIFF
--- a/hubblestack/extmods/fileserver/s3fs.py
+++ b/hubblestack/extmods/fileserver/s3fs.py
@@ -356,10 +356,10 @@ def _init():
     """
     cache_file = _get_buckets_cache_filename()
     s3_key_kwargs = _get_s3_key()
-    cache_expire_time = s3_key_kwargs['cache_expire'] if s3_key_kwargs.get('cache_expire', None) else S3_CACHE_EXPIRE
-    exp = time.time() - float(cache_expire_time)
+    cache_expire_time = float(s3_key_kwargs['cache_expire'] if s3_key_kwargs.get('cache_expire', None) else S3_CACHE_EXPIRE)
+    exp = time.time() - cache_expire_time
 
-    log.info('S3 cache expire time is {}s'.format(cache_expire_time))
+    log.info('S3 cache expire time is %ds', cache_expire_time)
     # check mtime of the buckets files cache
     metadata = None
     try:
@@ -445,7 +445,7 @@ def _refresh_buckets_cache_file(cache_file):
                                         https_enable=s3_key_kwargs['https_enable'],
                                         params={'marker': marker})
             if not tmp:
-                return
+                return None
 
             headers = []
             for header in tmp:
--- a/hubblestack/extmods/fileserver/s3fs.py
+++ b/hubblestack/extmods/fileserver/s3fs.py
@@ -349,6 +349,7 @@ def _get_s3_key():
 
     return ret
 
+
 def _init():
     """
     Connect to S3 and download the metadata for each file in all buckets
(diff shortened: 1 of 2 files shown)


hubblestack/extmods/fileserver/s3fs.py
```diff
@@ -354,8 +355,11 @@ def _init():
     specified and cache the data to disk.
     """
     cache_file = _get_buckets_cache_filename()
-    exp = time.time() - S3_CACHE_EXPIRE
+    s3_key_kwargs = _get_s3_key()
+    cache_expire_time = s3_key_kwargs['cache_expire'] if s3_key_kwargs.get('cache_expire', None) else S3_CACHE_EXPIRE
+    exp = time.time() - float(cache_expire_time)
 
+    log.info('S3 cache expire time is {}s'.format(cache_expire_time))
```
Tenebriso: Shouldn't this be something like `log.info('S3 cache expire time is %ds', cache_expire_time)` to avoid `logging-format-interpolation`?

hubblestack/extmods/utils/s3.py
```diff
@@ -250,6 +250,9 @@ def query(key, keyid, method='GET', params=None, headers=None,
         return 'Saved to local file: {0}'.format(local_file)
 
     if result.status_code < 200 or result.status_code >= 300:
+        if err_code == 'SlowDown':
+            log.warning('{}, {}'.format(err_code, err_msg))
```
Tenebriso: This too, with `%s` instead?

hubblestack/extmods/fileserver/s3fs.py
```diff
@@ -440,6 +444,9 @@ def __get_s3_meta(bucket, key=s3_key_kwargs['key'], keyid=s3_key_kwargs['keyid']
                                         path_style=s3_key_kwargs['path_style'],
                                         https_enable=s3_key_kwargs['https_enable'],
                                         params={'marker': marker})
+            if not tmp:
+                return
```
Tenebriso: Let's have this explicitly `return None`, to avoid `inconsistent returns` (a return statement returns something, the other returns empty) even though it still technically returns None

daveed12:
@Tenebriso I'm all about shooting for that high pylint score. I'll make those changes.

hubblestack/extmods/utils/s3.py
```diff
@@ -250,6 +250,9 @@ def query(key, keyid, method='GET', params=None, headers=None,
         return 'Saved to local file: {0}'.format(local_file)
 
     if result.status_code < 200 or result.status_code >= 300:
+        if err_code == 'SlowDown':
+            log.warning('%s, %s', err_code, err_msg)
```
fossam: Currently during the main daemon loop the raised exception would generate the log (below) that is easy to search for.
Since this is to change, couple of questions: 
- How would an example log look like if a 503 is hit? Can we prepend a string like the one below that helps to identify that it's a matter of s3?
- Will the log end up being sent to Splunk ?
- Would it make sense to handle other error codes to avoid a crash on start up? For example, I think ServiceUnavailable, RequestTimeTooSkewed, RequestTimeout, OperationAborted, InternalError can be eventually recoverable or may be intermittent. (https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html)

daveed12: I can prefix the log with a string that will help us identify its an s3 issue. Currently, it would just say 'Slowdown, Please reduce your request rate' if that 503 error were to occur. In terms of being sent to Splunk, that would only be through CloudTrail & Object level logging on s3 or access level logging(ListObject api calls). Some of these errors would make sense to not raise an exception and handle that exception properly.

PR #784, opened 2020-01-28
384 fileperms

jettero:
if the configfile doesn't exist, setting file perms to 384 (aka `0o600`, aka `rw-------`) will raise a FileNotFoundError.

This is a reasonable fix for that.


fossam:
There seems to be a **whole** lot more going on than fixing permission issues.

jettero:
yeah, this is wrong... I'm going to close and re-open

PR #787, opened 2020-01-28
git describe update

jettero:
Because the `v3.0.8` and `v3.0.9` tags are unsigned, `git describe` is giving build info of, "`v3.0.5-573-g03e55c8`" ... while technically correct... `v3.0.9` is `03e55c8` and it's `573` commits after `v3.0.5` ... this isn't really a desirable description.

I propose we use `git describe --dirty --always --tags` so the build info is always detailed and based off the actual current tags.


jettero:
If we decide to use this, we should probably apply to 3.0 and 4.0 as well... I wanted to collect some thoughts first.

fossam:
Can you provide examples of how this will look with the new command line?

PR #790, opened 2020-01-30
s3-fix merge into 4.0 branch

jettero:
This is already in develop, but there were conflicts to resolve merging develop to make 4.0 branch. Until develop *is* the py3/4.0 branch, cherry-pick merges feel like less work.

### DIFF
--- a/hubblestack/extmods/utils/s3.py
+++ b/hubblestack/extmods/utils/s3.py
@@ -207,7 +207,7 @@ def query(key, keyid, method='GET', params=None, headers=None,
     if os.environ.get('MOCK_SLOW_DOWN'):
         result.status_code = 503
         err_code = 'SlowDown'
-        err_msg = 'slow down you jerks'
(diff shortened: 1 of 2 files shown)


hubblestack/extmods/utils/s3.py
```diff
@@ -203,6 +204,11 @@ def query(key, keyid, method='GET', params=None, headers=None,
             err_code = 'http-{0}'.format(result.status_code)
             err_msg = err_text
 
+    if os.environ.get('MOCK_SLOW_DOWN'):
+        result.status_code = 503
+        err_code = 'SlowDown'
+        err_msg = 'slow down you jerks'
```
Tenebriso: I don't think this is an appropriate error message, could we change it to something more informative?

jettero: Yeah, that seems reasonable. Although it's worth nothing, this message is only for manual testing and it's always a fatal condition in the s3fs.

PR #795, opened 2020-02-04
use pyenv and allow alt checkouts

jettero:


pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -136,12 +136,22 @@ RUN mkdir -p "$LIBGIT2TEMP" \
...
+ENV PATH=$PYENV_ROOT/bin:$PATH
+RUN umask 022 \
+ && curl -s -S -L "$PYENV_INSTALLER_URL" -o /usr/bin/pyenv-installer \
+ && chmod 0755 /usr/bin/pyenv-installer \
+ && /usr/bin/pyenv-installer \
+ && eval "$(pyenv init -)" \
+ && env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install $PYENV_VERSION \
+ && pyenv global $PYENV_VERSION
+
+RUN eval "$(pyenv init -)" \
```
fossam: can the earlier `COPY pyinstaller-requirements.txt /` be moved right before this line so that requirements file would stay together with the usage?

jettero: sure

pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -151,7 +161,7 @@ RUN yum install -y ruby ruby-devel rpmbuild rpm-build rubygems gcc make \
 #pyinstaller start
 #commands specified for ENTRYPOINT and CMD are executed when the container is run, not when the image is built
 #use the following variables to choose the version of hubble
-ENV HUBBLE_CHECKOUT=v3.0.8
+ARG HUBBLE_CHECKOUT=v3.0.8
 ENV HUBBLE_VERSION=3.0.8
```
MoodyMudit: Please change the HUBBLE_VERSION to 4.0.0 in all dockerfiles

PR #803, opened 2020-02-19
Changing file offset handling workflow

devagupt:
During grain refresh activity, the 'nebula_osquery' module gets refreshed and as a result the offset stored for file in memory gets reset every time. As a result duplicate logs are going in splunk. Following are the changes in new behaviour:

- File Offset for log files is stored in cache files instead of memory. 
- Apart from storing the offset of log file, we also store Initial CRC (CRC for first 256 bytes) and last CRC (CRC for last 256 bytes) which we validate before returning the offset.
- In case if there is any mismatch in the CRC, file rotation or exception cases, we return the offset as 0.


hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -34,6 +34,8 @@
 import os
 from os import path
 import yaml
+import zlib
+import traceback
```
MoodyMudit: Are these imports already installed? Do we need to install them by specifying them in pyinstaller_requirements.txt?

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -1390,32 +1391,100 @@ def _parse_log(path_to_logfile,
...
             else:
                 log.error('Unable to open log file for reading: {0}'.format(path_to_logfile))
     else:
         log.error("Log file doesn't exists: {0}".format(path_to_logfile))
 
     return event_data
 
-
-def _set_cache_offset(path_to_logfile, offset):
+def _set_file_offset(path_to_logfile, offset):
```
MoodyMudit: Ensure that the previous function wasn't called from other places.

hubblestack/extmods/modules/nebula_osquery.py
```diff
@@ -1390,32 +1391,100 @@ def _parse_log(path_to_logfile,
...
+    offset = 0
+    try:
+        log_filename = os.path.basename(path_to_logfile)
+        offsetfile = os.path.join(__opts__.get('cachedir'), 'osqueryd', 'offset', log_filename)
+        if not os.path.isfile(offsetfile):
+            log.info("Offset file: {0} does not exist. Returning offset as 0.".format(offsetfile))
+        else:
+            with open(offsetfile, 'r') as file:
+                offset_data = json.load(file)
+            offset = offset_data['offset']
```
MoodyMudit: fetching values in this form from a dictionary is error prone. This will throw an exception if 'offset' key is not found, and could break Hubble's workflow. Instead fetch with **offset_data.get('offset')**. Similar for other places.

hubblestack/extmods/modules/nebula_osquery.py
devagupt: Checked

hubblestack/extmods/modules/nebula_osquery.py
devagupt: fixed this issue. Good to know.

hubblestack/extmods/modules/nebula_osquery.py
devagupt: These libraries are part of core python libraries.

fossam:
@jettero I was primarily the one waiting for this bug fix and would prefer this to get merged to 3.0 and have it also find a way to 4.0.

devagupt:
Testing strategy for this feature on windows and Centos7 build
- Verified CRC match feature by modifying the CRC for initial and last log line so that whole file should be read again.
- Offset is getting correctly picked up and written in the cache files
- Log file rotation should work as expected and no duplicate logs should be send during it.

hubblestack/extmods/modules/nebula_osquery.py
jettero: I really thought I posted those change requests days and days ago, but they were pending (not showing)...

fossam: @devagupt can more into depth on this issue, but the goal here is not to check file integrity - the goal is to know if the file we're continuing to send events from is the same file as before or has it been rotated and just happens to be again as big as the previous one. File integrity would not help as the file may grow between the checks.

Using a mechanism to validate parts of the content can also serve as indication of potential loss of collected data if Hubble was not the one that rotated it.

devagupt: @fossam It is correct that CRC mechanism is used to identify if we have missed sending any events to Splunk. If we found some deviations, then we are sending the whole log file again to ensure no events are missed. 
As 4.0 release deadline is approaching, we can pick up the changes proposed by @jettero in the future versions to use salt lib.

devagupt: @fossam It is correct that CRC mechanism is used to identify if we have missed sending any events to Splunk. If we found some deviations, then we are sending the whole log file again to ensure no events are missed.
As 4.0 release deadline is approaching, we can pick up the changes proposed by @jettero in the future versions to use salt lib.

praksinha:
As discussed over slack , @devagupt will incorporate changes requested by @jettero in develop branch.
Merging this PR to trigger builds.

PR #804, opened 2020-02-19
Osqueryd returner timestamp bug fix

devagupt:
- Changed the logic to fetch osquery results from current time to the timestamp of event.
- As a result now events will be published in splunk with their intended timestamp rather than time of ingestion of events.


jettero:
Interesting catch

PR #805, opened 2020-02-19
Adding exception handling in running schedules

devagupt:
- Currently if a schedule gets an exception it leads to subsequent schedules not running. Fixed this behaviour by adding exception handling while running schedules so that if an exception occurs in running a schedule, it gets logged and next schedule gets picked up.


jettero:
I've never noticed this behavior. I'm going to want to try to reproduce.

MoodyMudit:
We might not plan to release any more 3.0 builds. Let's make sure that this change is merged into develop and 4.0 branch as well.

devagupt:
> make

Will create a sepafrate PR for merging in develop branch as well.

jettero:
Please don't merge things into develop. We're trying to finish up 4.0 to merge to develop, that will just cause conflicts.

jettero:
I would say that we shouldn't merge this into 3.0 at all (except that it's clearly ready to go). The next release will be 4.0 so we should probably only merge it there.

PR #807, opened 2020-02-19
4.0 pkg fix

jettero:
Originally I took pyinstaller out of the equation, but ultimately put it back to keep the package size down.


fossam:
I would like to hold off on merging this until changes in this PR produce reliable packages for at least all of the Linux distros we need to run in.

fossam:
@jettero does this pr include all the pyinstaller and python3 related changes for a successful build?

fossam:
That's a lot of changes suddenly - can you explain what happened?

fossam:
I think d91d1a2 was part of another PR. (changes to azurefs) How come they're here also?

PR #808, opened 2020-02-19
some kind of conflict resolution typo most likely

jettero:
probably a typo from forward porting the rate limit fix into 4.0 branch.


fossam:
The typo is from https://github.com/hubblestack/hubble/commit/d12652c2113b8211a01a793c6533ef4673dd7ebf .

PR #809, opened 2020-02-21
profile signing and verification logging fixes

git-gud-1:
This commit formats log statements and dampens how often they are sent to Splunk.


fossam:
@git-gud-1 I wasn't aware earlier, but per https://github.com/hubblestack/hubble/pull/805#issuecomment-589401516 4.0 branch would have probably been the right choice.

I'll let Paul to weigh in on the right course of action.

jettero:
#810 seems to be this PR against 4.0 branch. Rejecting this one to avoid conflicts later.

PR #811, opened 2020-02-27
Filter bad host values on queue log messages

mew1033:
Found a hubble_log message that was setting a bad value in the `host` field. This should fix that. Can we make sure this is merged into 4.0/develop?


jettero:
How was that not already in there?? I guess originally stdrec was everything except the hostname??

PR #817, opened 2020-03-03
4.0 azure py3 fix and forward port of PR811

jettero:
'1a1b1c'.decode('hex') is now bytes.fromhex('1a1b1c'); conveniently, already bytes for b64encode()


jettero:
I'm also pulling along a commit that should have been forward ported already and is marked "forward port complete" ... #811

Apparently I did the work in my 4.0 branch at home and then failed to actually PR the change to the 4.0 branch in the cloud.

PR #825, opened 2020-03-20
Bug fix

goravsingal:
When the disk is full, and user cleaned up the directories. Hubble failed to create pid file again due to disk space unavailable, and raised an exception.

Since, it was an unhandled exception. It results in crashing of hubble process.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -170,7 +170,16 @@ def main():
         pidfile_count += 1
         if __opts__['daemonize'] and pidfile_count > pidfile_refresh:
             pidfile_count = 0
-            create_pidfile()
+            try:
+                # Fix when disk is full, and pid file gets deleted. 
+                # Hubble tries to create this file. Due to no-space. This result in exception, and process exits
+                # Fix is to put this create_pidfile in try-catch
+
+                create_pidfile()
+            except Exception as exc:
+                log.exception('Error in creating pid file. Probably disk is full: {0}'.format(exc))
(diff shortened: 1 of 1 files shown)


hubblestack/daemon.py
```diff
@@ -170,7 +170,16 @@ def main():
...
+            try:
+                # Fix when disk is full, and pid file gets deleted. 
+                # Hubble tries to create this file. Due to no-space. This result in exception, and process exits
+                # Fix is to put this create_pidfile in try-catch
+
+                create_pidfile()
+            except Exception as exc:
+                log.exception('Error in creating pid file. Probably disk is full: {0}'.format(exc))
+                # continue for further operations
+
```
praksinha: I think this might worsen the issue. Since now we are allowing Hubble to continue running and not create the pidfile, we can now have multiple Hubble processes. Current design is that when Hubble is started it checks if pidfile exists, if it exists and process is running then Hubble exits by saying Hubble is already running. Now what will happen is , another Hubble process can be started along-with existing process. This can have cascading effect user or some script is blindly restarting Hubble at some frequency resulting in many Hubble processes. In if disk is full, all of these Hubble processes probably won't be doing anything anyways and would impact system performance eventually.

In short, it's better to exit, just gracefully instead of exit without logging what happened.

praksinha: @goravsingal

goravsingal: Thanks for providing insights. This is indeed not good.

PR #834, opened 2020-03-27
4.0 sign script

jettero:
This change needn't be rolled into 4.0 builds, nor should it hold it up at all.

The changes to hubblestack.utils.signing are logging and whitespace and should not affect signing (and verifying in particular) in the slightest.

Mainly this is to provide the convenience script `contrib/sign-profiles.sh`


daveed12:
LGTM

PR #839, opened 2020-04-20
Adding osquery blacklisted flags

praksinha:
There are few osquery daemon flags that can be exploited if remote fileserver is compromised.
Creating a blacklist of such flags so that they can't be overridden using remote flagsfile in case fileserver is compromised.


MoodyMudit:
You will have to import the hubble_constants.py in daemon.py as well. Refer this - https://github.com/hubblestack/hubble/blob/develop/hubblestack/daemon.py#L36

PR #840, opened 2020-04-23
Adding a hubble grain that gives docker process details

MoodyMudit:


hubblestack/extmods/grains/docker_details.py
```diff
@@ -0,0 +1,63 @@
...
+      docker_details['installed'] = True
+      docker_details['running'] = _is_docker_process_running()
+
+    log.debug('docker_details = {0}'.format(docker_details))
+
+    grains['docker_details'] = docker_details
+
+    return grains
+  except Exception as e:
+    log.exception('The following exception occurred while fetching docker details {0}'.format(e))
```
goravsingal, line 36: In case of an exception, nothing is returned. Is it expected?

hubblestack/utils/osquery_lib.py
```diff
@@ -0,0 +1,41 @@
...
+
+__salt__ = {'cmd.run': salt.modules.cmdmod._run_quiet,
+            'cmd.run_all': salt.modules.cmdmod.run_all}
+
+log = logging.getLogger(__name__)
+
+def query(query_sql='', osquery_path=None, args=None, max_file_size=104857600, timeout=10000):
+  try:
+    if not query_sql:
+      return ''
```
goravsingal, line 14: Should return boolean False or null in case of these error cases.

If this is some practice of writing grains, then its ok.

hubblestack/utils/osquery_lib.py
```diff
@@ -0,0 +1,41 @@
...
+    if not osquery_path:
+      if not os.path.isfile(__grains__['osquerybinpath']):
+        log.error('osquery binary not found: %s', __grains__['osquerybinpath'])
+        return ''
```
goravsingal, line 18: same here, please check at every place.

hubblestack/utils/osquery_lib.py
```diff
@@ -0,0 +1,41 @@
...
+    # Run the command
+
+    res = __salt__['cmd.run_all'](cmd, timeout=timeout, python_shell=False)
+    if res['retcode'] == 0:
+      ret = json.loads(res['stdout'])
+      if len(ret) != 0:
+        for result in ret:
+          for key, value in result.items():
+            if value and isinstance(value, unicode):
+              return value
```
goravsingal, line 37: For queries returning multiple results. We are returning just first.
Here Example: dockerd, dockerd-current, dockerd-default

This case might fit here. But, this function is for running any query (as the name suggests).
What will happen if caller is expecting all such output.

fossam:
Why is this PR adding code from #839 in commit 714b7aa ?

MoodyMudit:
Closing PR, since this needs logical code changes.

PR #843, opened 2020-04-28
Docker details grain

MoodyMudit:
* Added a docker grains that lets Hubble know if docker is installed and is running.


hubblestack/daemon.py
```diff
@@ -788,6 +790,7 @@ def refresh_grains(initial=False):
 
     hubblestack.utils.stdrec.__grains__ = __grains__
     hubblestack.utils.stdrec.__opts__ = __opts__
+    hubblestack.utils.osquery_lib.__grains__ = __grains__
```
mew1033, line 793: Can we just add `global __grains__` to the util module?

hubblestack/daemon.py
```diff
@@ -33,12 +33,14 @@
...
 import hubblestack.hec.opt
 import hubblestack.utils.stdrec
+import hubblestack.utils.osquery_lib
 from hubblestack import __version__
 from croniter import croniter
 from datetime import datetime
 from hubblestack.hangtime import hangtime_wrapper
 import hubblestack.status
 import hubblestack.saltoverrides
+import hubblestack.hubble_constants
```
mew1033, line 43: Why do we need this import?

hubblestack/extmods/grains/docker_details.py
```diff
@@ -0,0 +1,76 @@
...
+    if salt.utils.platform.is_windows():
+      log.debug('This grain is only available on linux')
+      return docker_grains
+
+    docker_details = {}
+    docker_details['installed'] = _is_docker_installed(grains)
+    docker_details['running'] = False
+
+    if docker_details['installed']:
+      docker_details['running'] = _is_docker_process_running()
```
mew1033, line 26: Instead of using osquery, which adds overhead, let's just use salt's built in package/process checker.
```
docker_details['installed'] = bool(__salt__['pkg.version']('docker'))
docker_details['running'] = bool(__salt__['ps.pgrep']('docker'))
```

hubblestack/utils/osquery_lib.py
```diff
@@ -0,0 +1,41 @@
...
+Author - Mudit Agarwal (muagarwa@adobe.com)
+"""
+import logging
+import os
+import salt.modules.cmdmod
+import json
+
+__salt__ = {'cmd.run': salt.modules.cmdmod._run_quiet,
+            'cmd.run_all': salt.modules.cmdmod.run_all}
+
```
mew1033, line 12: Just add
```
global __grains__
```
here?

hubblestack/daemon.py
MoodyMudit: This is required so that this is available to nebula_osquery as an import.

hubblestack/extmods/grains/docker_details.py
MoodyMudit: I thought of using it, but salt is initialised after the grains are loaded. And so the salt module is unavailable in the grains.

hubblestack/utils/osquery_lib.py
MoodyMudit: This module is used as an import in one the docker_details grain. Since __grains__ is not yet defined, it cannot be used here.

MoodyMudit:
Discussed with Andres, about the code being split in 2 PRs. Separate PR was needed so that it could be raised in 4.0 branch as well. I will re raise the Docker grain PR in 4.0 later.

hubblestack/daemon.py
MoodyMudit: Removed it from here. But actually we can't use `__grains__` in util, since the util is imported in docker grain and the variable `__grains__` is only initialised after loading grains.

MoodyMudit:
Removed code from PR #839. That PR can be separately merged.

hubblestack/extmods/grains/docker_details.py
mew1033: Ah, got it. That makes sense.

hubblestack/daemon.py
```diff
@@ -33,6 +33,7 @@
 import hubblestack.log
 import hubblestack.hec.opt
 import hubblestack.utils.stdrec
+import hubblestack.utils.osquery_lib
```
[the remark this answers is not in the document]
MoodyMudit: This is needed otherwise all other modules do not see this module.

hubblestack/extmods/grains/docker_details.py
```diff
@@ -0,0 +1,75 @@
...
+    return False
+
+
+def _is_docker_process_running():
+  osquery_sql = 'select name from processes where name LIKE "%dockerd%"'
+  query_result = osquery_util(query_sql=osquery_sql)
+  if len(query_result) != 0:
+    for result in query_result:
+      process_name = result.get('name')
+      if 'dockerd' in process_name:
```
praksinha, line 69: @MoodyMudit Any specific reason for checking 'dockerd' string in process name again ? You have already specified the SQL with LIKE clause, this just seems to be overhead. If we are unusure about the query output, you should update the SQL instead.

hubblestack/extmods/grains/docker_details.py
```diff
@@ -0,0 +1,75 @@
...
+    else:
+      log.debug("OS not supported")
+      return False
+    query_result = osquery_util(query_sql=osquery_sql)
+    if len(query_result) != 0:
+      for result in query_result:
+        if isinstance(result, dict):
+          package_name = result.get('name')
+          log.debug('package_name = {0}'.format(package_name))
+          if package_name and 'docker' in package_name:
```
praksinha, line 55: @MoodyMudit Same here as well. Any specific reason of checking for 'docker' string in package_name as well when SQL query already handles that ?

hubblestack/extmods/grains/docker_details.py
MoodyMudit: Since the output of the query is returned by an external module, this is checked here as a precaution. Consider for example, the case when you know that an external function returns you a string, but even then you check if the returned output is an instance of string or not.

hubblestack/extmods/grains/docker_details.py
MoodyMudit: Same as above.

hubblestack/extmods/grains/docker_details.py
praksinha: Checking for type of return object or number of elements returned seems fine but somehow this seems overhead to me. Considering a utility function as a separate module may not be required here. If you fear that future changes in utility function might break the grain's functionality then it'll happen anyways since you'll fill default values ('False') in that case. Also, regressions can be avoided using unit tests (for both grain and utility function).

It's fine we want to be double check just to avoid any issues for now and later optimise if required.

MoodyMudit: I think the lib is an external entity. Since we are only calling its function and this particular grain does not know how that function is executed.

hubblestack/extmods/grains/docker_details.py
```diff
@@ -0,0 +1,76 @@
...
+Example Output - {u'installed': True, u'running': True}
+Author - Mudit Agarwal (muagarwa@adobe.com)
+"""
+import salt.utils.platform
+import logging
+from hubblestack.utils.osquery_lib import query as osquery_util
+log = logging.getLogger(__name__)
+
+def get_docker_details(grains):
+  try:
```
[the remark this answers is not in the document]
jettero: I actually posted these comments a week ago, I thought. I failed to hit submit again.

I don't think this is a huge problem or anything, but it might be worth looking at regardless.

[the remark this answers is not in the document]
MoodyMudit: removed try/except.

PR #846, opened 2020-05-05
resurrect old memory leak hunter

jettero:
this used to be a lot more difficult to get working. it's a no brainer now because tracemalloc comes with python3


daveed12:
LGTM

PR #847, opened 2020-05-05
S3 access denied

jettero:
access denied, like slowdown, was causing the entire schedule to cease functioning.

These are my proposed fixes, but please consider the implications carefully. I'm not 100% confident this is the right thing to do.


daveed12:
LGTM

fossam:
I didn't notice this pr in time.
Did it get backported ?

PR #848, opened 2020-05-12
osquery dockerfiles and augeas folder fix

fossam:
* include new Dockerfile for building osquery
* remove osquery build from all other Dockerfiles
* have augeas lens folder to be passed in by Hubble for snapshot queries

### DIFF
--- a/pkg/amazonlinux2016.09/Dockerfile
+++ b/pkg/amazonlinux2016.09/Dockerfile
@@ -1,81 +1,35 @@
-# This Dockerfile aims to make building Hubble v2 packages easier.
-# To build an image: 1. copy pkg/scripts/pyinstaller-requirements.txt to directory with this Dockerfile
-#                    2. docker build -t <image_name> .
-# The resulting image is ready to run the pyinstaller on container start and drop hubble<version>-coreos.tar.gz
-# in the /data directory. Mount /data volume into a directory on the host to access the package.
+# This Dockerfile aims to make building Hubble v4 packages easier.
+# Starting with version 4 building osquery is removed from individual Dockerfiles to its own.
+# osquery needs to be built once. Resulting tar file can be used in hubblev4 Dockerfiles.
+# Before building hubble, build osquery using a Dockerfile in pkg/osquery/ directory.
+# To build this image: 1. copy previously built osquery_4hubble.tar to directory with this Dockerfile
+#                      2. docker build -t <image_name> --build-arg=HUBBLE_CHECKOUT=<tag or commit> .
+# The resulting image is ready to build and run pyinstaller on container start that should 
+# create hubble<version>-al.tar.gz in the /data directory inside the container.
+# Mount /data volume into a directory on the host to access the package.
 # To run the container:  docker run -it --rm -v `pwd`:/data <image_name>
-# Requires docker 17.05 or higher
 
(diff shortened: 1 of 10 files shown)


pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -1,81 +1,35 @@
...
- && chmod -R 500 /opt/osquery/* \
-#put augeas lenses into the default directory that we changed earlier
- && mkdir -p /opt/osquery/lenses \
- && cp -r /usr/local/osquery/share/augeas/lenses/dist/* /opt/osquery/lenses \
- && chmod -R 400 /opt/osquery/lenses/*
-RUN ls -lahR /opt/osquery/ && /opt/osquery/osqueryi --version
+
+#copy in and process osquery files. optionally pass in osquery filename with OSQUERY_TAR_FILENAME build-arg
+ARG OSQUERY_TAR_FILENAME=osquery_4hubble.tar
+COPY ${OSQUERY_TAR_FILENAME} /opt/osquery/${OSQUERY_TAR_FILENAME}
```
jettero, line 21: The COPY creates a new layer (taking up space in the image). My understanding of how this works in practice leads me to believe there's really no point in removing the file in the next layer... it'll still take up space in the image.

Although, if the build takes place without cache, that may not be true anymore.

Have you looked at this? Is it crazy talk?

praksinha: +1 . COPYing a file and then deleting it in next layer would just make it inaccessible to newer layers but file would still be part of the image layers.

I'd prefer ADD over COPY in this case, since it supports file extraction and would not result in addition of tar file in image layer.

Example:

ADD ${OSQUERY_TAR_FILENAME} /opt/osquery/

Above statement would extract and add the files in 'tar' to '/opt/osquery' directory and actual tar file won't get copied to any layer.

PS: This might be a case of over-optimisation and may not be required. But since we are discussing these optimisations, I thought of commenting as well :)

fossam: Thank you, both. Good catch on layer efficiency.
While in this case the total image size difference is marginal, the reduction of unnecessary commands is well worth it.

jettero: It's certainly a case of over-optimization if disk space isn't an issue at all... but I have trouble building more than about 3 of these packages at my house without filling up my disks. If we were super consistent in keeping our layers as small as reasonably possible (without really annoying backflips); then diskspace would be even less of an issue.

Of course, this whole entire PR is about that, so yeah... it's a pretty small point by comparison.

fossam:
This is hopefully the last change and I think it is ready to be merged.
The last commit moved the instructions lower in the Dockerfiles so that if an osquery related change happens, all dependencies do get rebuilt. It should save some build time.

fossam:
@devagupt i think the Hubble git related change is best handled in a separate PR.

jettero:
> @devagupt i think the Hubble git related change is best handled in a separate PR.

jettero:
> > @devagupt i think the Hubble git related change is best handled in a separate PR.

I mean to cancel this reply ... ... but instead I closed the PR. 

Then I reopened it. ... next I'll push approve.

(it's been that kind of day.)

PR #850, opened 2020-05-18
4.0 revert checkout

jettero:
I officially have no idea why I changed the checkout from a naked checkout. I tested a bunch of things that might have caused me to change it and none of them are valid.

This reverts the checkout to the original method.

(Side note: I also opened this against develop... #849)


jettero:
> This change looks fine but dev docker files still have old osquery building logic in them.
> Need to update them as per separate osquery build logic.

OK, agreed.

I normally use a for loop to copy back the Dockers... I assumed there was some reason the files were super duper different. Now I think it's just cuz they got missed maybe.

```
for i in $(ls -1 pkg/*/Dockerfile | grep -v osquery)
do echo cp $i $(sed -e s,^pkg/,pkg/dev/, <<< "$i")
done | sh
```

Copying.

jettero:
> command needs to be changed from `git "$HUBBLE_CHECKOUT"` to `git checkout "$HUBBLE_CHECKOUT"`

wow, thanks.

It was right at one point. I wonder how I messed that up so badly.

devagupt:
Tested these files with branch. Working fine now.

PR #851, opened 2020-05-19
giant git squash merge to get develop branch in line with 4.0 branch

jettero:
Generally speaking it's a bad idea to do anything like this:
   `git merge --squash 4.0 -X theirs`

I couldn't think of any better way to get to exactly even parity between 4.0 branch and develop branch though.

It's 100% important to me that there's no difference between develop and 4.0 branches. It's less important to me to have completely perfect revision history for the develop branch.

`git diff develop..4.0`  <--- we are clean! Hooray??

This is super janky and I'd really prefer another way. Suggestions welcome! If we don't come up with anything else, this'll work I guess.


devagupt:
The drawback with this approach is that all commits of 4.0 are merged into the single commit. If we need to rollback a commit in future, then it will not be an easy task to do that in develop branch (We need to rollback the change in 4.0 branch and then merge it with develop for parity).

But going with normal merge will require a considerable time due to large number of conflicts. So we can go ahead this time but we should update and sync the two branches at frequent intervals (say 2 weeks) so that differences between the two does not pile up.

jettero:
agreed. I see this as a one time workaround for letting these branches get 6 months out of sync

fossam:
There is no way to review the change of this size.
I very much agree with Devansh' comments - this makes using history to understand where and why changes came from impossible.
May be attempt the merges file by file? And iterate until done.

PR #852, opened 2020-05-21
Modifying Dockerfiles to support use pre-built osquery

MoodyMudit:
All dockerfiles (dev and prod) have been modified to support the usage of pre-build osquery.

### DIFF
--- a/pkg/amazonlinux2016.09/Dockerfile
+++ b/pkg/amazonlinux2016.09/Dockerfile
@@ -8,8 +8,8 @@
 # Requires docker 17.05 or higher
 
 #Hubble builds can have the following 3 combinations wrt osquery:
-# 1. Package already built osquery. (BUILD_OSQUERY=false)
-# 2. Build osquery from local git repo. (BUILD_OSQUERY=true BUILD_OSQUERY_LOCALLY=true)
+# 1. Package already built osquery. (BUILD_OSQUERY=false). Keep the osquery tar file besides this Dockerfile.
+# 2. Build osquery from local git repo. (BUILD_OSQUERY=true BUILD_OSQUERY_LOCALLY=true). Keep the osquery source code besides this Dockerfile.
 # 3. Build osquery from https://github.com/facebook/osquery.git. (BUILD_OSQUERY=true BUILD_OSQUERY_LOCALLY=false)
 #--------------- ACTUAL DOCKERFILE FOR BUILD CREATION  --------------------------
 FROM amazonlinux:2016.09
--- a/pkg/amazonlinux2016.09/Dockerfile
+++ b/pkg/amazonlinux2016.09/Dockerfile
@@ -1,38 +1,16 @@
 # This Dockerfile aims to make building Hubble v2 packages easier.
 # To build an image: 1. copy pkg/scripts/pyinstaller-requirements.txt to directory with this Dockerfile
-#                    2. docker build -t <image_name> .
+#                    2. copy osquerybuilder.sh to this directory. This is needed to bundle osquery with Hubble
+#                    3. docker build -t <image_name> .
 # The resulting image is ready to run the pyinstaller on container start and drop hubble<version>-coreos.tar.gz
 # in the /data directory. Mount /data volume into a directory on the host to access the package.
 # To run the container:  docker run -it --rm -v `pwd`:/data <image_name>
 # Requires docker 17.05 or higher
 
-# Set this arguement to "local" if you want to build osquery for local code.
-# In that case, osquery folder must exist besides Dockerfile
-#ARG OSQUERY_BUILD_ENV=remote
-
-#--------------- TEMP CONTAINER FOR LOCAL OSQUERY -------------------------
-#FROM alpine as osquery_local
-#ONBUILD COPY osquery /osquery
-#ONBUILD RUN echo "Copying osquery from local folder"
-
-
-
-#--------------- TEMP CONTAINER FOR GIT OSQUERY ----------------------------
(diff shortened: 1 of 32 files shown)


pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -1,38 +1,16 @@
...
- && echo "Fetching osquery from git"
-
-
-#--------------- TEMP CONTAINER FOR OSQUERY ( BASED ON ARGUMENT ) --------------
-FROM osquery_"$OSQUERY_BUILD_ENV" as osquery_image
-
-
+#Hubble builds can have the following 3 combinations wrt osquery:
+# 1. Package already built osquery. (BUILD_OSQUERY=false)
+# 2. Build osquery from local git repo. (BUILD_OSQUERY=true BUILD_OSQUERY_LOCALLY=true)
```
devagupt, line 12: Please mention here that if we choose to built from local repo, then we need to provide the fetched code in osquery folder alongwith Dockerfile.

pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -1,38 +1,16 @@
...
- && git checkout "$OSQUERY_SRC_VERSION" \
- && echo "Fetching osquery from git"
-
-
-#--------------- TEMP CONTAINER FOR OSQUERY ( BASED ON ARGUMENT ) --------------
-FROM osquery_"$OSQUERY_BUILD_ENV" as osquery_image
-
-
+#Hubble builds can have the following 3 combinations wrt osquery:
+# 1. Package already built osquery. (BUILD_OSQUERY=false)
```
devagupt, line 11: Please mention here that we need to provide an osquery tar file along with docker file for this step.

devagupt:
Provided comments for a single docker file but they are applicable to all docker files.

pkg/amazonlinux2016.09/osquerybuilder.sh
```diff
@@ -0,0 +1,47 @@
...
+  echo $build_osquery
+else
+  echo $build_osquery_locally
+  if [ $build_osquery_locally == true ]
+  then
+    echo "Building osquery from local folder"
+  else
+    rm -rf /osquery
+    OSQUERY_SRC_VERSION=3.3.2
+    OSQUERY_GIT_URL=https://github.com/facebook/osquery.git
```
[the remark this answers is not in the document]
MoodyMudit: This is kept in line with the current workflow. In the current workflow there is no such provision to pass these parameters to dockerfile.

pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -44,35 +22,23 @@ RUN mkdir -p /etc/hubble/hubble.d /opt/hubble /opt/osquery /var/log/hubble_osque
...
-#put augeas lenses into the default directory that we changed earlier
- && mkdir -p /opt/osquery/lenses \
- && cp -r /usr/local/osquery/share/augeas/lenses/dist/* /opt/osquery/lenses \
- && chmod -R 400 /opt/osquery/lenses/*
-RUN ls -lahR /opt/osquery/ && /opt/osquery/osqueryi --version
+ADD osquerybuilder.sh /osquerybuilder.sh
+RUN sh /osquerybuilder.sh "$BUILD_OSQUERY" "$BUILD_OSQUERY_LOCALLY"
 
 #install packages that should be needed for ligbit2 compilation and successful pyinstaller run
+USER root
```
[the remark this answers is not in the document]
MoodyMudit: This is to execute the following steps as root user.

pkg/amazonlinux2016.09/Dockerfile
```diff
@@ -44,35 +22,23 @@ RUN mkdir -p /etc/hubble/hubble.d /opt/hubble /opt/osquery /var/log/hubble_osque
 #osquery build start
 #osquery should be built first since requirements for other packages can interfere with osquery dependencies
 #to build, osquery scripts want sudo and a user to sudo with.
+ENV OSQUERY_BUILD_USER=osquerybuilder
+ARG BUILD_OSQUERY=true
+ARG BUILD_OSQUERY_LOCALLY=false
+ARG OSQUERY_SRC_VERSION=osquery_3.3.2_tob
+COPY osquery /osquery
+ENV OSQUERY_TAR_FILENAME=$OSQUERY_SRC_VERSION"_amazonlinux2016.09.tar"
+COPY $OSQUERY_TAR_FILENAME /
```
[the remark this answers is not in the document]
MoodyMudit: This is to save the step of extracting. Since this tar is to be extracted only when pre-built osquery binary is being used, extracting it always gives us no benefit.

pkg/amazonlinux2016.09/Dockerfile
MoodyMudit: Done

pkg/amazonlinux2016.09/Dockerfile
MoodyMudit: Done

pkg/amazonlinux2016.09/osquerybuilder.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+build_osquery=$1
+build_osquery_locally=$2
+
+if [ $build_osquery != true ]
+then
+  tar -C /opt/ -zxvf osquery_*_tob_amazonlinux2016.09.tar
```
fossam, line 8: Any way to make this optionally user specified filename and not as specific?

PR #855, opened 2020-05-28
Change in permissions for windows config file

devagupt:
- Earlier the permissions were set on '/etc/hubble' folder and they were making the effective permissions of 'hubble.conf' as 0. Changed the logic to have 'hubble.conf' same permissions as '/etc/hubble'.

### DIFF
--- a/pkg/windows/hubble_config_permissions.ps1
+++ b/pkg/windows/hubble_config_permissions.ps1
@@ -1,5 +1,6 @@
 $hubble_path = $args[0]
-$hubble_conf_path = $hubble_path + "\etc\hubble\"
+$hubble_conf_path = $hubble_path + "\etc\hubble"
+$hubble_conf_file_path = $hubble_conf_path + "\hubble.conf"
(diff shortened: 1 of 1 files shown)


pkg/windows/hubble_config_permissions.ps1
```diff
@@ -1,35 +1,53 @@
...
+Set-Acl $hubble_conf_path $acl
+
+$items = Get-ChildItem -Recurse -Path $hubble_conf_path
+foreach ($item in $items) {
+  $acl = Get-Acl -Path $item.FullName
+  $acl.SetOwner($adminsGroup)
+  Set-Acl $item.FullName $acl
+}
+
+$acl1 = Get-Acl $hubble_conf_file_path
```
MoodyMudit, line 45: can you please set a more meaningful variable name here.

devagupt: thanks for pointing. Changed the name.

PR #867, opened 2020-06-15
dnf was getting 404 and bad mirror results

jettero:
I think we had an old stale layer above and the dnf -y update
(which seems duplicate) can fix the issue. At the same time,
combine up a couple run lines.


daveed12:
LGTM

PR #868, opened 2020-06-16
these files never get pushed to artifactory because of subtle naming issues

jettero:
The builds are all working, but literally half never make it to artifactory from jenkins due to subtle naming issues.

It seems like we should probably be building packages and naming artifacts in pipeline groovy to avoid these things.

### DIFF
--- a/pkg/amazonlinux2016.09/entrypoint.sh
+++ b/pkg/amazonlinux2016.09/entrypoint.sh
@@ -168,7 +168,7 @@ fpm -s dir -t rpm \
 # edit to change iteration number, if necessary
 PKG_BASE_NAME=hubblestack-${HUBBLE_VERSION}-${HUBBLE_ITERATION}
 PKG_OUT_EXT=x86_64.rpm
-PKG_FIN_EXT=all1609.$PKG_OUT_EXT
+PKG_FIN_EXT=al1609.$PKG_OUT_EXT
 PKG_ONAME="$PKG_BASE_NAME.$PKG_OUT_EXT"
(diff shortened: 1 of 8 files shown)


pkg/amazonlinux2016.09/entrypoint.sh
```diff
@@ -168,7 +168,7 @@ fpm -s dir -t rpm \
 # edit to change iteration number, if necessary
 PKG_BASE_NAME=hubblestack-${HUBBLE_VERSION}-${HUBBLE_ITERATION}
 PKG_OUT_EXT=x86_64.rpm
-PKG_FIN_EXT=all609.$PKG_OUT_EXT
+PKG_FIN_EXT=all1609.$PKG_OUT_EXT
```
devagupt, line 171: Please change the name to 'al1609'

pkg/dev/amazonlinux2016.09/entrypoint.sh
```diff
@@ -168,7 +168,7 @@ fpm -s dir -t rpm \
 # edit to change iteration number, if necessary
 PKG_BASE_NAME=hubblestack-${HUBBLE_VERSION}-${HUBBLE_ITERATION}
 PKG_OUT_EXT=x86_64.rpm
-PKG_FIN_EXT=all609.$PKG_OUT_EXT
+PKG_FIN_EXT=all1609.$PKG_OUT_EXT
```
devagupt, line 171: Please change the name to 'al1609'

pkg/amazonlinux2016.09/entrypoint.sh
jettero: ugh, still wrong. thanks

PR #871, opened 2020-06-16
Reverting amazon linux to previous revision

devagupt:
- New naming convention is throwing error. Earlier one was running fine for amazon linux. Reverting it to previous state.


MoodyMudit:
A similar PR #870 has been opened already

PR #882, opened 2020-07-01
Salt-less Hubble - Next set of libraries for Grains

goravsingal:
- moved salt.utils.user to hubblestack.util.user
- moved salt.platform.win to hubblestack.platform.win
- moved salt.utils.win_functions to hubblestack.utils.win_functions
- moved salt.utils.win_runas to hubblestack.utils.win_runas
- moved salt.utils.network to hubblestack.utils.network
- moved salt.utils.dns to hubblestack.utils.dns
- moved salt._compat to hubblestack.utils._compat
- moved salt.utils.win_osinfo to hubblestack.utils.win_osinfo
- moved salt.utils.pkg.rpm to hubblestack.utils.pkg.rpm
- moved salt.modules.smbios to hubblestack.modules.smbios
- moved salt.utils.win_reg to hubblestack.utils.win_reg.py

Now in grains/hubble_core.py, only salt.log is remaining.
We will take it up in separate salt.log activity


hubblestack/grains/hubble_core.py
```diff
@@ -2376,11 +2372,11 @@ def dns():
     if hubblestack.utils.platform.is_windows() or 'proxyminion' in __opts__:
         return {}
 
-    resolv = salt.utils.dns.parse_resolv()
+    resolv = hubblestack.utils.dns.parse_resolv()
     for key in ('nameservers', 'ip4_nameservers', 'ip6_nameservers',
                 'sortlist'):
         if key in resolv:
-            resolv[key] = [six.text_type(i) for i in resolv[key]]
+            resolv[key] = [str(i) for i in resolv[key]]
```
jettero, line 2379: Are you sure str() is right?

I mean, I think you're changing numbers to strings here, but in the worst case, str() doesn't correctly convert bytes to strings.

hubblestack/grains/hubble_core.py
```diff
@@ -2559,7 +2533,7 @@ def _hw_data(osdata):
             'serialnumber': 'serial#',
             'productname': 'DeviceDesc',
         }
-        for grain_name, cmd_key in six.iteritems(hwdata):
+        for grain_name, cmd_key in iter(hwdata.items()):
```
jettero, line 2536: why iter()? isn't hwdata.items() enough? I think iter might be extra here.

jettero:
This is really exciting stuff. I requested some really minor changes (or at least confirming the code is right in at least one spot); but I admit that my eyes glazed over at one point, so I may have missed other small things.

Maybe review any changes that aren't simply renames for correctness? I may still do the same before we click merge on this one.

hubblestack/grains/hubble_core.py
goravsingal: Since we are working on Python-3. Old code's six.text_type is str only.

hubblestack/grains/hubble_core.py
goravsingal: In six.py, for Python-3 it was:

def iteritems(d, **kw):
        return iter(d.items(**kw))

So, no change.

jettero: Right, but isn't for item in dictthing.items() the same thing as iter(dictthing.items()) -- except with an extra iterator around it?

goravsingal: You are right. I tried running the test code, and it was the same behavior without extra iter()
I changed it

PR #883, opened 2020-07-01
Hubblestack loader

jettero:
This is a much stripped down loader that's more like what we want for `hubblestack.loader`. It's still extremely salt friendly for now, but I expect we'll have to do at least one more round of changes to it to get it ready for the next release anyway (eg, it's full of salt.* imports).

With this update, the tests are repaired or otherwise still working and the loader can find loadable modules in any of `prefix/hubblestack/extmods/modules`, `prefix/hubblestack/modules`, `prefix/salt/modules`, or `/var/cache/hubble/extmods`; that last one may seem suprising, but even more surprising (imo) was that previous to this change hubble would load salt modules it found in `/var/cache/salt/minion/extmods` without mentioning it (nightmare).

I've also renamed a mess of things to be more hubble-y instead of salt-y and I renamed `__salt__` to `__mods__` but left `__salt__` as a working alias.

Please be critical! \o/

I want to make sure we get this piece right, so if you see something we should change for style points or correctness, definitely speak up -- but, you know, be mindful this is a first pass. I've already marked a dozen things with `XXX` meaning: we still need to fix this later. The correct fix for some things you spot may be marking `XXX` for the next pass.


hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+            continue
+
+        maybe_dir = os.path.join(_dir, '_{0}'.format(ext_type))
+        if os.path.isdir(maybe_dir):
+            cli_module_dirs.insert(0, maybe_dir)
+
+    return cli_module_dirs + ext_type_types + [ext_types, sys_types, old_types]
+
+
+def modules(
```
goravsingal, line 132: Its a rename from minion_mod(), this change should also be done in daemon.py, it still uses minion_mods()

hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+    .. code-block:: python
+
+        import salt.config
+        import salt.loader
+
+        __opts__ = salt.config.minion_config('/etc/salt/minion')
+        __grains__ = salt.loader.grains(__opts__)
+        __opts__['grains'] = __grains__
+        __utils__ = salt.loader.utils(__opts__)
+        __salt__ = salt.loader.minion_mods(__opts__, utils=__utils__)
```
goravsingal, line 173: This should also change to reflect __mods__

hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+        loaded_base_name=loaded_base_name,
+        static_modules=static_modules,
+    )
+
+    ret.pack['__salt__'] = ret
+
+    # Load any provider overrides from the configuration file providers option
+    #  Note: Providers can be pkg, service, user or group - not to be confused
+    #        with cloud providers.
+    providers = opts.get('providers', False)
```
goravsingal, line 194: I think we do not need Providers. Do we use them anywhere?

hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+def modules(
+        opts,
+        context=None,
+        utils=None,
+        whitelist=None,
+        initial_load=False,
+        loaded_base_name=None,
+        notify=False,
+        static_modules=None,
+        proxy=None):
```
goravsingal, line 141: We don't need proxy. We can remove its references too.

hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+            log.trace('Loading %s grain', key)
+            parameters = salt.utils.args.get_function_argspec(funcs[key]).args
+            kwargs = {}
+            if 'proxy' in parameters:
+                kwargs['proxy'] = proxy
+            if 'grains' in parameters:
+                kwargs['grains'] = grains_data
+            ret = funcs[key](**kwargs)
+        except Exception:
+            if salt.utils.platform.is_proxy():
```
goravsingal, line 368: we can also remove this

hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+            )
+            continue
+        if not isinstance(ret, dict):
+            continue
+        if grains_deep_merge:
+            salt.utils.dictupdate.update(grains_data, ret)
+        else:
+            grains_data.update(ret)
+
+    if opts.get('proxy_merge_grains_in_module', True) and proxy:
```
goravsingal, line 383: I believe this is also for proxy minions.

hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+        for mod_dir in self.module_dirs:
+            try:
+                # Make sure we have a sorted listdir in order to have
+                # expectable override results
+                files = sorted(
+                    x for x in os.listdir(mod_dir) if x != '__pycache__'
+                )
+            except OSError:
+                continue  # Next mod_dir
+            if six.PY3:
```
goravsingal, line 662: Since we are on Python-3 now. We can actually straight away use statements for PY3, and can remove code branches for PY2

hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+                    return False
+        else:
+            virtual_aliases = ()
+
+        # If this is a proxy minion then MOST modules cannot work. Therefore, require that
+        # any module that does work with salt-proxy-minion define __proxyenabled__ as a list
+        # containing the names of the proxy types that the module supports.
+        #
+        # Render modules and state modules are OK though
+        if 'proxy' in self.opts:
```
goravsingal, line 1010: It is for proxy minions

hubblestack/loader.py
```diff
@@ -0,0 +1,1292 @@
...
+                elif virtual is True and virtualname != module_name:
+                    if virtualname is not True:
+                        module_name = virtualname
+
+        except KeyError:
+            # Key errors come out of the virtual function when passing
+            # in incomplete grains sets, these can be safely ignored
+            # and logged to debug, still, it includes the traceback to
+            # help debugging.
+            log.debug('KeyError when loading %s', module_name, exc_info=True)
```
goravsingal, line 1270: Can we make it ERROR level. I faced this issue when I tried to load some grains, there was no error and my grain was not loaded.
Making it error can upfront convey the issues in grains. Anyway, we don't put incomplete grains in our folders.

hubblestack/extmods/modules/pulsar.py
```diff
@@ -634,7 +634,7 @@ def mark(self,name):
 @hubble_status.watch
 def process(configfile='salt://hubblestack_pulsar/hubblestack_pulsar_config.yaml',
             verbose=False):
-    """
+    r"""
```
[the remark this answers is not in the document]
jettero: false. That is because there is an escaped character in the docstring. the 'r' is required for it to be properly parsed and formatted (at least by linters).

hubblestack/loader.py
jettero: I can do that. I provided the alias so they'll both work for now. But we should certainly aim to remove the alias eventually.

hubblestack/loader.py
jettero: agreed. Nice catch. I didn't see that on any of my 30 passes.

hubblestack/loader.py
jettero: indeed

hubblestack/loader.py
jettero: I wanted to get this done sooner than later so I elected to skip removing six everywhere. I removed some. I think you're right though. I'll remove the rest today.

hubblestack/loader.py
jettero: FYI, I figure at least some of the mods will expect proxy arguments and the global `__proxy__` to exist; so I didn't strip it out of args lists and still provide the `__proxy__` global to modules at load. 

You're right about this one for sure though. removed.

hubblestack/loader.py
jettero: Yes. This is a good idea, but ... we'll have to keep an eye on this. I imagine it'll get very noisy and send millions of boring duplicate errors to Splunk.

jettero:
Thanks for your detailed review @goravsingal. That's a lot of work.

goravsingal:
@jettero I realized you want to merge this into develop branch.
Develop branch is the one from where we can release some feature builds. But, this code is untested and can bring whole code into an unstable state. This code should go in 4.1, which is kind of feature branch for salt-less work.

jettero:
> @jettero I realized you want to merge this into develop branch.
> Develop branch is the one from where we can release some feature builds. But, this code is untested and can bring whole code into an unstable state. This code should go in 4.1, which is kind of feature branch for salt-less work.

I noticed this just a short while ago... then read your comment.

I changed the base branch of the PR.

jettero:
> @jettero I realized you want to merge this into develop branch.
> Develop branch is the one from where we can release some feature builds. But, this code is untested and can bring whole code into an unstable state. This code should go in 4.1, which is kind of feature branch for salt-less work.

Further thought. 4.1 should definitely be in develop (soonishly); master should be what is currently develop and we should lose the 4.0 branch...

develop: 4.1.x -- untested, etc
master: 4.0.x
3.0: old 3.0.x support

We're not currently configured that way, but that's what the group decided I think.

hubblestack/loader.py
jettero: removed six

PR #885, opened 2020-07-09
fixes for hubblestack.loader

jettero:
nova_loader is a mess. But the problem here was a pathing problem that should really be the same problem in salt.loader, but isn't for some reason.

This quick little patch (namely `sys.path.append( os.path.dirname(fpath_dirname) )`) solves that problem. I've also included a couple other little notes for things to remove, improved some error logging, and removed a salt event we'll never need or use.


hubblestack/extmods/modules/hubble.py
```diff
@@ -28,8 +29,8 @@
 
 try:
     from nova_loader import NovaLazyLoader
```
goravsingal, line 31: Shouldn't we remove this file altogether, and start using hubble.loader in hubble.py

jettero: Yes. I was tempted to do this, but decided to mark it XXX and leave it for later. I think you're right though. Let me figure that out quick today.

jettero:
I think this PR is finally ready. Please look at the very last patch so you can scratch your head and think, "why would that prevent the filesystems from loading or caching files???"

The exception was completely hidden by try/except within try/except within try/except, but the keyword mismatch was in fact crashing the fileserver and causing it to not actually cache files but claim that it did.

jettero:
In case the email thread is super confusing ... I edited my last comment twice so it would point to the right commit.

PR #888, opened 2020-07-13
Nova rewrite - first draft

devagupt:
- Added audit core module
- Added submodules for nova along with their test cases


jettero:
FYI, we may have to co-ordinate later regards to the NovaLazyLoader. We're removing that loader (and the salt.loader) in the saltless hubble branch and it's probably going to conflict with the nova re-write.

devagupt:
> FYI, we may have to co-ordinate later regards to the NovaLazyLoader. We're removing that loader (and the salt.loader) in the saltless hubble branch and it's probably going to conflict with the nova re-write.

Yes Paul. This is a to do item in the list when both of these branches will merge sometime in future.

mew1033:
Agreed. Let's use hubblestack/extmods/audit.
@jettero, we need to run nova v1 and the new audit system simultaneously for a little while while we transition.

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,556 @@
...
-        - cis.linux
-        - adobebaseline
-      'G@os:CoreOS':
-        - coreos_additional_checks
-
-Audit modules take arbitrary args and kwargs and return a tuple
-``(success, data_dict)`` where ``success`` is a boolean True/False on whether
-it was a success or a failure, and ``data_dict`` is a dictionary of any
-information that should be added to the check's data dictionary in the return.
-"""
```
mew1033, line 96: Can we keep documentation like this in the file? This specific documentation might not work exactly; can you write something like this and add it to your audit.py?

hubblestack/daemon.py
```diff
@@ -700,9 +700,17 @@ def _setup_dirs():
     fdg_dirs = __opts__.get('fdg_dirs', [])
     fdg_dirs.append(os.path.join(os.path.dirname(__file__), 'extmods', 'fdg'))
     __opts__['fdg_dirs'] = fdg_dirs
-    audit_dirs = __opts__.get('audit_dirs', [])
-    audit_dirs.append(os.path.join(os.path.dirname(__file__), 'extmods', 'audit'))
-    __opts__['audit_dirs'] = audit_dirs
+    
+    # Audit module written by coulton, no longer in use
```
mew1033, line 704: Typo. It's Colton.

jettero:
Oh, we do need nova.v1 for now and we need nova.v2 at the same time for a while; but we call all nova things 'audit'. Handy that we don't need the module already in that audit dir, so we can just do all this under the extmods/audit folder.

devagupt:
> Agreed. Let's use hubblestack/extmods/audit. In fact, @goravsingal let's call this the `audit` module as opposed to nova_v2. We weren't using the audit module that Colton wrote: this should replace/enhance that.
> 
> @jettero, we need to run nova v1 and the new audit system simultaneously for a little while while we transition.

Made changes to use 'hubblestack/extmods/audit' instead of 'hubblestack/extmods/nova_v2_modules'

hubblestack/extmods/modules/audit.py
devagupt: Hi @mew1033,
Thanks for input. I have added documentation for the audit module.

hubblestack/daemon.py
devagupt: Fixed

hubblestack/extmods/audit/bexpr.py
```diff
@@ -0,0 +1,197 @@
...
+                if check_result == "Success":
+                    referred_checks_result[check] = "True"
+                elif check_result == "Failure":
+                    referred_checks_result[check] = "False"
+                else:
+                    error[check] = "Unable to fetch result of referred check due to errors. Check: %s" % (check)
+                break
+        if not check_found:
+            error[check] = "Referred check: %s is not available. Please verify correct check is referred." % (check)
+    if error:
```
goravsingal, line 70: Other error condition can be if someone pass empty expr.

Also, do we also check if I pass any other operand (other than and, or, not)?

hubblestack/extmods/audit/bexpr.py
```diff
@@ -0,0 +1,197 @@
...
+    referred_checks_result = {}
+    for expr in expr_list:
+        if expr.upper() not in keyword_list:
+            referred_checks_list.append(expr)
+
+    #Fetch the result of check from result list and store the result of referred checks
+    error = {}
+    for check in referred_checks_list:
+        check_found = False
+        for result in result_list:
```
goravsingal, line 57: result_list must be a dictionary. No need to iterate and break.

hubblestack/extmods/audit/bexpr.py
```diff
@@ -0,0 +1,197 @@
...
+                    referred_checks_result[check] = "False"
+                else:
+                    error[check] = "Unable to fetch result of referred check due to errors. Check: %s" % (check)
+                break
+        if not check_found:
+            error[check] = "Referred check: %s is not available. Please verify correct check is referred." % (check)
+    if error:
+        raise AuditCheckValidationError(error)
+
+    #Convert the expression now in the format to be parsed by pyparsing module
```
goravsingal, line 73: This code actually is confusing in between.
We can write a method in this file, and do this processing there (the one required by 3rd party parsing module.)

hubblestack/extmods/audit/bexpr.py
```diff
@@ -0,0 +1,197 @@
...
+        boolOperand,
+        [
+            ("NOT", 1, opAssoc.RIGHT, BoolNot),
+            ("AND", 2, opAssoc.LEFT, BoolAnd),
+            ("OR", 2, opAssoc.LEFT, BoolOr),
+        ],
+    )
+    result = boolExpr.parseString(parsed_expr)[0]
+    if not bool(result):
+        return {"result": False,
```
goravsingal, line 106: We need to put in the logs, which check is pass/fail.

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,702 @@
...
-            log.exception('Error loading audit file %s: %s', audit_file, exc)
-            return False, audit_data
-    if not audit_data or not isinstance(audit_data, dict):
-        log.error('audit data from %s was not formed as a dict', audit_file)
-        return False, audit_data
-    return True, audit_data
+            log.error(exc)
+
+    #Evaluate boolean expressions
+    if boolean_expr_check_list:
```
goravsingal, line 287: A separate function would be good in this file only.

hubblestack/extmods/audit/bexpr.py
devagupt: Handled the empty string error condition in validate params method. 
We only have allowed operands ('AND', 'OR', 'NOT', '(', ')'). Apart from these and whitespaces, every other char is treated as referenced check. So in case a user supplied an extra operand like 'XOR', then it will be treated as a referenced check and it will result in failure.

hubblestack/extmods/audit/bexpr.py
```diff
@@ -0,0 +1,197 @@
...
+    - filter:
+        grains: '*'
+      module: bexpr
+      checks:
+        # If docker is not installed OR is not running, we would want result as TRUE
+        - expr: NOT (docker_installed AND dockerd_running) OR find_docker_file
+
+Here we are evaluating the boolean expression given in the 'expr' string value
+In boolean expressions we refer to the checks present in same profile and evaluate the result based on the result of those checks.
+A boolean expressio is evaluated in the end after all other checks have been evaluated.
```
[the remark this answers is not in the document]
devagupt: Fixed

hubblestack/extmods/modules/audit.py
devagupt: Changed this code to have a separate function for boolean expressions.

hubblestack/extmods/audit/bexpr.py
devagupt: Result list is a list of dictionaries. Hence the logic to iterate and check for each item in the list.

hubblestack/extmods/audit/bexpr.py
devagupt: Created a separate function for expression evaluation

hubblestack/extmods/audit/bexpr.py
devagupt: Added more detailed failure reason for boolean expressions

devagupt: Added more detailed failure reason for boolean expression

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,726 @@
...
+    audit_result['run_config']['checks'] = []
+    execute_method = audit_impl['module'] + '.execute'
+    filtered_log_method = audit_impl['module'] + '.get_filtered_params_to_log'
+    # calculate the check result based on type parameter.
+    # If type is 'and', all subchecks should pass for success.
+    # If type is 'or', any passed subcheck will result in success.
+    overall_result = type == 'and'
+    failure_reasons = []
+    for audit_check in audit_impl['checks']:
+        if result_list is not None:
```
mew1033, line 396: Let's add some comments here to indicate why we're doing this. (For bexpr checks.)

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,726 @@
...
+        audit_result['check_result'] = CHECK_STATUS['Success']
+    else:
+        # If check result is failure, fetch failure reason. If it is not present in profile, combine all individual checks reasons.
+        audit_result['check_result'] = CHECK_STATUS['Failure']
+        if failure_reason:
+            audit_result['failure_reason'] = failure_reason
+        else:
+            if failure_reasons:
+                failure_reasons = set(failure_reasons)
+                audit_result['failure_reason'] = ', '.join(failure_reasons)
```
mew1033, line 436: Let's just submit a list object as the failure reasons. It can be flattened into a comma separated string in Splunk if needed.
```suggestion
            audit_result['failure_reason'] = [failure_reason]
        else:
            if failure_reasons:
                failure_reasons = set(failure_reasons)
                audit_result['failure_reason'] = list(failure_reasons)
```

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,726 @@
...
+            else:
+                if invert_result:
+                    audit_result['failure_reason'] = 'Check failed due to invert result is set to true'
```
mew1033, line 439: Typo
```suggestion
                    audit_result['failure_reason'] = 'Check failed because invert result is set to true'
```

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,726 @@
...
-        raise CommandExecutionError('Could not load topfile: {0}'.format(exc))
 
-    if not isinstance(topdata, dict) or 'audit' not in topdata or \
-            not isinstance(topdata['audit'], dict):
-        raise CommandExecutionError('Audit topfile not formatted correctly.')
+def _get_matched_implementation(audit_check_id, audit_data, tags, labels):
+    log.debug('Getting matching implementation')
 
-    topdata = topdata['audit']
+    # check if label passed is matching with the check or not.
```
mew1033, line 521: Typo
```suggestion
    # check if label passed matches the check or not.
```

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,726 @@
...
 
-        if data_dict and isinstance(data_dict, dict):
-            data_dict.update(data)
-        else:
-            data_dict = data
+    Returns:
+        list -- List of audit files
+    """
+    if not audit_files:
+        log.warning('nova.audit called without any audit_files')
```
mew1033, line 591: nova should be audit

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,726 @@
...
-def _process_data(data, tags, labels, label_list):
-    """ Validate data, log if necessary, return False to skip the audit, True to continue """
-    audit_id = data['audit_id']
-    tag = data.get('tag', audit_id)
+def _evaluate_results(result_dict, combined_dict, show_compliance, verbose):
+    """
+    Evaluate the result dictionary to be returned by the audit module
+    :param result_dict: Final dictionary to be returned
+    :param combined_dict: Initial dictionary with results for all profiles
+    :param show_compliance: Para to show compliance percentage
```
mew1033, line 607:
```suggestion
    :param show_compliance: Param to show compliance percentage
```

hubblestack/extmods/modules/audit.py
```diff
@@ -1,450 +1,726 @@
...
+    for audit_file in combined_dict:
+        result_list = combined_dict[audit_file]
+        for result in result_list:
+            sub_check = result.get('sub_check', False)
+            if not sub_check:
+                dict = {}
+                if verbose:
+                    dict[result['tag']] = result
+                else:
+                    dict[result['tag']] = result['description']
```
mew1033, line 620: In non-verbose mode, we should probably use either the tag (`ADOBE-00033`) or the check_id.

tests/unittests/nova_modules/test_fdg.py
```diff
@@ -0,0 +1,103 @@
+from unittest import TestCase
```
mew1033, line 1: Lots of references to Nova here.

hubblestack/extmods/audit/stat.py
```diff
@@ -0,0 +1,202 @@
...
+          allow_more_strict: true
+
+It is checking for file on the given path and matches the permissions on it with given params.
+Mandatory parameters:
+    path - file path
+    gid - group id
+    group - group name
+    mode - file mode
+    uid - user id
+    user - user name
```
mew1033, line 33: Maybe we should allow only some of the parameters? Would we ever only want to enforce some of these options in a check?

hubblestack/extmods/audit/stat.py
```diff
@@ -0,0 +1,202 @@
...
+        given_x = True
+
+    if given_r and (not allowed_r):
+        return False
+    if given_w and (not allowed_w):
+        return False
+    if given_x and (not allowed_x):
+        return False
+
+    return True
```
mew1033, line 202: Missing newline

hubblestack/extmods/audit/sysctl.py
```diff
@@ -0,0 +1,103 @@
...
+    if not salt_ret or "No such file or directory" in salt_ret:
+        return {"result": False,
+                "failure_reason": "Could not find attribute %s in the kernel" %(name)}
+    if salt_ret.lower().startswith("ERROR".lower()):
+        return {"result": False,
+                "failure_reason": "An error occurred while reading the value of kernel attribute %s" %(name)}
+    if match_output_regex:
+        if not re.search(match_output, salt_ret):
+            return {"result": False,
+                    "failure_reason": "Current value of kernel attribute %s is %s It is not matching with regex: %s" %(name, salt_ret, match_output)}
```
mew1033, line 58:
```suggestion
                    "failure_reason": "Current value of kernel attribute %s is %s It does not match with regex: %s" %(name, salt_ret, match_output)}
```

hubblestack/extmods/audit/sysctl.py
```diff
@@ -0,0 +1,103 @@
...
+    mandatory_params = ['name', 'match_output']
+    error = {}
+    for mandatory_param in mandatory_params:
+        if mandatory_param not in audit_check:
+            error[mandatory_param] = 'Mandatory parameter: "%s" not found for check-id: %s' % (mandatory_param, check_id)
+
+    if error:
+        raise AuditCheckValidationError(error)
+
+    log.debug('Validation success for check-id: %s' %(check_id))
```
mew1033, line 103: newline

hubblestack/extmods/audit/bexpr.py
```diff
@@ -0,0 +1,237 @@
...
+      module: bexpr
+      checks:
+        # If docker is not installed OR is not running, we would want result as TRUE
+        - expr: NOT (docker_installed AND dockerd_running) OR find_docker_file
+
+Here we are evaluating the boolean expression given in the 'expr' string value
+In boolean expressions we refer to the checks present in same profile and evaluate the result based
+on the result of those checks.
+A boolean expression is evaluated in the end after all other checks have been evaluated.
+While evaluating boolean expression the result of checks referred is taken into the account and no
```
mew1033, line 20: Typo
```suggestion
While evaluating boolean expression the result of checks referred is taken into account and no
```

hubblestack/extmods/audit/bexpr.py
```diff
@@ -0,0 +1,237 @@
...
+check is actually run.
+We can provide a string in expression which consists of following expressions:
```
mew1033, line 22:
```suggestion
We can provide a string in `expr` which consists of the following expressions:
```

PR #890, opened 2020-07-16
Docker file changes for fast build system

devagupt:
- Moved git cloning logic to entrypoint so that it is not cached by docker and runs everytime.
- Cleaned up various redundant files to make docker images smaller which will be helpful in caching.


devagupt:
> Please flatten this or justify the very first commit being a merge commit in a non public branch.
> 
> (You can certainly rebase this properly.)
> 
> In fact, the more I look at this, the more I can see it's all merges from various places. It looks like a 3-way merge or worse, which is the very thing we'd like to avoid with the flat history policy.
> 
> At the very least, please try to explain why you have constructed this PR in this way or what these merges are from and why, then I can at least start to think about whether this is necessary or if there's a way to do this so we can figure it out later.

Hi Paul, 
I will certainly rebase this commit.

PR #891, opened 2020-07-19
Adding FDG Command line parser module

MoodyMudit:
The module can be used to fetch value of keys provided in command lines. 
Example:
"dockerd --config-file=/etc/docker/daemon.json --log-level=debug"
In the above command line, the module can be used to fetch the value of `config-file` and `log-level`.

### DIFF
--- a/hubblestack/extmods/fdg/command_line_parser.py
+++ b/hubblestack/extmods/fdg/command_line_parser.py
@@ -1,0 +1,195 @@
+"""
+This module is used to find the value of a command line parameter.
+Inputs : key_aliases (List) : List of keys whose value is to be fetched
+         delimiter (String): the assignment operator between the key and the value
+         chained (String): The command line in which to search for keys.
+                    This needs to be passed from another FDG module via chaining.
+         chained_status (boolean): Status returned by the chained method.
+Output (tuple):  tuple with two values. First item is the status of the module's execution.
+                Second item is list of values corresponding to the given key_aliases, fetched from the chained string.
+
+BASIC EXAMPLES:
+Inputs:
+    key_aliases = ["config-file"]
+    delimiter = "="
+    chained = "dockerd --config-file=/etc/docker/daemon.json --log-level=debug"
+Output: (True, ["/etc/docker/daemon.json"])
+
+UNIT TESTS:
+Check the unit test case file 'test_command_line_parser.py' to understand more examples on how this module works.
+
+NOTES:
+
+Understanding the regex(s):
+You can copy paste the regex(s) on this webpage - https://regex101.com/ to understand more about the regex.
+However, here is a small explanation
+
+regex_pattern :
+    "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s?" + delimiter + "\s?openingbracket.*closingbracket(?=(\s|$))"
+    This regex primarily focuses on whether the value is embedded inside brackets.
+
+1. '?<=' : denotes a positive lookbehind
+2. '\s' : matches the <space> character
+3. '{' : matches opening braces
+4. (\s|{) : () brackets signify a group. This is the first group in this regex
+4. -{0,2} : matches (zero to two) hyphens.
+5. \"{0,1}\'{0,1} : matches if any quotes are present.
+5. key_alias : matches the given key_alias.
+6. /s? : matches 0 or 1 <space> character
+7. delimiter : matches the input delimiter
+8. openingbracket : is later replaced by one of the three opening bracket types
+9. .*? : matches any character lazily.
+10. closingbracket : is later replaced by one of the three closing bracket types
+11. (?=(\s|$|})) : matches that the end of the pattern must contain either <space> or EOL
+
+regex1 :
+    "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s?" + delimiter + "\s?([\",\']).*?\\2"
+    This regex primarily focuses on whether the value is embedded inside double or single quotes
+
+The starting regex is similar to the regex_pattern.
+1. ([\", \']) : () brackets signify a group. This is the second group in this regex.
+                    [\",\'] inside the () matches either a double-quote or a single-quote.
+2. .*? : matches any character lazily.
+3. \\2 : matches the second group of the regex.
+
+regex2 :
+    "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s?" + delimiter + "\s?.+?(?=(\s|$|}))"
+The starting regex is similar to the regex_pattern.
+
+1. (?=(\s|$)) : matches that the end of the pattern must contain either <space> or EOL
+"""
+
+import logging
+import re
+
+log = logging.getLogger(__name__)
+default_delimiters = [":", "=", " "]
+logging.basicConfig(level=logging.DEBUG)
+open_bracket_list = ["[","{","("]
+close_bracket_list = ["]","}",")"]
+
+
+def parse_cmdline(params='', chained=None, chained_status=None):
+    try:
+        if not params:
+            ret = {'Failure': 'invalid input, no params provided'}
+            return False, ret
+
+        if not params.get('key_aliases'):
+            log.error("No key_aliases provided in params to parse_cmdline, returning False")
+            return False, ''
+        else:
+            key_aliases = params.get('key_aliases')
+
+        if not chained:
+            log.error("No chained value provided to parse_cmdline function, returning False")
+            return False, ''
+        elif not chained.get('cmdline'):
+            log.error("cmdline not provided to parse_cmdline function, returning False")
+            return False, ''
+        else:
+            command_line = chained.get('cmdline')
+
+        if not params.get('delimiter') or len(params.get('delimiter')) > 1:
+            log.info("Either length of delimiter exceeds 1 or no delimiter provided with "
+                     "command line '%s'", command_line)
+            return False, ''
+        else:
+            delimiter = params.get('delimiter')
+
+        log.debug("value of command_line is %s", command_line)
+        log.debug("value of key_aliases is %s", key_aliases)
+        log.debug("value of delimiter is '%s'", delimiter)
+        ret_match_list = []
+
+        for key_alias in key_aliases:
+            log.debug("looping with key_alias %s", key_alias)
+            key_alias = key_alias.lstrip("-")
+
+            if key_alias not in command_line:
+                log.info("key_alias %s not found in command line %s", key_alias, command_line)
+                continue
+
+            regex_list = []
+            regex_pattern = "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s?" + delimiter + "\s?openingbracket.*closingbracket(?=(\s|$))"
+            braces_list = [('\[','\]'), ('\{','\}'), ('\(','\)')]
+            for item in braces_list:
+                regex = re.sub("openingbracket", item[0], regex_pattern)
+                regex = re.sub("closingbracket", item[1], regex)
+                regex_list.append(regex)
+            regex1 = "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s?" + delimiter + "\s?([\",\']).*?\\2"
+            regex2 = "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s?" + delimiter + "\s?.+?(?=(\s|$))"
+            regex_list.append(regex1)
+            regex_list.append(regex2)
+
+            for regex in regex_list:
+                log.debug("looping with regex : %s", regex)
+                match_list = _get_match_list(regex, key_alias, command_line, delimiter)
+                if match_list:
+                    ret_match_list.extend(match_list)
+                    break
+
+        return True, ret_match_list
+    except Exception as e:
+        log.exception("Some exception occurred in command_line_parser's parse_cmdline function %s", e)
+        return False, ''
+
+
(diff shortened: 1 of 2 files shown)


hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,193 @@
...
+            log.error("No chained value provided to parse_cmdline function, returning False")
+            return False, ''
+        elif not chained.get('cmdline'):
+            log.error("cmdline not provided to parse_cmdline function, returning False")
+            return False, ''
+        else:
+            command_line = chained.get('cmdline')
+
+        if not params.get('delimiter') or len(params.get('delimiter')) > 1:
+            log.info("Either length of delimiter exceeds 1 or no delimiter provided with "
```
devagupt, line 92: Please also log the delimiter provided and change the log level from info to error

devagupt, line 92: Is there any specific reason of restricting delimiters to single chars only?

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,193 @@
...
+            regex_list = []
+            regex_pattern = "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s*" + delimiter + "\s*openingbracket.*closingbracket(?=(\s|$))"
+            braces_list = [('\[','\]'), ('\{','\}'), ('\(','\)')]
+            for item in braces_list:
+                regex = re.sub("openingbracket", item[0], regex_pattern)
+                regex = re.sub("closingbracket", item[1], regex)
+                regex_list.append(regex)
+            regex1 = "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s*" + delimiter + "\s*([\",\']).*?\\2"
+            regex2 = "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s*" + delimiter + "\s*.+?(?=(\s|$))"
+            regex_list.append(regex1)
```
devagupt, line 120: Please validate the scenario where command can be split across multiple lines

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,193 @@
...
+                                                                                   match=match.group()))
+        value = match.group().lstrip("-")
+        value = value.replace("\"" + key_alias + "\"", key_alias)
+        value = value.replace("\'" + key_alias + "\'", key_alias)
+        prefix = key_alias + "\s*" + delimiter + "\s*"
+        value = re.sub(prefix, '', value)
+        value = value.rstrip('\"\' ')
+        value = value.lstrip('\"\' ')
+        if value[0] in open_bracket_list:
+            value = _fetch_bracketed_value(value)
```
devagupt, line 164: Please add error logging here in case empty string is returned from _fetch_bracketed_value function

tests/unittests/test_command_line_parser.py
```diff
@@ -0,0 +1,653 @@
...
+        'delimiter': '='
+    }
+    val = command_line_parser.parse_cmdline(params=params, chained=command_line)
+    log.debug("return value is {0}".format(val))
+    expected_value = (True, ["a:b", "d:e"])
+    assert val[0] == expected_value[0]
+    assert val[1] == expected_value[1]
+
+
+def test_second_regex():
```
devagupt, line 116: Please add a testcase in case there are multiple spaces in between alias and value

tests/unittests/test_command_line_parser.py
```diff
@@ -0,0 +1,653 @@
...
+    key_aliases = ["v"]
+    params = {
+        'key_aliases': key_aliases,
+        'delimiter': '='
+    }
+    val = command_line_parser.parse_cmdline(params=params, chained=command_line)
+    log.debug("return value is {0}".format(val))
+    expected_value = (True, [('a:b')])
+    assert val == expected_value
+
```
devagupt, line 653: Please add a test case where value is split to new line.

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,193 @@
...
+
+import logging
+import re
+
+log = logging.getLogger(__name__)
+open_bracket_list = ["[","{","("]
+close_bracket_list = ["]","}",")"]
+
+
+def parse_cmdline(params='', chained=None, chained_status=None):
```
goravsingal, line 70: Should use 'None' to represent null or nothing.

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,193 @@
...
+    try:
+        if not params:
+            ret = {'Failure': 'invalid input, no params provided'}
+            return False, ret
+
+        if not params.get('key_aliases'):
+            log.error("No key_aliases provided in params to parse_cmdline, returning False")
+            return False, ''
```
goravsingal, line 78: Should use 'None' to represent null or nothing. This is at more places in this file

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,193 @@
...
+
+            if key_alias not in command_line:
+                log.info("key_alias %s not found in command line %s", key_alias, command_line)
+                continue
+
+            regex_list = []
+            regex_pattern = "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s*" + delimiter + "\s*openingbracket.*closingbracket(?=(\s|$))"
+            braces_list = [('\[','\]'), ('\{','\}'), ('\(','\)')]
+            for item in braces_list:
+                regex = re.sub("openingbracket", item[0], regex_pattern)
```
goravsingal, line 115: This constant string should be in a variable, as it is being used at more than one place. If we need to change this string in future (for some reason), new dev can miss changing it at multiple places.

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,193 @@
...
+                regex = re.sub("closingbracket", item[1], regex)
+                regex_list.append(regex)
+            regex1 = "(?<=(\s|{))-{0,2}\"{0,1}\'{0,1}" + key_alias + "\"{0,1}\'{0,1}\s*" + delimiter + "\s*([\",\']).*?\\2"
```
goravsingal, line 118: String concatenation by '+' is bad idea in almost every language, especially when we are appending many strings.

Can use join or f-string.

hubblestack/extmods/fdg/command_line_parser.py
goravsingal: I think this case will never happen, as I think it is getting result from osquery

hubblestack/extmods/fdg/command_line_parser.py
MoodyMudit: It is logged at the very next line. Changed info to error.

MoodyMudit: The delimiters `=` `:` `<space>` are all single characters. However, I have removed that check.

hubblestack/extmods/fdg/command_line_parser.py
MoodyMudit: Done.

tests/unittests/test_command_line_parser.py
MoodyMudit: There is already such a test case `test_extra_spaces_different_delimiter`

hubblestack/extmods/fdg/command_line_parser.py
MoodyMudit: Done.

hubblestack/extmods/fdg/command_line_parser.py
MoodyMudit: Done.

hubblestack/extmods/fdg/command_line_parser.py
MoodyMudit: Done.

hubblestack/extmods/fdg/command_line_parser.py
MoodyMudit: Done. However can you please tell why is '+' not a good option?

hubblestack/extmods/fdg/command_line_parser.py
MoodyMudit: Gorav is right in saying this. I tested as follows `dockerd --config-file=\
> /home/centos/mudit.json
`
`ps ax| grep docker
 4834 pts/1    Sl+    0:00 dockerd --config-file=/home/centos/mudit.json`
`osquery> select cmdline from processes where name like "%docker%";
+-----------------------------------------------+
| cmdline                                       |
+-----------------------------------------------+
| dockerd --config-file=/home/centos/mudit.json |
+-----------------------------------------------+`

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,198 @@
...
+            log.error("No chained value provided to parse_cmdline function, returning False")
+            return False, None
+        elif not chained.get('cmdline'):
+            log.error("cmdline not provided to parse_cmdline function, returning False")
+            return False, None
+        else:
+            command_line = chained.get('cmdline')
+
+        if not params.get('delimiter'):
+            log.error("Either length of delimiter exceeds 1 or no delimiter provided with "
```
devagupt, line 94: Please fix the error log as now we are not checking the length of delimiter

hubblestack/extmods/fdg/command_line_parser.py
goravsingal: Multiple appends with '+' creates new string object for each two appends. Since string is mutable. Also, if appending more than 2 strings with '+' causes complexity of O(n^2)

Also see this PEP recommendation: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

fossam:
Perhaps we should cover this in one of our meetings, but it seems that this is somewhat complex solution to the problem.

Since for a given check or a query we already know which argument we're after, isn't it a bit simpler to rex out the right thing in query itself ?

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -129,6 +129,9 @@ def parse_cmdline(params=None, chained=None, chained_status=None):
                     ret_match_list.extend(match_list)
                     break
 
+        if not ret_match_list:
+            log.info("No Match was found for %s in %s, returning False", key_aliases, command_line)
+            return False, ret_match_list
```
goravsingal, line 134: Its wrong to return False here. It represents module did not execute, like in error conditions.
This is what fdg module does.

But here, return value is based on expectation.

MoodyMudit: Agreed, removed this code.

tests/unittests/test_command_line_parser.py
```diff
@@ -0,0 +1,653 @@
...
+                   "-Dcatalina.base=/opt/oobe/oobe-tomcat-9.0.31 -Dcatalina.home=/opt/oobe/apache-tomcat-9.0.31 " \
+                   "-Djava.io.tmpdir=/opt/oobe/oobe-tomcat-9.0.31/temp org.apache.catalina.startup.Bootstrap start"}
+
+    key_aliases = ["Djdk.tls.ephemeralDHKeySize"]
+    params = {
+        'key_aliases': key_aliases,
+        'delimiter': '='
+    }
+    val = command_line_parser.parse_cmdline(params=params, chained=command_line)
+    log.debug("return value is {0}".format(val))
```
jettero, line 417: There's quite a few of these...

They should all be fixed because otherwise the logic is something like:

```
   log.debug("message".format())
     formatted_message = msg % args
     handler.do_stuff(formatted_message)
```

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,197 @@
...
+
+    This function will fetch all matching patterns of regex in command_line and will then strip the
+    matched value to remove the leading and trailing hyphens (-), <spaces> and quotes.
+    Since the regex also contains the 'key' and 'delimiter', that is also stripped away to get the final value.
+    """
+    match_list = []
+
+    matches = re.finditer(regex, command_line)
+    for matchNum, match in enumerate(matches, start=1):
+        log.info("Match {matchNum} was found at {start}-{end}: {match}".format(matchNum=matchNum,
```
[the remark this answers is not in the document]
MoodyMudit: Done

hubblestack/extmods/fdg/command_line_parser.py
```diff
@@ -0,0 +1,197 @@
...
+    :return: List of all matching patterns
+
+    This function will fetch all matching patterns of regex in command_line and will then strip the
+    matched value to remove the leading and trailing hyphens (-), <spaces> and quotes.
+    Since the regex also contains the 'key' and 'delimiter', that is also stripped away to get the final value.
+    """
+    match_list = []
+
+    matches = re.finditer(regex, command_line)
+    for matchNum, match in enumerate(matches, start=1):
```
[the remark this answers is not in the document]
MoodyMudit: Done

tests/unittests/test_command_line_parser.py
```diff
@@ -0,0 +1,653 @@
...
+    """
+    log.info("Executing test_match_key_alias_in_middle_of_cmdline")
+    command_line = {"cmdline" : "dockerd --config-file=\"/etc/docker/daemon.json\" --log-level=\"debug\""}
+    key_aliases = ["config-file"]
+    params = {
+        'key_aliases': key_aliases,
+        'delimiter': '='
+    }
+    val = command_line_parser.parse_cmdline(params=params, chained=command_line)
+    log.debug("return value is {0}".format(val))
```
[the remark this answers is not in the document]
MoodyMudit: Done

tests/unittests/test_command_line_parser.py
MoodyMudit: Done

goravsingal (approved):

PR #896, opened 2020-08-04
Modifying pidfile path: nebula_osqueryd_monitor

zachsis:
hello, 

While trying to install and test the 4.0.0 debian 10 release, I noticed that the default path listed for the `nebula_osqueryd_monitor:pidfile` is pointed at a folder that doesn't exist. When enabling this schedule, its not clear in debug logs why this is failing without manually running the following command: 

```
root@ip-172-31-12-24:/opt/osquery# /opt/osquery/hubble_osqueryd --pidfile=/var/run/hubble_osquery/osquery.pidfile --logger_path=/var/log/hubble_osquery --config_path=/etc/hubble/osquery.conf --flagfile=/etc/hubble/osquery.flags --database_path=/var/cache/hubble/osquery 

E0804 18:42:01.685073 10098 init.cpp:384] osqueryd initialize failed: Could not create file: /var/run/hubble_osquery/osquery.pidfile
```
I propose we point this to `/var/run/osquery.pidfile` by default so it works out of the gate and let users modify that if they wish.

 https://github.com/hubblestack/hubble/releases/download/v4.0.0/hubblestack-4.0.0-10.deb10.amd64.deb

### DIFF
--- a/conf/hubble
+++ b/conf/hubble
@@ -88,7 +88,7 @@ fileserver_backend:
 #      flagfile: /path/to/confbase/osquery/osquery.flags
 #      logdir: /var/log/hubble_osquery
 #      databasepath: /var/cache/hubble/osquery
-#      pidfile: /var/run/osquery.pidfile
+#      pidfile: /var/run/hubble_osquery.pidfile
 #      hashfile: /var/cache/hubble/osquery/hash_of_flagfile.txt
 #    run_on_start: True
 #  nebula_osqueryd_log_parser:
--- a/conf/hubble
+++ b/conf/hubble
@@ -88,7 +88,7 @@ fileserver_backend:
 #      flagfile: /path/to/confbase/osquery/osquery.flags
 #      logdir: /var/log/hubble_osquery
 #      databasepath: /var/cache/hubble/osquery
-#      pidfile: /var/run/hubble_osquery/osquery.pidfile
+#      pidfile: /var/run/osquery.pidfile
 #      hashfile: /var/cache/hubble/osquery/hash_of_flagfile.txt
 #    run_on_start: True
 #  nebula_osqueryd_log_parser:


conf/hubble
```diff
@@ -88,7 +88,7 @@ fileserver_backend:
 #      flagfile: /path/to/confbase/osquery/osquery.flags
 #      logdir: /var/log/hubble_osquery
 #      databasepath: /var/cache/hubble/osquery
-#      pidfile: /var/run/hubble_osquery/osquery.pidfile
+#      pidfile: /var/run/osquery.pidfile
```
MoodyMudit, line 91: I don't think this is a good idea. It is possible that someone has installed a different copy of osquery on their machine and then they install Hubble. The original installation of osquery would write its pidfile to the location you mentioned. If in this case, hubble's osqueryd also writes its pidfile to the same location then that would cause some trouble.

jettero: I agree. Though, under systemd, it probably wouldn't cause all that much trouble. Systemd tracks the processes in a cgroup, so it would hardly care about the contents of a pidfile. It's more like just confusing for the humans looking in the file.

jettero: @zachsis could you explain the thinking here? Is this a problem with the /var/run/hubble_osquery dir being absent ... or maybe some other thing I'm not thinking about?

zachsis: if we account for creating the `/var/run/hubble_osquery` at some point during the install, I think that would be sufficient. The idea is that the boiler plate configs should point to a folder that exists so there is less manual intervention by the user to set that up. i.e. if you uncomment line 91, it should work without additional troubleshooting.

zachsis: how about we change it to a lesss generic pidfile name and still use `/var/run/` as the path unless you guys would rather account for creating `/var/run/hubble_osquery` on install? https://github.com/hubblestack/hubble/pull/896/commits/02c4c598161e1264258744c22bd8e6657808d7c3

jettero (approved):

conf/hubble
jettero: @zachsis @MoodyMudit where did we leave off on this? I don't think we need a dir in /var/run; but if we do use a dir like that, we should at least make sure it exists in each of the systemd/init-d startups (or whatever). I was about to merge this (because it's just a comment in a default config file); but it's blocked waiting to see the outcome of this change request.

MoodyMudit: If we choose to have `/var/run/osquery.pidfile` folder as default, then this can overlap with any other osquery installation on the system (although having two osquery would be highly unlikely). But in case someone has two osqueries installed and writing pid to the same file then the one corresponding to Hubble would face problems. This is because, Hubble monitors that the pid in the pidfile is same as the pid of the process. In case it is not the same, Hubble would restart osquery. 
This is the reason we chose to have a separate folder `hubble_osquery` too keep the pidfile separate. 
I agree we should create the folder `/var/run/hubble_osquery/` to remove that manual folder creation step. 
@zachsis Let me know if this addressed properly. 
As such, this PR can be closed, and a new one (or even modify this one) with the folder creation should be raised.

jettero:
@MoodyMudit @zachsis made changes to this last summer that may have mitigated your review. Thoughts?

jettero:
I'm going to go ahead and close this. There's no way it'd merge now and we just plain spent to long without merging it. Sorry about that.

Issue #895, opened 2020-08-04
Dockerfile - osquery packaging broken


zachsis:
Hello, 

when I build the osquery package first, that runs and completes. When copying the `osquery_4hubble.tar` to `pkg/ubuntu1804/` (copy of `pkg/debian10/`) and running `docker build -t 'ubuntu:bionic' .` it gets to the following step and fails: 

https://github.com/hubblestack/hubble/blob/6456af53c5eff10a7ca91ecba4abdbfe18e8084e/pkg/debian10/Dockerfile#L121

```
[zach@atreides:~/hubble/pkg/ubuntu1804/tmp]$ docker build -t 'ubuntu:bionic' .
...
Step 31/49 : RUN /opt/osquery/osqueryi --version
 ---> Running in 101b428b975b
/bin/sh: 1: /opt/osquery/osqueryi: not found
The command '/bin/sh -c /opt/osquery/osqueryi --version' returned a non-zero code: 127
```

 and fails when it tries to run `/opt/osquery/osqueryi` which doesn't exist in the tarball that gets made in the previous step. Here are the contents of the `osquery_4hubble.tar` file. Notice that `osqueryi` is not in the tarball that gets made. 

 

```
[zach@atreides:~/hubble/pkg/ubuntu1804/tmp]$ tar -tvvf osquery_4hubble.tar
dr-x------ root/root         0 2020-08-04 00:09 extensions/
dr-x------ root/root         0 2020-08-04 00:09 lenses/
-r-x------ root/root      1602 2020-08-04 00:09 lenses/afs_cellalias.aug
-r-x------ root/root     10287 2020-08-04 00:09 lenses/mdadm_conf.aug
-r-x------ root/root      4120 2020-08-04 00:09 lenses/xinetd.aug
-r-x------ root/root      4426 2020-08-04 00:09 lenses/hosts_access.aug
-r-x------ root/root       869 2020-08-04 00:09 lenses/collectd.aug
-r-x------ root/root       745 2020-08-04 00:09 lenses/shells.aug
-r-x------ root/root      3087 2020-08-04 00:09 lenses/crypttab.aug
-r-x------ root/root      2151 2020-08-04 00:09 lenses/postgresql.aug
-r-x------ root/root      1966 2020-08-04 00:09 lenses/wine.aug
-r-x------ root/root      1449 2020-08-04 00:09 lenses/spacevars.aug
-r-x------ root/root      1697 2020-08-04 00:09 lenses/puppetfile.aug
[...]

PR #897, opened 2020-08-04
missing osqueryi to build hubble package.

zachsis:
ref: https://github.com/hubblestack/hubble/issues/895 

This resolved building a custom hubble package for me.

### DIFF
--- a/pkg/osquery/Dockerfile
+++ b/pkg/osquery/Dockerfile
@@ -48,6 +48,7 @@ USER root
 #collect all files and prepare them into a tar file
 RUN mkdir -p /opt/osquery/lenses /opt/osquery/yara /opt/osquery/extensions \
  && cp /home/"$OSQUERY_BUILD_USER"/osquery/build/osquery/osqueryd /opt/osquery \
+ && cp /home/"$OSQUERY_BUILD_USER"/osquery/build/osquery/osqueryi /opt/osquery/ \
  && cp -r /home/"$OSQUERY_BUILD_USER"/osquery/libraries/cmake/source/augeas/src/lenses /opt/osquery/ \
  && rm -rf /opt/osquery/lenses/tests/ \
  && chown -R root. /opt/osquery \


pkg/osquery/Dockerfile
```diff
@@ -48,6 +48,7 @@ USER root
 #collect all files and prepare them into a tar file
 RUN mkdir -p /opt/osquery/lenses /opt/osquery/yara /opt/osquery/extensions \
  && cp /home/"$OSQUERY_BUILD_USER"/osquery/build/osquery/osqueryd /opt/osquery \
```
MoodyMudit, line 50: I know this is a separate issue altogether. However, osqueryd needs to be copied as **/opt/osquery/hubble_osqueryd**. Refer [here](https://github.com/hubblestack/hubble/blob/develop/hubblestack/extmods/modules/nebula_osquery.py#L1291)

MoodyMudit:
@jettero This error is a recent one, when the dockerfiles for osquery and hubble were separated.

pkg/osquery/Dockerfile
```diff
@@ -48,6 +48,7 @@ USER root
...
+ && cp /home/"$OSQUERY_BUILD_USER"/osquery/build/osquery/osqueryi /opt/osquery/ \
```
MoodyMudit, line 51: Also **/home/"$OSQUERY_BUILD_USER"/osquery/build/osquery/osqueryi** doesn't actually exists. The osqueryd binary is the same as osqueryi binary. So osqueryd needs to be copied at both places. The correct cp statement should be **cp /home/"$OSQUERY_BUILD_USER"/osquery/build/osquery/osqueryd /opt/osquery/osqueryi**

MoodyMudit:
Closing this one since I have opened a new PR with fixes https://github.com/hubblestack/hubble/pull/899

pkg/osquery/Dockerfile
jettero: or maybe a hardlink? if it's really the same file, symlinking or hardlinking would seem the right thing to do to me.

PR #898, opened 2020-08-05
I failed to push some commits in 885

jettero:
(see: #885)

... this is the rest of the commits that were meant to be at the end of the PR.

Plus some bugfixes.


hubblestack/extmods/modules/hubble.py
```diff
@@ -671,13 +671,14 @@ def load():
     log.debug('loading nova modules')
 
     global __nova__
-    __nova__ = NovaLazyLoader(_hubble_dir(), __opts__, __grains__, __pillar__, __salt__)
+    __nova__ = NovaLazyLoader(_hubble_dir(), __opts__, __mods__)
 
     ret = {'loaded': list(__nova__._dict.keys()),
```
goravsingal, line 676: Still typo: "_dict"

jettero: That's actually not a typo. I thought it was when I commented on it, but it's actually something that's populated in salt.utils.lazy in LazyDict

PR #899, opened 2020-08-07
fixing osquery package issue

MoodyMudit:


jettero:
I think #897 was closed improperly (this change should have been appended there); but this otherwise seems fine.

PR #900, opened 2020-08-09
forward port Docker details grain #843

MoodyMudit:


hubblestack/extmods/grains/docker_details.py
```diff
@@ -0,0 +1,70 @@
...
+import salt.utils.platform
+import logging
+from hubblestack.utils.osquery_lib import query as osquery_util
+log = logging.getLogger(__name__)
+
+def get_docker_details(grains):
+  docker_grains = {}
+
+  if salt.utils.platform.is_windows():
+    log.debug('This grain is only available on linux')
```
jettero, line 16: In order to properly integrate with the error handling and grains loading, this grain needs a virtual section.

Look how they do it in the zfs grain: https://github.com/saltstack/salt/blob/master/salt/grains/zfs.py#L42

```
def __virtual__():
    if something:
        return False, "this won't load here"
    return True
```

MoodyMudit: Done.

Issue #860, opened 2020-06-10
Use a single request for GCP cloud details


basepi:
While working on [cloud metadata](https://github.com/elastic/apm-agent-python/pull/826) for the Elastic APM python agent, I was using the [cloud_details](https://github.com/hubblestack/hubble/blob/54062cf07bf2462ea9be149d740f38defd849b25/hubblestack/extmods/grains/cloud_details.py#L145-L232) grain as a reference and realized with a little testing that you can just use a single REST API request to get all of the data from GCP. This could save some overhead. Just use the URL `http://metadata.google.internal/computeMetadata/v1/?recursive=true`:

```
colton_myers@basepi-test:~$ curl "http://metadata.google.internal/computeMetadata/v1/?recursive=true" -H "Metadata-
Flavor: Google" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 15287  100 15287    0     0  1066k      0 --:--:-- --:--:-- --:--:-- 1066k
{
  "instance": {
    "attributes": {},
    "cpuPlatform": "Intel Skylake",
    "description": "",
    "disks": [
      {
        "deviceName": "basepi-test",
        "index": 0,
        "interface": "SCSI",
        "mode": "READ_WRITE",
        "type": "PERSISTENT"
      }
    ],
    "guestAttributes": {},
    "hostname": "basepi-test.c.elastic-apm.internal",
    "id": 4306570268266786300,
    "image": "projects/ubuntu-os-cloud/global/images/ubuntu-2004-focal-v20200609",
    "legacyEndpointAccess": {
      "0.1": 0,
      "v1beta1": 0
    },
    "licenses": [
      {
        "id": "2211838267635035815"
      }
    ],
    "machineType": "projects/513326162531/machineTypes/n1-standard-1",
    "maintenanceEvent": "NONE",

<snip>
```

PR #901, opened 2020-08-10
Bug fix in cloud details grain in fetching instance metadata

devagupt:
- Fixed bug in fetching cloud account id and instance id in different clouds (AWS, Azure, GCP). Now an error will be logged and empty value is filled for that particular cloud, if there is any error or redirection for metadata URL.
- Changed order of calling metadata API for different clouds. Earlier, all cloud API's were called and only one of the present was expected to return the values. Now, the order of calls for clouds is as follows:
1.AWS
2.Azure
3.GCP
The calls will go to next cloud only when the current cloud is unable to provide instance metadata.

As this bug is impacting current releases as well as future releases, it needs to be back ported to 3.0 along with 4.0 branch.

Added logic to call a single API in GCP rather than calling multiple API.

### DIFF
--- a/hubblestack/extmods/grains/cloud_details.py
+++ b/hubblestack/extmods/grains/cloud_details.py
@@ -3,7 +3,9 @@ HubbleStack Cloud Details Grain
 """
 
 import requests
+import logging
 
+log = logging.getLogger(__name__)
 
 def get_cloud_details():
     """
@@ -12,15 +14,23 @@ def get_cloud_details():
     grains = {}
 
     aws = _get_aws_details()
-    azure = _get_azure_details()
-    gcp = _get_gcp_details()
-
-    if aws['cloud_details']:
+    if not aws['cloud_details']: # Unable to fetch details from AWS. Let's try from Azure
+        log.debug("Unable to fetch AWS details. Now trying to fetch Azure details")
+        azure = _get_azure_details()
+        if not azure['cloud_details']: # Unable to fetch details from Azure. Let's try from GCP
+            log.debug("Unable to fetch Azure details. Now trying to fetch GCP details")
+            gcp = _get_gcp_details()
+            if gcp['cloud_details']:
+                log.debug("Fetched instance metadata from GCP")
+                grains.update(gcp)
+            else:
+                log.error("Unable to fetch details from AWS/Azure/GCP. Please verify the instance settings.")
+        else:
+            log.debug("Fetched instance metadata from Azure")
+            grains.update(azure)
+    else:
+        log.debug("Fetched instance metadata from AWS")
         grains.update(aws)
-    if azure['cloud_details']:
-        grains.update(azure)
-    if gcp['cloud_details']:
-        grains.update(gcp)
 
     return grains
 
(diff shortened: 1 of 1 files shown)


hubblestack/extmods/grains/cloud_details.py
```diff
@@ -12,15 +14,16 @@ def get_cloud_details():
     grains = {}
 
     aws = _get_aws_details()
-    azure = _get_azure_details()
-    gcp = _get_gcp_details()
-
-    if aws['cloud_details']:
+    if not aws['cloud_details']: # Unable to fetch details from AWS. Let's try from Azure
+        azure = _get_azure_details()
```
MoodyMudit, line 18: you can add debug log statements here for future debugging cases. Similarly, you can add debug statements at multiple points in the code below as well.

hubblestack/extmods/grains/cloud_details.py
```diff
@@ -154,19 +171,29 @@ def _get_gcp_details():
     proxies = {'http': None}
 
     try:
-        gcp['cloud_instance_id'] = requests.get(
+        response = requests.get(
             'http://metadata.google.internal/computeMetadata/v1/instance/id',
```
jettero, line 175: as long as we're messing around in here, can we look at #860 ?

hubblestack/extmods/grains/cloud_details.py
```diff
@@ -40,17 +43,27 @@ def _get_aws_details():
...
+            res = response.json()
+            aws['cloud_account_id'] = res.get('accountId', 'unknown')
+
+            # AWS account id is always an integer number
+            # So if it's an aws machine it must be a valid integer number
+            # Else it will throw an Exception
+            int(aws['cloud_account_id'])
+        else:
+            log.error("Error while fetching AWS account id. Got status code: %s " % (response.status_code))
+            raise ValueError("Error while fetching AWS account id. Got status code: %s " % (response.status_code))
```
goravsingal, line 58: 1. You can move this log.error in catch() block. It will remove this duplicate error string at two places.
    Same for code at line 65
2. In error message, put account id as well.

hubblestack/extmods/grains/cloud_details.py
```diff
@@ -40,17 +43,27 @@ def _get_aws_details():
...
-        aws['cloud_account_id'] = res.get('accountId', 'unknown')
-
-        # AWS account id is always an integer number
-        # So if it's an aws machine it must be a valid integer number
-        # Else it will throw an Exception
-        int(aws['cloud_account_id'])
-        aws['cloud_instance_id'] = requests.get('http://169.254.169.254/latest/meta-data/instance-id',
-                                                headers=aws_token_header, timeout=3,
-                                                proxies=proxies).text
+        response = requests.get('http://169.254.169.254/latest/dynamic/instance-identity/document',
```
goravsingal, line 46: For these kind of calls, retry always helps.
Does retry impact performance much? (Retry for non-200 response code)

hubblestack/extmods/grains/cloud_details.py
```diff
@@ -178,34 +205,49 @@ def _get_gcp_details():
...
-            headers=gcp_header, timeout=3, proxies=proxies).text,
-        'cloud_tags': requests.get(
-            'http://metadata.google.internal/computeMetadata/v1/instance/tags?recursive=true',
-            headers=gcp_header, timeout=3, proxies=proxies).json()}
-    interface_list = requests.get(
-        'http://metadata.google.internal/computeMetadata/v1/instance/network-interfaces'
-        '/?recursive=true', headers=gcp_header, timeout=3, proxies=proxies).json()
+    response = requests.get('http://metadata.google.internal/computeMetadata/v1/project/project-id',
+            headers=gcp_header, timeout=3, proxies=proxies)
+    cloud_project_id = response.text if response.status_code == requests.codes.ok else None
```
goravsingal, line 212: This is inconsistent with aws/azure impl.
There we raised exception for non-200 response. Here its None.

hubblestack/extmods/grains/cloud_details.py
```diff
@@ -12,15 +14,16 @@ def get_cloud_details():
     grains = {}
 
     aws = _get_aws_details()
-    azure = _get_azure_details()
-    gcp = _get_gcp_details()
-
-    if aws['cloud_details']:
+    if not aws['cloud_details']: # Unable to fetch details from AWS. Let's try from Azure
```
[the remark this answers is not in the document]
devagupt: cloud_details will always be present in aws dict. However, it's value will either be None or actual value.

hubblestack/extmods/grains/cloud_details.py
devagupt: added debug statements

devagupt:
Added changes for review comments along with the functionality in #860

PR #908, opened 2020-08-31
add grains module

Tenebriso:
I have added the grains.py module. I have removed the functions that modify the grains and the ones that I are kept:

- ls

- get

- items

- item

- equal

- has_value

Not sure if I should remove some of these too or add some that I've removed.
I have also added the _utils.json.py_ file on which the module depends.

Also, while trying to run hubble and test that the module works correctly, I have encountered some issues with the _utils.stringutils.py_ file and solved its dependencies issues by adding a couple of functions, as well as tests for those.


hubblestack/utils/decorators/jinja.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+Jinja-specific decorators
+"""
+# Import Python libs
+import logging
+
+log = logging.getLogger(__name__)
+
+
+class JinjaFilter(object):
```
goravsingal, line 11: We actually do not need Jinja thing. It's for Jinja template support.
Its used in Salt for some other purpose when Master/Minion is there. 
https://docs.saltstack.com/en/master/topics/jinja/index.html#understanding-jinja

We can safely remove the usage of JinjaFilter

goravsingal (changes requested):

hubblestack/utils/decorators/jinja.py
Tenebriso: Now that yoou mention it, I remeber we talked about this and decided it wasn't needed. I'll remove it right away

goravsingal (approved):

PR #914, opened 2020-09-08
Update splunklogging SplunkHandler class with new grains on refresh

mew1033:


fossam:
The main review will take place at #915 (pr against the develop branch). Head over there for longer comments and discussion.

hubblestack/splunklogging.py
```diff
@@ -164,6 +164,14 @@ def emit(self, record):
             hec.flushBatch()
         return True
 
+    def update_event_std_info(self):
+        '''
+        Update the `event` template in the `endpoint_list` object. This allows
+        grains and other values that were updated to be updated here.
+        '''
+        for entry in self.endpoint_list:
+            entry[1] = entry[1].update(hubblestack.utils.stdrec.std_info())
```
jettero, line 173: this should probably be simply `entry[1].update(...)`. The return value of `dict.update()` is None.

jettero (changes requested):

hubblestack/splunklogging.py
mew1033: Fixed.

jettero (approved):

PR #915, opened 2020-09-08
Update splunklogging SplunkHandler class with new grains on refresh

mew1033:
Also remove unneeded code

### DIFF
--- a/hubblestack/splunklogging.py
+++ b/hubblestack/splunklogging.py
@@ -70,7 +70,7 @@ class SplunkHandler(logging.Handler):
             args, kwargs = make_hec_args(opts)
             hec = http_event_collector(*args, **kwargs)
 
-            fqdn = SplunkHandler._get_fqdn(__grains__['id'])
+            fqdn = hubblestack.utils.stdrec.get_fqdn()
 
             event = {}
             event.update(hubblestack.utils.stdrec.std_info())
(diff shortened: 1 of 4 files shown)


fossam:
This pr seems to be making subtle changes to already subtle code. Dealing with the downstream effects of hard to foresee corner cases has been an ongoing issue. I am afraid that it will be a bigger issue going forward unless we make it easier to understand the changes to those who will need to look at this later.

@mew1033, can you record an explanation of these changes in two parts:
* at the high level the problem the pr aims to solve
* at the low level how these particular changes fix or avoid issues (preferably without affecting anything else).

hubblestack/splunklogging.py
```diff
@@ -171,6 +137,14 @@ def emit(self, record):
             hec.flushBatch()
         return True
 
+    def update_event_std_info(self):
+        """
+        Update the `event` template in the `endpoint_list` object. This allows
+        grains and other values that were updated to be updated here.
+        """
+        for entry in self.endpoint_list:
+            entry[1] = entry[1].update(hubblestack.utils.stdrec.std_info())
```
jettero, line 146: This should probably be `entry[1].update(...)` the return value of `dict.update` is None.

jettero (changes requested):

hubblestack/splunklogging.py
mew1033: Fixed. Thanks!

mew1033:
@fossam I agree with you. This is a confusing change, and I want to do it better. However, we need this in 3.0 and 4.0. Once we fixed that, we should have a bigger discussion about doing this the right way.

mew1033:
I created #933 to get this change merged into 4.0. We can continue to use this PR to discuss the proper way to implement this/revamp the logging.

jettero (approved):

PR #916, opened 2020-09-09
Porting fileserver/fileclient from salt

devagupt:
- Code for fileserver/fileclient modules from salt.
- Their dependencies and other referred modules.
- Adding win_reg and win_network modules


hubblestack/utils/gzip_util.py
```diff
@@ -0,0 +1,61 @@
+# -*- coding: utf-8 -*-
+"""
+    salt.utils.gzip
```
goravsingal, line 3: to be replaced with hubblestack

goravsingal (changes requested):

goravsingal (approved):

PR #919, opened 2020-09-10
use the modern requirements rather than that old test-req file

jettero:
this might just fix the various test problems installing old pprint packages and salt 3000 builds and things.


devagupt (changes requested):
pprint is still included in requirements file (https://github.com/hubblestack/hubble/blob/4.1/requirements.txt)
Need to remove pprint as a dependency from all requirement files.

devagupt (approved):

PR #936, opened 2020-09-28
Audit module rewrite

devagupt:
- Changes for Audit module rewrite

1. Audit module
2. Comparators
3. All sub modules
4. Unit tests


jettero (changes requested):
I'm pretty much OK with most of these changes (they've been through committee); but could we get this rebased to develop? it's full of (unnecessary to preserve) merge commits and the parent fork point is from some seven weeks ago.

devagupt:
> I'm pretty much OK with most of these changes (they've been through committee); but could we get this rebased to develop? it's full of (unnecessary to preserve) merge commits and the parent fork point is from some seven weeks ago.

@jettero Thanks for pointing it out. I have fixed the commit history to have relevant commits only.

jettero:
Let's wait to see how the jenkins job turns out. It's pretty overloaded (I re-queued everything). But assuming it passes, I'll approve/merge later today.

goravsingal:
It's failing as the step mentioned in .pipeline is expecting a file: combined.svg. But it is not found.

jettero (changes requested):
Please fix the history. There's a merge commit at the tail end of the commit list that's from some unknown branch.

hubblestack/extmods/hubble_mods/win_reg.py
```diff
@@ -0,0 +1,223 @@
...
+        grains: 'G@osfinger:Microsoft Windows Server 2016*'
+      hubble_version: '>3 AND <7 AND <8'
+      module: win_reg
+      items:
+        - args:
+            name: 'HKEY_LOCAL_MACHINE\Software\Policies\Microsoft\Windows\Explorer\NoDataExecutionPrevention'
+          comparator:
+            type: "dict"
+            match:
+              HKEY_LOCAL_MACHINE\Software\Policies\Microsoft\Windows\Explorer\NoDataExecutionPrevention:
```
jettero, line 21: these backslashes are incorrect and will fail in python 3.7 or thereabouts.

Please add `r` to the `"""` (ie, `r"""`) or add another backslash to escape your escape. 

Python 3.6 and below can deal with this and does what you'd expect, but future pythons will not.

jettero: In fact, this file won't load enough for me to even show the result under default warnings:

```
PYTHONPATH=. python -W default hubblestack/extmods/hubble_mods/win_reg.py 
  File "hubblestack/extmods/hubble_mods/win_reg.py", line 67
    """
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 426-427: malformed \N character escape
```

jettero (changes requested):
Instead of inventing a whole new module runner system, please use the module runner already built into the hubblestack.loader (well, it's still the salt loader in this branch). There's no need to re-invent the wheel on this.

devagupt:
> Please fix the history. There's a merge commit at the tail end of the commit list that's from some unknown branch.

fixed

jettero (approved):

PR #937, opened 2020-09-28
Bug fix: Package list not getting refreshed

goravsingal:
Bug fix: Package list not getting refreshed

List of installed packages on a system is not refreshed in consecutive executions of Nova. As __context__ dictionary was being used to cache this result and was not being cleared.


jettero (approved):

jettero:
I think we're dangerously close to merging 4.1 (should be called saltless-hubble) branch to develop, so it probably doesn't matter, but I feel like this should be applied to develop branch (the next version) rather than the saltless-hubble branch.

jettero:
Oh, this doesn't apply cleanly at all if you chnagne the base branch...

jettero (approved):

PR #944, opened 2020-10-19
hubblestack.config

jettero:
There's some really spooky things in this patch. 

(We're getting dangerously close to the point where we try this without loading the `salt-ssh` package.)


hubblestack/config.py
```diff
@@ -0,0 +1,1995 @@
...
+    opts = {}
+
+    if hubblestack.utils.validate.path.is_readable(path):
+        try:
+            opts = _read_conf_file(path)
+            opts["conf_file"] = path
+        except HubbleConfigurationError as error:
+            log.error(error)
+            if exit_on_config_errors:
+                sys.exit(salt.defaults.exitcodes.EX_GENERIC)
```
Tenebriso, line 1647: I think we can use the hubblestack.defaults.exitcodes here

hubblestack/config.py
```diff
@@ -0,0 +1,1995 @@
...
+                )
+
+        for fn_ in sorted(glob_matches):
+            log.debug("Including configuration from '%s'", fn_)
+            try:
+                opts = _read_conf_file(fn_)
+            except HubbleConfigurationError as error:
+                log.error(error)
+                if exit_on_config_errors:
+                    sys.exit(salt.defaults.exitcodes.EX_GENERIC)
```
Tenebriso, line 1696: Here too

hubblestack/modules/config.py
```diff
@@ -0,0 +1,373 @@
...
+    # Gated for salt-ssh (salt.utils.cloud imports msgpack)
+    import salt.utils.cloud
+    HAS_CLOUD = True
+except ImportError:
+    HAS_CLOUD = False
+
+import salt._compat
+import salt.syspaths as syspaths
+import salt.utils.sdb as sdb
+
```
Tenebriso, line 29: Can these imports be switched to hubblestack.* or does it cause a failure?

hubblestack/modules/config.py
```diff
@@ -0,0 +1,373 @@
...
+# Import 3rd-party libs
+from salt.ext import six
```
Tenebriso, line 31: We should probably get rid of this, the `salt.ext` has already been removed everywhere else I think and we'd be stuck with it, and change `six.string_types` to just `str`

hubblestack/config.py
```diff
@@ -0,0 +1,1995 @@
+# -*- coding: utf-8 -*-
+"""
+All Hubble configuration loading and defaults should be in this module
+"""
+
+# Import python libs
+from __future__ import absolute_import, print_function, unicode_literals, generators
```
Tenebriso, line 7: I think we can remove then __future__ imports we're running python 3, do we still need them for some reason?

hubblestack/defaults/exitcodes.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
```
Tenebriso, line 1: I think we only use a couple of these (I've only seen EX_GENERIC) but I'm all for keeping them to use in the future

hubblestack/status.py
```diff
@@ -598,9 +598,9 @@ def _setup_for_testing():
...
     import hubblestack.daemon
     parsed_args = hubblestack.daemon.parse_args()
-    import salt.config
+    import hubblestack.config
     parsed_args['configfile'] = config_file = '/etc/hubble/hubble'
-    __opts__ = salt.config.minion_config(config_file)
+    __opts__ = hubblestack.config.get_config(config_file)
     __opts__['conf_file'] = config_file
     __opts__.update(parsed_args)
     import salt.loader
```
Tenebriso, line 606: should we change it to hubblestack.loader instead?

jettero: I intend to do a huge cleanup wave for stuff like that. You're totally right of course, but I don't think it's necessary just now.

hubblestack/config.py
jettero: Excellent point... I'll have to go fix that.

hubblestack/modules/config.py
jettero: I thought I got all these. Nice catch.

hubblestack/config.py
jettero: prolly just use sys.exit(0)

hubblestack/status.py
jettero: Yeah, I went ahead and changed this. It only lives in an unused setup and testing function ... or, at least, I doubt anyone's used it since I developed that module.

hubblestack/defaults/exitcodes.py
jettero: I can't recall... I discussed removing the exit codes with someone. I can't recall if we decided to remove it or not or who I discussed it with, but I definitely decided to keep this -- perhaps vacillating on a previous decision.

I'm not totally sure I understand your comment either, but I think it may be in reference to deciding to remove this file and then deciding to keep it.

Tenebriso (approved):

PR #960, opened 2020-12-08
Versioning changes

devagupt:
- Hubble version and checkout tag name are passed as argument in docker file. It will help in creating builds by parsing tag name from git.


jettero (approved):
I was tempted to change this to an ARG myself... it would have simplified a couple things during my testing.

Approve.

PR #962, opened 2020-12-09
Changing rc0 from version

devagupt:
- Removing rc0 from version as '_' character with it is causing issues in creating hubble binaries for debian systems.
Instead of rc versioning scheme, we should just bump up the minor version.
- Exporting requirements file as it is required for Tessa posting

### DIFF
--- a/hubblestack/version.py
+++ b/hubblestack/version.py
@@ -1,4 +1,4 @@
 # coding: utf-8
 
-__version__ = '4.5.0_rc0'
+__version__ = '4.5.0'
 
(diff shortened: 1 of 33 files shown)


MoodyMudit (dismissed):

jettero:
no, we need to be able to do rc versions and things. this is something that definitely did work previously. we had trouble with a '-' and now there's trouble with '_' ?? 

can we just say 4.5.0rc0 ? just no character?

devagupt:
> no, we need to be able to do rc versions and things. this is something that definitely did work previously. we had trouble with a '-' and now there's trouble with '_' ??
> 
> can we just say 4.5.0rc0 ? just no character?

I don't understand why do we require 'rc' versioning. Can't we bump up minor version like '4.5.0' to '4.5.1' if different build/tag comes up ?

jettero (approved):

PR #965, opened 2020-12-22
Bringing back the fdg function to support fdg.fdg functionality

MoodyMudit:


jettero (changes requested):
We don't really have deprecation markings that we use (as far as I know).

Could we maybe add something to the description to indicate this is the /old/ fdg.fdg and the /new/ fdg.run is mean to replace it? Am I getting that totally wrong? I recall that being the tail end of the discussion.

jettero:
Er, by "description" I mean the function docs for fdg.run and/or fdg.fdg...

goravsingal (approved):

jettero (approved):

PR #972, opened 2021-01-08
Added MD5 FIPS compatibility 

daveed12:


MoodyMudit:
There is one more places where this needs to be done
https://github.com/hubblestack/hubble/blob/develop/hubblestack/modules/nebula_osquery.py#L1232

hubblestack/modules/network.py
```diff
@@ -1560,7 +1561,10 @@ def connect(host, port=None, **kwargs):
         if proto == "udp":
             # Generate a random string of a
             # decent size to test UDP connection
-            md5h = hashlib.md5()
+            if isFipsEnabled:
+                md5h = hashlib.md5(usedforsecurity=False)
+            else:
+                md5h = hashlib.md5()
```
MoodyMudit, line 1567: Need to make this change at https://github.com/hubblestack/hubble/blob/develop/hubblestack/modules/nebula_osquery.py#L1232
also

MoodyMudit (changes requested):

hubblestack/modules/network.py
daveed12: Good catch Mudit

MoodyMudit (approved):

jettero:
this will pass tests after #968 merges; I had to fix tests mudit broke by re-enabling various config options

PR #983, opened 2021-01-17
pip 20.2 fix for coreos

jettero:
(plus, copy #982 fix back to dev/ for centos6)


jettero:
the actual purpose of this PR was missed #984 has the patch this was supposed to have...

PR #985, opened 2021-01-18
fixing sqlite returner

MoodyMudit:
Sqlite returner wasn't working. Fixing that.

### DIFF
--- a/hubblestack/returners/sqlite.py
+++ b/hubblestack/returners/sqlite.py
@@ -125,7 +125,7 @@ def _open_close_conn(func):
         if not kwargs['conn']:
             log.exception('failed to retrieve sqlite connection object')
             return
-        results = func(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


hubblestack/returners/sqlite.py
```diff
@@ -187,13 +187,17 @@ def get_load(jid, conn=None):
 @_open_close_conn
 def _insert_helper(ret, conn=None):
     log.debug('populating jids table with %s', ret.get('jid'))
+    ret_jid = '' if ret.get('jid') is None else ret.get('jid')
```
goravsingal, line 190: Dictionary has a simple method, this line can be replaced by:

ret_jid = ret.get('jid', '')

Same for the rest of the lines.

goravsingal (changes requested):

goravsingal (approved):

PR #993, opened 2021-01-21
amazonlinux (as late as 2018) doesn't use systemd

jettero:


daveed12 (approved):

fossam:
@jettero We should differentiate AL2 from earlier versions as it definitely uses systemd.

fossam (approved):

jettero:
@fossam and I decided over chat that AL2 can probably use init.d files just fine. While not ideal, this at least means we don't have to figure out which startup files to use in a post install script.

PR #997, opened 2021-01-23
Fixing chaining issues

MoodyMudit:
A few audit modules did not support chaining. Fixed them.

### DIFF
--- a/hubblestack/audit/time_sync.py
+++ b/hubblestack/audit/time_sync.py
@@ -300,7 +300,10 @@ def get_filtered_params_to_log(block_id, block_dict, extra_args=None):
 
 
 def _get_ntp_servers(block_id, block_dict, extra_args):
-    ntp_servers = runner_utils.get_param_for_module(block_id, block_dict, 'ntp_servers')
+    try:
+        ntp_servers = runner_utils.get_param_for_module(block_id, block_dict, 'ntp_servers')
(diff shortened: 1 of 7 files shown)


hubblestack/audit/time_sync.py
```diff
@@ -300,7 +300,10 @@ def get_filtered_params_to_log(block_id, block_dict, extra_args=None):
 
 
 def _get_ntp_servers(block_id, block_dict, extra_args):
-    ntp_servers = runner_utils.get_param_for_module(block_id, block_dict, 'ntp_servers')
+    try:
+        ntp_servers = runner_utils.get_param_for_module(block_id, block_dict, 'ntp_servers')
+    except Exception:
+        log.debug("Input profile %s does not contain ntp_servers", block_id)
```
goravsingal, line 306: 1. I hope we should not get any exception here. As we did not put except block in any of module. May be we can fix it in get_param_for_module
2. If we need to keep above, log message should convey that it might also try chaining params.

goravsingal (changes requested):

hubblestack/audit/time_sync.py
MoodyMudit: since I have handled it in **get_param_for_module**, reverting it from here.

goravsingal (approved):

PR #999, opened 2021-01-25
Adding missing Misc module method

goravsingal:
Adding missing Misc module method: test_mount_attrs()

### DIFF
--- a/hubblestack/audit/misc.py
+++ b/hubblestack/audit/misc.py
@@ -200,6 +200,14 @@ Functions supported:
         pattern (Mandatory)
         values (Mandatory)
         comparetype (Default 'regex')
+- test_mount_attrs
+    Ensure that a given directory is mounted with appropriate attributes
+    If check_type is soft, then in absence of volume, True will be returned
+    If check_type is hard, then in absence of volume, False will be returned
+    Params:
+        mount_name (Mandatory)
+        attribute (Mandatory)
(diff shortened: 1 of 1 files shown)


hubblestack/audit/misc.py
```diff
@@ -1112,6 +1121,33 @@ def _check_sshd_parameters(block_id, block_dict, extra_args=None):
...
+    Ensure that a given directory is mounted with appropriate attributes
+    If check_type is soft, then in absence of volume, True will be returned
+    If check_type is hard, then in absence of volume, False will be returned
+    """
+    mount_name = runner_utils.get_param_for_module(block_id, block_dict, 'mount_name')
+    attribute = runner_utils.get_param_for_module(block_id, block_dict, 'attribute')
+    check_type = runner_utils.get_param_for_module(block_id, block_dict, 'check_type', 'hard')
+
+    # check that the path exists on system
+    command = 'test -e ' + mount_name
```
MoodyMudit, line 1135: please use join for string concatenation.

MoodyMudit (changes requested):

hubblestack/audit/misc.py
jettero: I don't think join is necessary here. That just introduces another datastructure.

MoodyMudit (approved):

PR #1004, opened 2021-01-27
Build changes for Centos8 and Debian8

goravsingal:
Build changes for Centos8 and Debian8

On CentOS-8, fpm 1.12 (latest) is broken. So 1.11 is being used.
For Debian8, its already EOL. And, to build Python >=3.7, we need to build openssl first.


MoodyMudit (approved):

MoodyMudit:
The description is a little incorrect though. It says **Centos9**

PR #1011, opened 2021-02-04
Updating pip to latest version - 21.0.1 on all platforms

goravsingal:
Updating pip to latest version - 21.0.1 on all platforms

This is after the azurefs changes.
Now, we have the latest azurefs libraries, and the latest pip

For pyinstaller-hooks, now it comes pre-bundled with pyinstaller. No need for extra install.


jettero (changes requested):
If we fixed the 2020 resolver problem, then we should use the latest pip, not some other frozen version. What is the thinking with freezing the version at 21.01?

goravsingal:
Initial thought was of taking the latest pip (and fix the version). As we frequently encounter issues with different version of things.
Reverting it to get latest version

jettero (approved):

PR #1012, opened 2021-02-04
win pkg audit result/audit_value bugfix

MoodyMudit:
Negative test case wasn't correctly written


jettero:
This seems fine, but could we get a better name for the PR? 

Fixing what bug? Would we be able to figure out what this PR does later going through the  history?

jettero (approved):
nm, I can just edit the name of it... 

(also, in case you're wondering, I'm aware some of my PRs have bad names too. I try to fix them when I notice.)

goravsingal (approved):

PR #1020, opened 2021-02-10
locale fix for LC_ALL=C/LANG=C hosts with no locales installed

jettero:
It's very difficult to find documentation on `str.encode('charmap')` but it seems to want to use the character map of the environment locale. For `LANG=C` hosts, the charmap is "ascii" which breaks things using OpenSSL because of this `byte_string` function.

The only fix I can think of is to enforce UTF8 with a monkeypatch. We can't know if a host actually has any UTF-8 locales installed, so the best we can do is assume UTF-8 like whatever other module is sending UTF-8 characters through OpenSSL.

### DIFF
--- a/hubblestack/utils/signing.py
+++ b/hubblestack/utils/signing.py
@@ -53,6 +53,7 @@ from Crypto.IO import PEM
 from Crypto.Hash import SHA256
 
 import OpenSSL.crypto as ossl
+import OpenSSL._util
 
 from cryptography.exceptions import InvalidSignature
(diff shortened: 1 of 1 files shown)


daveed12 (approved):

MoodyMudit:
Can you please describe what the monkey patch is and why is this being done.

jettero:
> Can you please describe what the monkey patch is and why is this being done.

Ah, yes. Frankly, I'm embarrassed by the previous description. Fixed.

MoodyMudit (approved):

PR #1040, opened 2021-03-02
FIPS changes with Python 3.9

goravsingal:
- Build FIPS compliant Openssl
- Patch Python 3.9 for get/set fips mode methods in Python
- Dockerfile changes

These are just POC changes. Code will also require changes for Python 3.9


jettero (changes requested):
please continue to use pyenv so the build process isn't so vastly different.

pyenv allows for patching:

https://github.com/pyenv/pyenv/blob/master/plugins/python-build/README.md#applying-patches-to-python-before-compiling

jettero:
Since I have my doubts that the fips support will ever merge with mainline, I wonder if we should have a fips branch (instead of pointing the PR at develop).

goravsingal:
Added Pyenv back. It works, took a while to see how patching works with Pyenv. Had to update patch. It was not accepting strip-level (patch -p flag).
Also added python patch.

And, we may actually want to merge this code to mainline. There is no issue that I see. Python patch is just for get/set fips mode.

jettero:
Per our offline discussion, I'm ok with merging to mainline when it's ready. 

> Added Pyenv back. It works, took a while to see how patching works with Pyenv. Had to update patch.

Thanks for restoring pyenv... but you shouldn't need to modify the patch at all.

```
$ env PYTHON_PATCH_PATH="./patches" PYTHON_PATCH_OPTS="-p1" pyenv install  3.7.9
```

PR #1042, opened 2021-03-05
adding support for old fdg

MoodyMudit:
With this code change, both fdg versions can be used to execute profiles. If for example, **hubble fdg.run** is called, hubble will execute profiles kept in **salt://fdg/<profiles>** folder. 
Secondly, if hubble fdg.run fdg_version=fdg_v2 is called, then profiles kept in **salt://fdg_v2/<profiles>** are executed. The profiles in this folder need to be in new format.

Can be executed via hubble schedule as follows
`  fdg_top:
    function: fdg.top
    kwargs:
      fdg_version: fdg_v2
    seconds: 60
    splay: 36
    returner: splunk_fdg_return
    run_on_start: True`

### DIFF
--- a/hubblestack/modules/fdg.py
+++ b/hubblestack/modules/fdg.py
@@ -130,9 +130,9 @@ __returners__ = None
 RETURNER_ID_BLOCK = None
 BASE_DIR_FDG_PROFILES = {'fdg_v2':'fdg_v2', 'fdg':'fdg'}
 
-def fdg(fdg_file, starting_chained=None, fdg_version='fdg'):
+def fdg(fdg_file=None, starting_chained=None, fdg_version='fdg'):
     """
-    fdg_version can take two values 'fdg' and 'fdg_v2' and it defines whether 
+    fdg_version can take two values 'fdg' and 'fdg_v2' and it defines whether
     fdg is executed with profiles in old format or new format.
 
     This is the old fdg function, which should ideally be deprecated, but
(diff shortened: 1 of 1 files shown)


hubblestack/modules/fdg.py
```diff
@@ -185,11 +194,13 @@ def fdg(fdg_file, starting_chained=None):
     ret = _fdg_execute('main', block_data, chained=starting_chained)
     return RETURNER_ID_BLOCK, ret
 
-def run(fdg_file=None, starting_chained=None):
+def run(fdg_file=None, starting_chained=None, fdg_version='fdg'):
```
goravsingal, line 197: Since fdg.run was meant for v2 only. So default value here should be kept: 'fdg_v2'.
Else in every case to run new fdg, we need to pass extra param

hubblestack/modules/fdg.py
```diff
@@ -247,7 +262,15 @@ def top(fdg_topfile='top.fdg'):
...
     in the dictionary are the associated returns from the fdg runs.
     """
-    fdg_routines = _get_top_data(fdg_topfile)
+    fdg_routines = _get_top_data(fdg_topfile, fdg_version)
+    # there are currently two cases for fdg_version 'fdg' and 'fdg_v2'. 
+    # We call the function 'fdg' when the old fdg is refered.
+    # And 'run' when fdg_v2 is called.
+    if fdg_version == 'fdg':
+        call_function = fdg
+    else:
```
goravsingal, line 271: As we have done input check at other places. We can check this value can be only in (fdg, fdg_v2)

Else here, for any value other than 'fdg', it will run v2

hubblestack/modules/fdg.py
```diff
@@ -256,20 +279,20 @@ def top(fdg_topfile='top.fdg'):
...
 
 
-def _get_top_data(topfile):
+def _get_top_data(topfile, fdg_version):
     """
     Helper method to retrieve and parse the FDG topfile
     """
     if not topfile.startswith('salt://'):
-        topfile = 'salt://' + BASE_DIR_FDG_PROFILES + os.sep + topfile
+        topfile = 'salt://' + BASE_DIR_FDG_PROFILES.get(fdg_version) + os.sep + topfile
```
goravsingal, line 295: Continuing for input validation comment.
In that case, BASE_DIR_FDG_PROFILES.get(fdg_version) will be  None

goravsingal (changes requested):
Now we have enabled both run, fdg to run both kind of profiles. That looks good.

hubblestack/modules/fdg.py
MoodyMudit: done

hubblestack/modules/fdg.py
MoodyMudit: done

hubblestack/modules/fdg.py
MoodyMudit: handled input validation

goravsingal (approved):

PR #1048, opened 2021-03-12
adding a method for fetching failure reason from module

MoodyMudit:
Adding a method in modules that can tell the audit runner what it was doing. This message can be then clubbed with failure reason from comparator to tell the user what actually failed on the system.

### DIFF
--- a/hubblestack/audit/command_line_parser.py
+++ b/hubblestack/audit/command_line_parser.py
@@ -457,3 +457,22 @@ def get_filtered_params_to_log(block_id, block_dict, extra_args=None):
         'key_aliases': key_aliases,
         'delimiter': delimiter
     }
+
+
+def get_failure_reason(block_id, block_dict, extra_args=None):
+    """
+
+    :param block_id:
+        id of the block
+    :param block_dict:
+        parameter for this module
+    :param extra_args:
+        Extra argument dictionary, (If any)
+        Example: {'chaining_args': {'result': {'cmdline': 'app --config-file=test test'}, 'status': True},
+                  'caller': 'Audit'}
+    :return:
+    """
+    key_aliases = runner_utils.get_param_for_module(block_id, block_dict, 'key_aliases')
+    command_line = runner_utils.get_param_for_module(block_id, block_dict, 'cmdline')
+    failure_reason = "Fetching value of {0} in {1}".format(key_aliases, command_line)
+    return failure_reason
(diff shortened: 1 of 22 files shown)


hubblestack/audit/command_line_parser.py
```diff
@@ -457,3 +457,22 @@ def get_filtered_params_to_log(block_id, block_dict, extra_args=None):
...
+    :param block_dict:
+        parameter for this module
+    :param extra_args:
+        Extra argument dictionary, (If any)
+        Example: {'chaining_args': {'result': {'cmdline': 'app --config-file=test test'}, 'status': True},
+                  'caller': 'Audit'}
+    :return:
+    """
+    key_aliases = runner_utils.get_param_for_module(block_id, block_dict, 'key_aliases')
+    command_line = runner_utils.get_param_for_module(block_id, block_dict, 'cmdline')
```
goravsingal, line 476: We are not considering chaining params. As we did in execute(). 
Please check in all modules.

hubblestack/audit/command_line_parser.py
```diff
@@ -457,3 +457,22 @@ def get_filtered_params_to_log(block_id, block_dict, extra_args=None):
         'key_aliases': key_aliases,
         'delimiter': delimiter
     }
+
+
+def get_failure_reason(block_id, block_dict, extra_args=None):
+    """
+
```
goravsingal, line 464: Also please add a one-line about this function (what it does) here.

hubblestack/audit/command_line_parser.py
```diff
@@ -457,3 +457,22 @@ def get_filtered_params_to_log(block_id, block_dict, extra_args=None):
...
+        parameter for this module
+    :param extra_args:
+        Extra argument dictionary, (If any)
+        Example: {'chaining_args': {'result': {'cmdline': 'app --config-file=test test'}, 'status': True},
+                  'caller': 'Audit'}
+    :return:
+    """
+    key_aliases = runner_utils.get_param_for_module(block_id, block_dict, 'key_aliases')
+    command_line = runner_utils.get_param_for_module(block_id, block_dict, 'cmdline')
+    failure_reason = "Fetching value of {0} in {1}".format(key_aliases, command_line)
```
goravsingal, line 477: We can directly return this.

goravsingal (changes requested):
These 3 comments are common for all modules.

hubblestack/audit/command_line_parser.py
MoodyMudit: As discussed, we will do this in future

hubblestack/audit/command_line_parser.py
MoodyMudit: Done

hubblestack/audit/command_line_parser.py
MoodyMudit: done

goravsingal (approved):

PR #1050, opened 2021-03-15
fdg module should return negative with error case

MoodyMudit:
With the current code, even if error was present, the fdg module resulted in **Success**. Fixed that.

### DIFF
--- a/hubblestack/audit/fdg.py
+++ b/hubblestack/audit/fdg.py
@@ -101,17 +101,11 @@ Module Arguments
     Example: 'salt://fdg/my_fdg_file.fdg'
 - starting_chained (Optional)
     Starting value as parameter for main(first) chaining block in fdg
-- true_for_success
-    The ``true_for_success`` argument decides how success/failure are decided
-    based on the fdg return. By default, any "truthy" value in the ``results`` piece
-    of the FDG return will constitute success. Set this option to False to treat
-    "falsey" values as success.
 - use_status
     The ``use_status`` argument determines whether the status result or the actual
     result returned from fdg will be used. If this is True, only the status result of
     the fdg run will be considered. If it is False, only the actual result of the
-    fdg run will be considered. Regardless, the ``true_for_success`` argument
-    will be respected.
+    fdg run will be considered.
 - consolidation_operator
     Only values allowed (and/or)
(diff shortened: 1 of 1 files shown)


hubblestack/audit/fdg.py
```diff
@@ -263,6 +263,8 @@ def execute(block_id, block_dict, extra_args=None):
 
     fdg_result, fdg_status = fdg_run
     check_value = fdg_status if use_status else bool(fdg_result)
+    if isinstance(fdg_result, dict) and 'error' in (k.lower() for k in fdg_result):
```
goravsingal, line 266: It has some issues:
1. fdg has a flag: true_for_success. If its false, False result is success. This code doesn't handle it.
2. If you look at https://github.com/hubblestack/hubble/blob/develop/hubblestack/files/hubblestack_nova/fdg.py#L118
 It is doing some branching based on true_for_success and value. 
In this current file, line 269 looks wrong, where we are checking just true_for_success == check_value.

3. And, if 'error' is coming in result dictionary. Then, its status should have been False. And we are not at all taking decision based on fdg_status variable.

goravsingal (changes requested):

hubblestack/audit/fdg.py
MoodyMudit: 1. The flag **true_for_success** is meant to be used when we have received either a positive or a negative result. In error cases, we do not receive a negative/positive result instead we receive this **({'error': 'pattern_not_found'}, False)**. So using the flag **true_for_success** does not seem to be an appropriate option.
2. Same reasoning as above. I want to exit early, because an error was encountered. I would go in the branching only if I received a positive/negative result.
3. **fdg_status** is only used when a flag **use_status** is set to **True** (default is **False**). Check [here](https://github.com/hubblestack/hubble/blob/develop/hubblestack/audit/fdg.py#L265). Actually this is what the bug is. If you notice, check_value is now equal to **bool(fdg_result)**, which means it is True foe even result of the form **{'error': 'pattern_not_found'}**. This is why finding if an error was found early and exiting is the best option here.

I would have wanted to put the check in an **error** (instead of a **Success** or **Failure**) category here, but the current design does not support that.

goravsingal (approved):

PR #1051, opened 2021-03-15
signing verification caching

jettero:
This is meant as a speedup to the verification checking.

I don't think this was ever the source of very much slowdown, but it certainly does seem silly to verify signatures over and over and over for every file in the profile.


daveed12 (approved):
LGTM

PR #1061, opened 2021-03-25
I think this will sort out The Elusive Signing Bug

jettero:
For the record, the Elusive Signing Bug is a subtle bug where when a file fails the repo signing checks, it's still sometimes served to the code that was looking for it. It was extremely difficult to reproduce for some reason (even though some people could reproduce it every time).

I spent a horrific amount of time tracking this down. I finally determined that `cp.cache_dir` was never meant to be a proper sync. The original authors of hubble.audit.sync provided a `clean` argument that would `file.remove` the cache dir to prevent objects deleted from the repo sticking around after being elided.

The problem is that removing the whole cache and re-downloading it on every schedule loop is expensive, so someone disabled it later. It was probably thought that the regular `file_client.channel.fs.update()` (which invokes the `fileserver.reap_fileserver_cache_dir()`) would properly handle these cases; but it definitely does not appear to do so. 

Essentially the problem is the double cache copy situation:

1. we first copy from the fileserver (roots, or gitfs or whatever) into `roots` (i.e. `/var/cache/hubble/roots`)
2. `cp.cache_dir` then copies from `roots` to `files` (i.e. `/var/cache/hubble/files`)
3. the daemon `file_client.channel.fs.update()` updates the `roots` (but never attempts to clean up the `files` copy).
4. signing needs these files to not be found, which is handled in `roots` but not `files`

It's worth noting that the unwelcome leftover files were at least downloaded in good faith. To get downloaded at all, they either had to be downloaded before signing was a thing or at some point when signing verified they were OK...

But they should definitely get cleaned up when they fail the signing pass.

The fix: I taught `cp.cache_dir` how to notice these files missing in `roots` that exist in `files`. It can now remove them automatically during what should have been a proper sync. Note that the old behavior can be restored by setting `cleanup_existing=False` but I can't think of a scenario where this would be desirable.

There's still the remaining question: 
Why did this affect `hubble.audit` but not `audit.run`?

Answer:
`hubble.audit` uses `cp.cache_dir` to copy the whole thing all at once; but `audit.run` uses `cp.cache_file` and checks the result. Specifically checking one file, will indeed reveal the file is missing in `roots` even if it exists in `files` as a spurious older file. (`module_runner.runner.make_file_available` also uses `cp.cache_file`, so the `hubble.audit` issue is entirely avoided in the successor).

my old (wrong) analysis:
~~Basically the problem is that various parts of the hubble code base read from the fileserver cache directly (rather than using the intended interfaces for such tasks). The repo signing checks are injected into the fileserver `find_file()` mechanisms, which normally trigger a `fileserver.reap_cache_dir()` hit, causing the file to be deleted from the cache.~~

~~But in some cases, apparently, a cache refresh isn't even part of the invocation of the execution module(s) in question. So the "reaping" never happens (at least not during the execution) and the failing file is serv^H^H^H^H read from the cache as if it was OK.~~

~~All this means, that it feels like a race condition to me. Eventually there would have been a cache refresh that would have cleared the failing file (via the reaping mechanism). This patch simply forces the reaping of files that fail the signing check so there's no need to wait for the reaping and no question about when it happens in the case of a signature check failure.~~

### DIFF
--- a/hubblestack/fileclient.py
+++ b/hubblestack/fileclient.py
@@ -271,6 +271,7 @@ class Client(object):
             to_remove = set(scan_files(scan_target, cachedir)) - set(ret)
             for i,item in enumerate(to_remove):
                 log.debug('cache_dir(%s) to_remove[%d]: %s', path, i, item)
+                os.unlink(item)
 
         return ret
 
--- a/hubblestack/fileclient.py
+++ b/hubblestack/fileclient.py
@@ -207,7 +207,8 @@ class Client(object):
         return ret
 
     def cache_dir(self, path, saltenv='base', include_empty=False,
-                  include_pat=None, exclude_pat=None, cachedir=None):
+                  include_pat=None, exclude_pat=None, cachedir=None,
+                  cleanup_existing=True):
         '''
         Download all of the files in a subdir of the master
         '''
@@ -255,6 +256,22 @@ class Client(object):
                     if not os.path.isdir(minion_dir):
                         os.makedirs(minion_dir)
                     ret.append(minion_dir)
+
+        def scan_files(target, cachedir):
+            log.debug('cache_dir(%s) scanning %s for files that should be removed', path, target)
+            for wpath, dirs, files in os.walk(target):
+                for file in files:
+                    log.trace('cache_dir(%s) %s/%s seems potentially to be worth removing', path, wpath, file)
+                    yield os.path.join(wpath,file)
+
+        if cleanup_existing:
+            cachedir = self.get_cachedir(cachedir)
+            scan_target = os.path.join(cachedir, 'files', saltenv, path)
+            log.debug('cache_dir(%s) cleanup_existing is set', path)
+            to_remove = set(scan_files(scan_target, cachedir)) - set(ret)
+            for i,item in enumerate(to_remove):
+                log.debug('cache_dir(%s) to_remove[%d]: %s', path, i, item)
+
         return ret
 
     def cache_local_file(self, path, **kwargs):
--- a/hubblestack/fileserver/__init__.py
+++ b/hubblestack/fileserver/__init__.py
@@ -171,7 +171,8 @@ def reap_fileserver_cache_dir(cache_base, find_func):
                 ret = find_func(filename, saltenv=saltenv)
                 # if we don't actually have the file, lets clean up the cache
                 # object
-                log.debug('REAP file_path=%s, file_rel_path=%s; ret = find_func => %s', ret)
+                log.debug('REAP file_path=%s, file_rel_path=%s; ret = find_func => %s',
+                    file_path, file_rel_path, ret)
                 if ret['path'] == '':
                     log.debug('REAP os.unlink(%s)', file_path)
                     os.unlink(file_path)
--- a/hubblestack/fileserver/__init__.py
+++ b/hubblestack/fileserver/__init__.py
@@ -171,7 +171,9 @@ def reap_fileserver_cache_dir(cache_base, find_func):
                 ret = find_func(filename, saltenv=saltenv)
                 # if we don't actually have the file, lets clean up the cache
                 # object
+                log.debug('REAP file_path=%s, file_rel_path=%s; ret = find_func => %s', ret)
                 if ret['path'] == '':
+                    log.debug('REAP os.unlink(%s)', file_path)
                     os.unlink(file_path)
 
 
--- a/hubblestack/fileserver/roots.py
+++ b/hubblestack/fileserver/roots.py
@@ -137,13 +137,13 @@ def update():
     When we are asked to update (regular interval) lets reap the cache
     """
     try:
-        hubblestack.fileserver.reap_fileserver_cache_dir(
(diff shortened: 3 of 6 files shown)


MoodyMudit:
Can you please describe what **The Elusive Signing Bug** was. Probably you and I know today what the bug is, but no one in the open source community knows. Also, I will forget what **The Elusive Signing Bug** is in a few weeks at max. So I would prefer either the heading or the description to mention what the bug is. Thanks.

MoodyMudit:
Also, this seems to have broken the unit tests somewhere.

jettero:
> Can you please describe what **The Elusive Signing Bug** was. Probably you and I know today what the bug is, but no one in the open source community knows. Also, I will forget what **The Elusive Signing Bug** is in a few weeks at max. So I would prefer either the heading or the description to mention what the bug is. Thanks.

Fair point. I will edit the description. I wasn't able to test t his very much yesterday. I'll figure out what the problems are and fix it presently.

jettero:
Also, the "intended interfaces" ... to be clear, what the audit modules are doing is OK. It'd be better to use `__mods__['cp.cache_file']('namehere')` rather than `__mods__['cp.get_cachedir']()`; but it'd be slower, which is why I suspect someone started reading from the cachedir directly like that.

What the `salt.pillar` loader does is build a fileclient and use that to fetch the files -- which is arguably what nova (and later audit) should have done... but it's s small point.

```
self.client = salt.fileclient.get_file_client(__opts__, True)
topfile = salt.client.cache_file('namehere', environmentnamehere)
if topfile:
    # proceed
```

But hubble doesn't do the above, so ... a few years later, we get this pull request.

jettero:
> Also, this seems to have broken the unit tests somewhere.

OK, I fixed up this PR. Thanks for your comments.

jettero:
Turns out @daveed12 can reproduce the bug in one of his VMs. This patch does not work...

Coming up with plan B ...

rikhuran:
@jettero I could see one more issue wherein the signing_verify: false in the config file doesn't seems to be working.
repo_signing:
  require_verify: false
When running audit commands with false value it still checks and verify the signature and if signature is manipulated it fails rather it should not verify the signature and move on as require_verify: false is set. 
Please have a look at it as well..Let me know if you want anything else.

jettero:
@daveed12 @MoodyMudit I was finally able to reproduce the problem and do a full analysis of it. I changed the description of this PR to reflect that and describe the whole broken workflow, the fix, and why this affected `hubble.audit` but not `audit.run`.

jettero:
> When running audit commands with false value it still checks and verify the signature and if signature is manipulated it fails rather it should not verify the signature and move on as require_verify: false is set.

This is the correct behavior. We could possibly add an additional option to disable signing all together, but the thinking at the time this was originally written was:

If the MANIFEST exists and files don't match it, they're probably bad. If the SIGNATURE exists and it doesn't match, the MANIFEST is probably bad. This is regardless of whether we can authenticate the SIGNATURE to a trusted authority. `require_verify` requires that trusted authority approves of the SIGNATURE and MANIFEST and makes no comment about anything else.

jettero:
I checked some related cache loads in audit... they all seem safe.

The only other place I'm finding cache loads is the `win_pkg.refresh_db` use of `cp.cache_dir` which I rendered safe with the patch.

daveed12 (dismissed):

rikhuran:
@jettero I created a rpm package from your branch and tested the signing fix. It is working fine now as expected. But what i could see in the console logs is that there is no need to print the stack trace after we have printed "FileNotFoundError". Below is the log. Let me know if you require anything else.

  File "/opt/hubble/hubble-libs/hubblestack/modules/hubble.py", line 739, in _get_top_data
    with open(topfile) as handle:
FileNotFoundError: [Errno 2] No such file or directory: '/var/cache/hubble/files/base/hubblestack_nova_profiles/top.nova'

During handling of the above exception, another exception occurred:

  File "hubble.py", line 5, in <module>
  File "hubblestack/daemon.py", line 73, in run
  File "hubblestack/daemon.py", line 163, in main
  File "hubblestack/daemon.py", line 426, in run_function
  File "hubblestack/status.py", line 413, in inner
  File "/opt/hubble/hubble-libs/hubblestack/modules/hubble.py", line 110, in audit
    labels=labels)
  File "hubblestack/status.py", line 413, in inner
  File "/opt/hubble/hubble-libs/hubblestack/modules/hubble.py", line 510, in top
    data_by_tag = _build_data_by_tag(topfile, results)
  File "/opt/hubble/hubble-libs/hubblestack/modules/hubble.py", line 549, in _build_data_by_tag
    top_data = _get_top_data(topfile)
  File "/opt/hubble/hubble-libs/hubblestack/modules/hubble.py", line 743, in _get_top_data
    log.error('error loading nova topfile: %s', e)
NameError: name 'e' is not defined

jettero:
> NameError: name 'e' is not defined

@rikhuran that's interesting. I thought we found all of those ... checking ... Yeah, indeed. There was a later commit you may not have pulled that addressed that ("spelling error" or some such).

We actually created an r11 pre-release tag with this #1061 PR built in.

PR #1071, opened 2021-04-06
allow for http://proxy:servers with https hec endpoints

jettero:
There seems to be some confusion about the proxy settings and I'm not sure how to fix the docs or indeed, if there are any docs.

A top level config item of `https_proxy` will be used by the hec obj unless there's a `proxy` setting in the hec options for an individual hec endpoint; in which case, that will be used instead.

The https/http protocol of the proxy was made to follow the `http_event_server_ssl` setting for the endpoint no matter what was in the proxy setting. This patch allows for the use of http proxies with https endpoints... which is a thing that should hopefully only come up in testing? probably? It certainly did for me.

I found that some https things would definitely work with squid3, but only if it was in certificateles https mode; which the PoolManager and/or splunk do not seem to like at all. In http mode, squid3 would simply 403 the connections to splunk without actually talking to splunk.

Tinyproxy did work for me in http mode with the https splunk endpoint via the HTTPS_CONNECT tunnel method and no extra settings.

Please let me know what you find with this. Splunk+proxy behaviors are a little baffling.

### DIFF
--- a/hubblestack/hec/obj.py
+++ b/hubblestack/hec/obj.py
@@ -198,10 +198,26 @@ class HEC(object):
         self.currentByteLength = 0
         self.server_uri = []
 
-        if proxy and http_event_server_ssl:
-            self.proxy = 'https://{0}'.format(proxy)
-        elif proxy:
-            self.proxy = 'http://{0}'.format(proxy)
+        if proxy:
+            proxy_s = proxy.split('://')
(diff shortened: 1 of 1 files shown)


sumanmehta (approved):
LGTM as it does not disturb the old configs.

MoodyMudit (approved):

PR #1102, opened 2021-06-10
Loader fixes

jettero:
There's a few things here... not the least of which is a pretty obvious memory leak that we've had for as long as we've had grains-refresh...

The primary purpose of this PR is the 'POP' though. The persistent `__opts__` protection mechanism. It's definitely an anti-pattern that we're logging data by dropping it in config and publishing config... man, I wish we didn't do that, but we do it in a huge number of places. 

Recently, for some reason, the lazy loader refreshes have been resetting some of the `__opts__` return data. I can't see why it hadn't been doing it before now, but it's doing it now... so the solution -- until we address the anti-pattern -- is to trick the lazy loader into preserving __opts__.

Lastly, you may wonder: why is there no unit test to make sure this works?

This was a "Heisenbug." Recreating the bug is difficult. You have to trick the garbage collection into firing at the right time to trigger the bug and hope your config pulisher publishes the broken state. It's probably not possible to recreate in a unit test. Or at least, if you could do it, you couldn't do it reliably.

### DIFF
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -704,10 +704,11 @@ def refresh_grains(initial=False):
     persist, old_grains = {}, {}
     if initial:
         if not os.environ.get('NO_PRESERVE_OPTS'):
-            # 'POP' is for tracking persistent opts protection
             if os.environ.get('NOISY_POP_DEBUG'):
                 log.error('POP setting __opts__ to preservable (id=%d)', id(__opts__))
             hubblestack.loader.set_preservable_opts(__opts__)
+        elif os.environ.get('NOISY_POP_DEBUG'):
+            log.error('POP we are not attemting to protect __opts__ from lazyloader reloads')
     else:
         old_grains = copy.deepcopy(__grains__)
         for grain in __opts__.get('grains_persist', []):
--- a/hubblestack/daemon.py
+++ b/hubblestack/daemon.py
@@ -697,8 +697,16 @@ def refresh_grains(initial=False):
     global __returners__
     global __context__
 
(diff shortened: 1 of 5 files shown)


daveed12 (approved):

hubblestack/daemon.py
```diff
@@ -697,8 +697,19 @@ def refresh_grains(initial=False):
     global __returners__
     global __context__
 
+    # 'POP' is for tracking persistent opts protection
+    if os.environ.get('NOISY_POP_DEBUG'):
+        log.error('POP refreshing grains (id=%d)', id(__opts__))
```
fossam, line 702: Should this be logged to log.debug, rather than to log.error?

setup.py
```diff
@@ -23,6 +23,7 @@
     'pyopenssl',
     'requests>=2.13.0',
     'daemon',
+    'defusedxml',
```
fossam, line 26: Can the reason for this be explained?

hubblestack/daemon.py
jettero: Yeah, you'd think so. That's why it's disabled without the magic symbols.

I use the pop messages to track events in splunk, ... I can't see it in splunk if it's at debug level

setup.py
jettero: It was added a few bugfixes ago by an outside contributor. I forget which module, but there were security concerns with the xml processing and he was tired of patching his copy of hubble, so he submitted a PR that fixed various things about xml processing in that module. I objected to defused xml at first, but after looking at it... it's worth using for sure.

PR #1129, opened 2022-04-01
Flatcar support

grogsaxle:
Support for Flatcar ARM64

Required moving a couple of symbolic links around for osquery, but otherwise tracks with the coreos build


grogsaxle:
On my tests flatcar requires: 

    export SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt

daveed12:
LGTM

PR #1139, opened 2022-07-29
removed shell=True

mzainea:
Refactored audit/misc.py and hubblestack/misc.py to remove bash code that was being called from python and remove `shell=True` and `python_shell=True` statements.

### DIFF
--- a/hubblestack/files/hubblestack_nova/misc.py
+++ b/hubblestack/files/hubblestack_nova/misc.py
@@ -46,7 +46,7 @@ import os
 import stat
 import pathlib
 import re
-from pystemd.systemd1 import Manager()
+from pystemd.systemd1 import Manager
 from hubblestack.exceptions import CommandExecutionError
 
 log = logging.getLogger(__name__)
--- a/hubblestack/files/hubblestack_nova/misc.py
+++ b/hubblestack/files/hubblestack_nova/misc.py
@@ -43,10 +43,11 @@ import logging
 
 import fnmatch
 import os
+import stat
+import pathlib
 import re
-import hubblestack.utils
+from pystemd.systemd1 import Manager()
 from hubblestack.exceptions import CommandExecutionError
-from collections import Counter
 
 log = logging.getLogger(__name__)
 
@@ -272,8 +273,13 @@ def check_password_fields_not_empty(reason=''):
     """
     Ensure password fields are not empty
     """
-    result = _execute_shell_command('cat /etc/shadow | awk -F: \'($2 == "" ) { print $1 " does not have a password "}\'', python_shell=True)
-    return True if result == '' else result
+    result = ''
+    with open('/etc/shadow','r') as shadow:
+        lines = shadow.readlines()
+        for line in lines:
+            if line.split(':')[1] is "":
+                result += f"{line.split(':')[0]} does not have a password \n"
+        return True if result == '' else result
 
 
 def ungrouped_files_or_dir(reason=''):
@@ -281,8 +287,6 @@ def ungrouped_files_or_dir(reason=''):
     Ensure no ungrouped files or directories exist
     """
     raise CommandExecutionError('Module disabled due to performance concerns')
-    result = _execute_shell_command('df --local -P | awk {\'if (NR!=1) print $6\'} | xargs -I \'{}\' find \'{}\' -xdev -nogroup', python_shell=True)
-    return True if result == '' else result
 
 
 def unowned_files_or_dir(reason=''):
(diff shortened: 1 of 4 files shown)


hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -400,55 +410,40 @@ def check_path_integrity(reason=''):
     """
     Ensure that system PATH variable is not malformed.
     """
+    path_value = os.environ.get('PATH')
+    output = ''
+    for item in path_value.split(':'):
```
vladmonea, line 415: if `path_value` is `None`, which is the default value should the `'PATH'` key not be present in `os.environ`, this call will fail. Maybe ensure an appropriate default value is given?

hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -511,15 +506,17 @@ def check_core_dumps(reason=''):
...
+    # hard_core_dump_value = _execute_shell_command("grep -R -E \"hard +core\" /etc/security/limits.conf /etc/security/limits.d/ | awk '{print $4}'", python_shell=True).strip()
+    # hard_core_dump_value = hard_core_dump_value.split('\n') if hard_core_dump_value != "" else []
+    # if '0' in hard_core_dump_value:
+    #     return True
+    #
+    # if hard_core_dump_value is None or hard_core_dump_value == [] or hard_core_dump_value == "":
+    #     return "'hard core' not found in any file"
+    #
+    # return str(hard_core_dump_value)
+    return True
```
vladmonea, line 519: Why is this just being commented out and replaced with `return True`?

hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -546,58 +543,49 @@ def check_ssh_timeout_config(reason=''):
...
     """
     Ensure all users' home directories exist
     """
-
-    max_system_uid = int(max_system_uid)
-    users_uids_dirs = _execute_shell_command("cat /etc/passwd | awk -F: '{ print $1 \" \" $3 \" \" $6 \" \" $7}'", python_shell=True).strip()
-    users_uids_dirs = users_uids_dirs.split('\n') if users_uids_dirs else []
+    with open('/etc/passwd', 'r') as passwd:
+        lines = passwd.readlines()
+        users_uids_dirs = [' '.join([item.split(':')[0], item.split(':')[2], item.split(':')[5], item.split(':')[6].strip()]) for item in lines]
```
vladmonea, line 588: Maybe, for clarity:
```suggestion
        items = item.split(':')[0], item.split(':')[2], item.split(':')[5], item.split(':')[6].strip()
        users_uids_dirs = [' '.join(items) for item in lines]
```
You decide, this is just a suggestion.

vladmonea (changes requested):

hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -361,26 +370,27 @@ def test_mount_attrs(mount_name, attribute, check_type='hard'):
...
-        return "neither ntpd nor timesyncd is running"
-    else:
-        return True
+    manager = Manager()
+    manager.load()
+    services = manager.Manager.ListUnitFiles()
+    success = any([item for item in services if b'systemd-timesyncd' in item[0]
+                                                or b'ntpd' in item[0]
+                                                and b'enabled' in item[1]])
+    return success or 'neither ntpd nor timesyncd is running'
```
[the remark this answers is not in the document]
mzainea: yeah it's a case of the truthies, we'll deffinetly have to tug on this string and see what it unravels

hubblestack/files/hubblestack_nova/misc.py
mzainea: the check the code runs is successful even for documentation lines which are commented out, a little more research is needed to fix it but in the interim I commented it out so it stops returning incorrect results

hubblestack/files/hubblestack_nova/misc.py
```diff
@@ -326,25 +326,34 @@ def sticky_bit_on_world_writable_dirs(reason=''):
...
 
 def root_is_only_uid_0_account(reason=''):
     """
     Ensure root is the only UID 0 account
     """
-    result = _execute_shell_command('cat /etc/passwd | awk -F: \'($3 == 0) { print $1 }\'', python_shell=True)
-    return True if result.strip() == 'root' else result
+    uid0_accounts = []
+    with open('/etc/passwd', 'r') as passwd:
+        lines = passwd.readlines()
```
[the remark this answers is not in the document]
mzainea: fixed this

PR #1141, opened 2022-08-09
Replace pycrypto

vladmonea:
Since pycrypto is unmaintained, we're replacing the methods used in our signing and signature verification workflows.
We're also updating the code styling.

### DIFF
--- a/hubblestack/utils/signing.py
+++ b/hubblestack/utils/signing.py
@@ -62,6 +62,9 @@ from cryptography.hazmat.primitives.serialization import load_pem_private_key
 
 from hubblestack.utils.signing_utils import encode_pem, decode_pem
 
+from enum import Enum
+
+
 MANIFEST_RE = re.compile(r"^\s*(?P<digest>[0-9a-fA-F]+)\s+(?P<fname>.+)$")
 log = logging.getLogger(__name__)
 
@@ -153,7 +156,7 @@ def check_verify_timestamp(target, dampener_limit=None):
     return False
 
 
-class STATUS:
+class STATUS(Enum):
     """container for status code (strings)"""
 
     FAIL = "fail"
--- a/hubblestack/utils/signing.py
+++ b/hubblestack/utils/signing.py
@@ -45,6 +45,7 @@ import io as cStringIO
 import hubblestack.utils.platform
 
 from time import time
+from binascii import b2a_base64, a2b_base64
 from collections import OrderedDict, namedtuple
 
 # In any case, pycrypto won't do the job. The below requires pycryptodome.
@@ -60,7 +61,7 @@ from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import padding, utils
 from cryptography.hazmat.primitives.asymmetric import rsa
 from cryptography.hazmat.backends import default_backend
-from cryptography.hazmat.primitives.serialization import load_pem_private_key
+from cryptography.hazmat.primitives.serialization import load_pem_private_key, Encoding, PrivateFormat, NoEncryption
 
 MANIFEST_RE = re.compile(r"^\s*(?P<digest>[0-9a-fA-F]+)\s+(?P<fname>.+)$")
 log = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


hubblestack/utils/signing.py
```diff
@@ -63,39 +61,47 @@
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.serialization import load_pem_private_key
 
-MANIFEST_RE = re.compile(r'^\s*(?P<digest>[0-9a-fA-F]+)\s+(?P<fname>.+)$')
+MANIFEST_RE = re.compile(r"^\s*(?P<digest>[0-9a-fA-F]+)\s+(?P<fname>.+)$")
 log = logging.getLogger(__name__)
 
-def _our_byte_string(x, charmap='utf-8'):
```
grogsaxle, line 69: I'm not sure we want to drop the charmap argument here.   I understand that it is "unused", but may make for a runtime error if we do not include it.

hubblestack/utils/signing.py
```diff
@@ -124,35 +131,36 @@ def check_verif_timestamp(target, dampener_limit=None):
...
     if ts_0 is None:
         ts_0 = time()
-        verif_log_timestamps[target] = ts_0
+        VERIFY_LOG_TIMESTAMPS[target] = ts_0
         return True
 
     ts_1 = time()
     # make a timedelta from the loaded ts vs now.
-    td =  ts_1 - ts_0
+    td = ts_1 - ts_0  # pylint: disable=invalid-name ; this is fine
```
grogsaxle, line 150: 😆

hubblestack/utils/signing.py
```diff
@@ -420,68 +434,75 @@ def __init__(self, public_crt=None, ca_crt=None, extra_crt=None):
...
+                    message,
+                )
 
             # add to list of keys we'll use to check signatures:
             self.public_crt.append(self.PublicCertObj(i, digest, status))
 
 
 def stringify_ossl_cert(a_cert_obj):
-    """ try to stryingy a cert object into its subject components and digest hexification.
+    """try to stryingy a cert object into its subject components and digest hexification.
```
grogsaxle, line 452: Please fix spelling of stringify

hubblestack/utils/signing.py
grogsaxle: Maybe add this: 
# pylint: disable=unused-argument

hubblestack/utils/signing.py
```diff
@@ -111,8 +117,9 @@ def check_is_ca(crt):
         pass
     return False
 
+
 def check_verif_timestamp(target, dampener_limit=None):
```
grogsaxle, line 121: Please rename to 'check_verify_timestamp'

hubblestack/utils/signing.py
```diff
@@ -741,24 +844,24 @@ def verify_files(targets, mfname=None, sfname=None,
...
                 matched = MANIFEST_RE.match(line)
                 if matched:
-                    digest,manifested_fname = matched.groups()
+                    digest, manifested_fname = matched.groups()
                     manifested_fname = normalize_path(manifested_fname)
                     if manifested_fname in digests:
                         digests[manifested_fname] = digest
     # number of seconds before a FAIL or UNKNOWN is set to the returner
-    global verif_log_timestamps
+    global VERIFY_LOG_TIMESTAMPS
```
grogsaxle, line 864: why `global` here?

hubblestack/utils/signing.py
```diff
@@ -508,16 +529,56 @@ def manifest(targets, mfname=None):
...
+    :returns: PEM encoded string
+    """
+    output = "-----BEGIN {}-----\n".format(marker)
+    chunks = [b2a_base64(data[i : i + 48]).decode("latin-1") for i in range(0, len(data), 48)]
+    output += "".join(chunks)
+    output += "-----END {}-----".format(marker)
+    return output
+
+
+def decode_pem(data: bytes) -> str:
```
grogsaxle, line 557: May be able to use this instead: https://cryptography.io/en/latest/x509/reference/#loading-certificates

hubblestack/utils/signing.py
vladmonea: i'll try to kill it. with fire

hubblestack/utils/signing.py
vladmonea: can we keep him?

hubblestack/utils/signing.py
vladmonea: I'm not sure. `PEM.encode` and `PEM.decode` were being used to get a `base64` representation of the key/cert. The functions from `cryptography.x509` are highly specialised in what they do.

hubblestack/utils/signing.py
```diff
@@ -272,27 +283,28 @@ class X509AwareCertBucket:
...
+            if hasattr(HPPC, "public_crt"):
+                log.debug("using pre-packaged-public_crt")
                 public_crt = HPPC.public_crt
-            if hasattr(HPPC, 'ca_crt'):
-                log.debug('using pre-packaged-ca_crt')
+            if hasattr(HPPC, "ca_crt"):
+                log.debug("using pre-packaged-ca_crt")
                 ca_crt = HPPC.ca_crt
         except ImportError:
             pass
```
mzainea, line 310: should this really pass silently?

vladmonea: i'm not sure this should be fixed in a PR addressing a totally different thing. otherwise, fair point

vladmonea: @grogsaxle should we fix this in a different PR, or since we're here might as well?

vladmonea:
Closing PR because untangling the circular dependencies is too much hassle to be worth the effort

PR #1144, opened 2022-08-17
Msg filter

grogsaxle:


hubblestack/filter/filter_chain.py
```diff
@@ -0,0 +1,83 @@
...
+            not(isinstance(self.config["filter"], dict)) or \
+            self.config_label not in self.config["filter"].keys() or \
+            not(isinstance(self.config["filter"][self.config_label], dict)):
+            raise CommandExecutionError("FilterChain config not formatted correctly")
+
+        self.config = self.config['filter'][self.config_label]
+ 
+        for filter_name in self.config:
+            new_fltr = self._get_filter_class(self.config[filter_name]["type"])(filter_name, self.config[filter_tag])
+            self.chain.append(new_fltr)
```
vladmonea, line 63: Follow-up on my `property` suggestion. It was missing the `return`, as @mzainea said, and now `config` can be used as a variable inside the method's scope. Nevertheless, when we do `FilterChain.config` in some other code, it will return the config proper.

```suggestion
        ```suggestion
        self._config = self.config
        self.chain = []

    @property
    def config(self):
        config_path = "salt://filter_chain.yaml"
        config_path = __mods__["cp.cache_file"](config_path)
        try:
            with open(config_path, 'r') as handle:
                config = yaml.safe_load(handle)
        except Exception as e:
                config = {"default": {"filter": { "default": {
                 "sequence_id": { "label": "seq", "type": "hubblestack.filter.seq_id" },
                 "hubble_version": { "label": "hubble_version", "type": "hubblestack.hubble_version"}}}}}
            raise CommandExecutionError(f"Could not load filter config: {e}")

        if not isinstance(config, dict) or \
            "filter" not in config or \
            not(isinstance(config.get("filter"), dict)) or \
            config_label not in config["filter"].keys() or \
            not(isinstance(config.get("filter", {}).get(self.config_label), dict)):
            raise CommandExecutionError("FilterChain config not formatted correctly")

        config = config['filter'][config_label]
        
        for filter_name in self.config:
            new_fltr = self._get_filter_class(config[filter_name]["type"])(filter_name, config[filter_tag])
            self.chain.append(new_fltr)

         return config
```

grogsaxle (commented):
LGTM.  Using property per @vladmonea 's suggestion for config and chain.  Thanks!

vladmonea (approved):

PR #1146, opened 2022-08-31
Systemctl might not output the expected info

vladmonea:
We add a simple check to see whether there is a `features` line


grogsaxle (approved):
Good safety.   Thank you

